PR #24, opened 2021-08-26
Feat/support media types

### LINKED ISSUE #23: Feature: mux-video and mux-audio should support playback of additional media types
### Which Mux Elements/Packages would you like a new feature for?

- mux-video
- mux-audio

## 🎥 Describe the feature you'd like to request

Currently, mux-video and mux-audio only support playback of hls/m3u8 media. However, they should be able to support playback of other standard, "simpler" media container formats that can be played natively via an audio/video tag.

## 💻 Describe the use cases the feature will solve

Using `<mux-audio/>` and `<mux-video/>` to play any type of media content other than hls/m3u8.

## ⭐ Describe the solution you'd like to see

Add a `type` attribute that would expect mime-type/"mime-codec" values like the `<source/>` element's type (See: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#attr-type).

If a type is provided, use it to distinguish between HLS/m3u8 vs. other types.
If no type is provided, attempt to infer a mime-type based on the URL's extension.
If the src type is identified as HLS/m3u8, proceed in a functionally equivalent way to `<mux-video/>` / <`mux-audio/>` behavior (e.g. respecting `prefer-mse`, checking support, etc.).
If the src type is identified as anything else, for now, simply attempt to play it natively.

## 🔄 Describe alternatives or workarounds you've considered

- Don't use `<mux-video/>` or `<mux-audio/>` for non-hls.
- Fork or extend the elements.

cjpillsbury:
Resolves #23 and updates docs to reflect changes.
Also fixes bug in playback-id query param parsing.

### DIFF
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -37,9 +37,9 @@ const toPlaybackIdParts = (
   playbackIdWithOptionalParams: string
 ): [string, string?] => {
   const qIndex = playbackIdWithOptionalParams.indexOf("?");
+  if (qIndex < 0) return [playbackIdWithOptionalParams];
   const idPart = playbackIdWithOptionalParams.slice(0, qIndex);
   const queryPart = playbackIdWithOptionalParams.slice(qIndex);
-  if (!queryPart) return [idPart];
   return [idPart, queryPart];
 };
 
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -16,6 +16,7 @@ type AttributeNames = {
   METADATA_VIEWER_USER_ID: "metadata-viewer-user-id";
   PLAYBACK_ID: "playback-id";
   PREFER_MSE: "prefer-mse";
+  TYPE: "type";
 };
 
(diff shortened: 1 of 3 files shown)


dylanjha (approved):
The main goal here is to handle someone passing in `.mp4` source and have it not fail. This accomplished that.

1. If someone passes in `.m3u8`, then it goes down HLS path 👍🏼 
2. If someone has an HLS playlist and the URL happens to not end in `.m3u8` they can specify `type="application/vnd.apple.mpegurl"` or `type="hls"` 👍🏼 
3. Otherwise, the source simply gets passed through to the video el 👍🏼 


This LGTM. In all reality I wouldn't expect many people to be using `src=` directly.

The one thing that comes to mind is if people want to use `playback-id=` with something like `type="mp4"` but I think that's problematic for a few reasons:

1. By default, Mux assets don't have mp4s
2. Mux mp4s come in 3 different sizes, so the user would have to specify the size
3. For _most all_ use cases, HLS is what people should be using.

For that reason, I think `playback-id="" type="mp4"` is correctly not a thing. Just leaving the comment here for possible future discussion

PR #32, opened 2021-08-30
Fix/snowpack cjs

dylanjha:
✅ Fix yarn error:

> error Workspaces can only be enabled in private projects

✅ Fix snowpack build with `type: "module"` by using `.cjs`

https://github.com/snowpackjs/snowpack/discussions/2309

✅ add a CI step to make sure `yarn build` works (this would have saved me on #30)


cjpillsbury (approved):
😎

PR #37, opened 2021-10-01
Feat/ll and beacon support

cjpillsbury:
* Adds support for stream-type & stream-type-specific hls.js config.
* Adds support for mux-data beaconDomain
* Upgrades to hls.js v1.0.11

### DIFF
--- a/packages/mux-video/package.json
+++ b/packages/mux-video/package.json
@@ -19,7 +19,7 @@
     "build": "npm-run-all build:snowpack build:types"
   },
   "dependencies": {
-    "hls.js": "^1.0.9",
+    "hls.js": "1.0.11",
     "mux-embed": "^4.1.1"
   }
 }
--- a/packages/mux-video/package.json
+++ b/packages/mux-video/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mux-elements/mux-video",
-  "version": "0.1.2",
+  "version": "0.1.3",
   "type": "module",
   "description": "A custom mux video element for the browser that Just Works™",
   "main": "dist/index.js",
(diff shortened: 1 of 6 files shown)


packages/mux-video/package.json
```diff
@@ -19,7 +19,7 @@
     "build": "npm-run-all build:snowpack build:types"
   },
   "dependencies": {
-    "hls.js": "^1.0.9",
+    "hls.js": "1.0.11",
```
dylanjha, line 22: Calling out a change here. An alternate would be to specify:

```
"hls.js": "^1.0.11",
```

However, I think we *do* want to pin to a precise version to avoid pulling in versions of Hls.js that have an accidental regression.

The tradeoff will be that we have to actively keep this dependency up-to-date when Hls.js gets new patch releases.

packages/mux-video/src/index.ts
```diff
@@ -220,10 +281,12 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElementWithMux> {
 
     // 1. if we are trying to play an hls media source create hls if we should be using it "under the hood"
     if (hlsType && !shouldUseNative && hlsSupported) {
+      const streamTypeConfig = getStreamTypeConfig(this.streamType);
       const hls = new Hls({
         // Kind of like preload metadata, but causes spinner.
         // autoStartLoad: false,
         debug,
+        ...streamTypeConfig,
```
dylanjha, line 289: confirmed this works with `stream-type="ll-live"`

packages/mux-video/src/index.ts
```diff
@@ -248,6 +312,7 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElementWithMux> {
 
       mux.monitor(this.nativeEl, {
         debug,
+        beaconDomain,
```
dylanjha, line 315: Confirmed `beacon-domain` works as expected when supplied and when it's not supplied then `mux-embed` uses the default

dylanjha (approved):
LGTM -- go ahead and version bump?

We will want to follow on `<mux-audio>` with the same changes

PR #44, opened 2021-11-29
feat/mux-video-react: initial impl of mux-video-react. Basic CRA yarn…

cjpillsbury:
…+ts example package. Basic next.js yarn+ts example package.

Creating a draft PR for starting review and conversation at a high level.

To test:
* for `mux-video-react`, install deps via `yarn` and then run either `yarn dev` (for watch mode rebuilds) or `yarn build`
* in the new `/examples` directory, you should be able to install and run either/both `create-react-app-with-typescript` or `nextjs-with-typescript`, which has a basic demo of using `MuxVideo`.


dylanjha (commented):
A few things jumping out on first glance.

1. Instead of `elements/packages/demo-*` I would expect the `elements/packages/mux-video-react/examples/cra`. That way the top level `elements/packages` remains a single directory for each package and then the examples (or demos) of each package live inside the directory for that package. Thoughts?

2. Need a Readme for mux-video-react (I'm sure you know this 😄 ) 

3. I don't think we need `metadataVideoId`, `metadataVideoTitle` for `<MuxVideo>` -- just `metadata={}` -- thoughts? The main reason we have the former in `mux-video` is because serializing JSON into HTML attribute values is annoying. Since we can pass in POJO as props in React I think we can just do everything with the `metadata` key

cjpillsbury:
> A few things jumping out on first glance.
> 
> 1. Instead of `elements/packages/demo-*` I would expect the `elements/packages/mux-video-react/examples/cra`. That way the top level `elements/packages` remains a single directory for each package and then the examples (or demos) of each package live inside the directory for that package. Thoughts?

Yeah I'm 100% down. (I think I had a false memory of us deciding this route for the short term).

> 2. Need a Readme for mux-video-react (I'm sure you know this 😄 )

Fo sho.

> 3. I don't think we need `metadataVideoId`, `metadataVideoTitle` for `<MuxVideo>` -- just `metadata={}` -- thoughts? The main reason we have the former in `mux-video` is because serializing JSON into HTML attribute values is annoying. Since we can pass in POJO as props in React I think we can just do everything with the `metadata` key

Sounds 👌 to me. I was aiming for greater parity, but we can always re-add easily enough if the people demand it.

cjpillsbury:
> A few things jumping out on first glance.
> 
> 1. Instead of `elements/packages/demo-*` I would expect the `elements/packages/mux-video-react/examples/cra`. That way the top level `elements/packages` remains a single directory for each package and then the examples (or demos) of each package live inside the directory for that package. Thoughts?

Yeah I'm 100% down. (I think I had a false memory of us deciding this route for the short term).

> 2. Need a Readme for mux-video-react (I'm sure you know this 😄 )

Fo sho.

> 3. I don't think we need `metadataVideoId`, `metadataVideoTitle` for `<MuxVideo>` -- just `metadata={}` -- thoughts? The main reason we have the former in `mux-video` is because serializing JSON into HTML attribute values is annoying. Since we can pass in POJO as props in React I think we can just do everything with the `metadata` key

Sounds 👌 to me. I was aiming for greater parity, but we can always re-add easily enough if the people demand it.

PR #46, opened 2021-12-03
Improv/media code sharing

cjpillsbury:
WIP but providing as a draft for early review & pokin'


packages/mux-video/examples/async-play.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
```
cjpillsbury, line 1: ignore all of these fresh examples. They are local test cases I've set up that got unintentionally pushed by some coarse-grained adds.

dylanjha (approved):
Good abstraction with playback-core + the addition of startTime

PR #50, opened 2021-12-14
build/remove-deps: Remove dependencies on mux-embed and hls.js from e…

cjpillsbury:
…lements. Provide relevant type exports from playback-core.

Per Dylan's callout, removing and refactoring to reduce cross-package dependency version management.

To test, you should be able to do a from scratch install and build of the workspace as a whole and then run any/all of our examples projects.


dylanjha (approved):
LGTM, I'm infavor of the PlaybackEngine abstraction here that is contained in playback-core

PR #53, opened 2022-01-06
Make Lerna work in independent mode to publish pks

luwes:
This will make it easier to publish packages separately, each package will have its own version.
See https://github.com/lerna/lerna#independent-mode

Added 1 npm script in the root `yarn release`, this will call `lerna publish` internally 
https://github.com/lerna/lerna/tree/main/commands/publish#readme

Going from the release step list here
https://github.com/muxinc/elements/blob/main/README_INTERNAL.md

this becomes:

1. **yarn release**
   Lerna will detect changed files of the packages and prompt to cut a new release (patch,minor,major,etc) of that package
   *build* of package will automatically be done by the "prepublishOnly" script, *git tag* is also done automatically
2. **git push && git push --tags**
3. Edit the pushed git tag to release in GitHub and write a description

### DIFF
--- a/README_INTERNAL.md
+++ b/README_INTERNAL.md
@@ -1,5 +1,5 @@
 # Releasing
 
-1. `yarn release`
+1. From the workspace root directory, run `yarn release`
 2. `git push && git push origin --tags`
 3. Create a [new release in GitHub](https://github.com/muxinc/elements/releases) and write a description
--- a/README_INTERNAL.md
+++ b/README_INTERNAL.md
@@ -1,7 +1,5 @@
 # Releasing
 
-1. Bump version in the `package.json`
-1. Rebase and merge changes into `main`
-1. Tag `@mux-elements/{package-name}@{version}` -- for example: `git tag @mux-elements/mux-video@0.1.1 && git push origin --tags`
-1. From the package directory (For example: `packages/mux-video` run `yarn build && npm publish`
-1. Create a new release in GitHub and write a description
+1. `yarn release`
+2. `git push && git push origin --tags`
+3. Create a [new release in GitHub](https://github.com/muxinc/elements/releases) and write a description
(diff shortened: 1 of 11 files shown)


cjpillsbury:
@luwes This one may merit a chat. Want to make sure we aren't going to have any issues under a few circumstances, but if those are good, I'm pumped about this.

luwes:
Screen Shot 2022-01-07 at 4 44 05 PM

Looks like it's working well with adding a new package, `yarn install` makes the symlink.

cjpillsbury:
> Looks like it's working well with adding a new package, `yarn install` makes the symlink. Added `mux-preview` as an example.

wooohooo! I say 🚢 🇮🇹 (ship it)

cjpillsbury:
Actually one last ask before merging this. Let's update https://github.com/muxinc/elements/blob/main/README_INTERNAL.md accordingly. Otherwise, 😎

cjpillsbury (changes requested):
Just update the publishing readme and consider it pre-approved for rebase+merging

README_INTERNAL.md
```diff
@@ -1,7 +1,5 @@
 # Releasing
 
-1. Bump version in the `package.json`
-1. Rebase and merge changes into `main`
-1. Tag `@mux-elements/{package-name}@{version}` -- for example: `git tag @mux-elements/mux-video@0.1.1 && git push origin --tags`
-1. From the package directory (For example: `packages/mux-video` run `yarn build && npm publish`
-1. Create a new release in GitHub and write a description
+1. `yarn release`
```
cjpillsbury, line 3: nit: 1. From the workspace root directory, run `yarn release`

cjpillsbury (changes requested):
one small nit on docs.

cjpillsbury (approved):

PR #54, opened 2022-01-06
Add `yarn dev` command in root

luwes:
Adds a simple `yarn dev` command that runs all the packages in dev including example sites.


package.json
```diff
@@ -22,6 +22,7 @@
     "examples/*"
   ],
   "scripts": {
+    "dev": "lerna run dev --parallel",
```
cjpillsbury, line 25: iirc can't use --parallel and maintain the dependency graph order of build 😢

examples/create-react-app-with-typescript/package.json
```diff
@@ -21,6 +21,7 @@
     "web-vitals": "^1.0.1"
   },
   "scripts": {
+    "dev": "PORT=3001 BROWSER=none yarn start",
```
cjpillsbury, line 24: this might get tricky for the examples in dev mode... how much smoke testing has been done here to make sure it's doing what we'd think/want, especially in live rebuilds/hmr cases across the project?

I think if we're going to add a global dev script for the workspace, we may want to filter it to only run on the `/packages` packages (at least for now). Most frequently we probably aren't going to want all of the example apps running and rebuilding concurrently, especially as the list and size of example apps grows.

There isn't an out of box solution to scope by directory, but there's an option issue and discussion here with some potential ideas to solve: https://github.com/lerna/lerna/issues/1840

Alternatively, you could simply use `--scope` and explicitly enumerate the scope for the time being (See: https://github.com/lerna/lerna/tree/main/commands/run#options)

cjpillsbury (changes requested):

examples/create-react-app-with-typescript/package.json
luwes: ended up using `--ignore demo-*` that could capture future demo sites as well to ignore.

luwes:
@cjpillsbury this should be working I think. I started with a fresh clone, yarn install and yarn dev is watching the files and building the `/packages/...` only.

cjpillsbury (approved):

PR #55, opened 2022-01-10
Fix esbuild --watch bailing

luwes:
See solution here https://github.com/evanw/esbuild/issues/1885#issuecomment-1000866366


cjpillsbury (approved):
🏄‍♂️

PR #56, opened 2022-01-10
Feat/mux player react

cjpillsbury:
mux player react with some legs. Have been testing using our `nextjs-with-typescript` app.


luwes:
Could a click on the video area act as play/pause? I think many users will expect this to work.

luwes:
On mobile 320px the UI overlaps and looks a bit unfinished.
Screen Shot 2022-01-11 at 9 56 17 AM

luwes:
should volume range be disabled for mobile with volume hardware buttons?

cjpillsbury:
> should volume range be disabled for mobile with volume hardware buttons?

yeah. It's not part of our beta checklist, but I think for sure yes.

cjpillsbury:
> On mobile 320px the UI overlaps and looks a bit unfinished.

Correct. This is still very much a WIP. The current goal is something that has all the core moving pieces so we can get a release out and start testing real integration while polishing.

luwes:
@cjpillsbury just noticed that these player builds are pulling in hls.js and mux-data twice via `@mux-elements/playback-core` and `@mux-elements/mux-video`. bundle is around 1MB, should be about half. 

could solve it in 2 ways, either only import from `@mux-elements/mux-video` or make tree shaking work correctly by importing the esm versions. will probably need a `"module"` key in `package.json` of the core packages

PR #57, opened 2022-01-13
Fix mux-video types in dist

luwes:
This change makes the build:types script add the missing types of `CustomVideoElement` by copying them from the `src` folder to the `dist/types` folder.

Related question
https://stackoverflow.com/questions/56018167/typescript-does-not-copy-d-ts-files-to-build


cjpillsbury (approved):
👌 . I think a good followup would be for us to rework this into its own package for "`CustomMediaElement`" with audio/video permutations in typescript and then try to upstream it to the "real" CustomVideoElement somehow https://www.npmjs.com/package/custom-video-element

PR #59, opened 2022-01-13
Add minimal mux-player

luwes:
Posting this to quickly get feedback on inheritance vs composition?

It seems for the custom element it would be easier to inherit from `MuxVideoElement` because it will need most if not all API 
methods from `MuxVideoElement`.


cjpillsbury:
I think this won't be a viable long term solution, but I also think this is a really clever idea for getting the beta player out the door unless we're worried about (what would most likely be edge cases of) backwards compat/consistency. Since switching from your inheritance approach to the compositional approach would primarily be additive wrt the code base, I say run with it and we can get some input from other folks before any final commitment for the beta version. 🚀 😎  (cc: @heff @mmcc)

High overview of why this likely won't be a viable long term:
As we expand functionality for Media Chrome, the slotted media element will be expected to (optionally) "advertise"/"provide" a superset of an `HTMLMediaElement` instance's interface, such as things only an "HTTP Adaptive Streaming" (HAS) media element would have. 

Here's a concrete and likely example:
* `<mux-video/>` (and any other HAS-supporting) media element would have property(/ies), attributes, and/or events for advertising the list of available video renditions (bitrates, resolutions, etc) and alternate audio renditions (e.g. alternate languages, descriptive audio, etc), as well as some interface for getting/setting the currently selected/playing audio/video rendition
* the `<media-controller/>` in Media Chrome would expect to get and set these things on the slotted media element, if available (just like most media state values)
* Media Chrome would support UI components for responding to/updating the aforementioned state (just like other Media Chrome components/Media State Receivers)

The example above assumes a compositional architecture or a whole bunch of code acrobatics (since `<mux-player/>` has its `<media-controller/>` as a descendant but its slotted `media` would need to be something that conforms to the above). Alternatively, we could rethink/rework/decompose some assumptions in how the media state source(s) are identified/defined in Media Chrome to make it less presumptuous, which I'm also not averse to, though that would be a larger conversation and a much bigger decision.

luwes:
Actually that makes a lot of sense, there's a lof of what ifs and composition would be more robust in the long run but it will take some more code initially. Probably also a good thing the "public" API will be methods we opt-in instead of exposing the whole HTMLMediaElement API.

cjpillsbury:
> Probably also a good thing the "public" API will be methods we opt-in instead of exposing the whole HTMLMediaElement API.

Though we could solve these problems via "opt out" instead of "opt in" using various "masking" strategies in an inheritance model

packages/mux-player/src/common/little-vdom.js
```diff
@@ -0,0 +1,158 @@
+// Adapted and fixed bugs from little-vdom.js
+// https://gist.github.com/developit/2038b141b31287faa663f410b6649a87
+// https://gist.github.com/marvinhagemeister/8950b1032d67918d21950b3985259d78
+// Added refs, style maps
```
luwes, line 4: Added a tiny JSX VDOM lib to keep the views similar between players

luwes:
I think it's gonna be a challenge to have 2 separate players and keeping them to behave the same.
`mux-video` is small enough to do that but it will be harder and harder as the player grows.

this led me to using the JSX views here, we could even add hooks and keep that the same too.
the hooks logic can be really small, less than a 1kB.

let me know your thoughts on this.

packages/mux-player/src/index.jsx
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+  constructor() {
+    super();
+
+    this.attachShadow({ mode: "open" });
+    renderMuxPlayer(this);
+
+    // Initialize all the attribute properties
+    Array.prototype.forEach.call(this.attributes, (attrNode) => {
```
cjpillsbury, line 358: I think at least some of this should probably be moved to `connectedCallback`.

packages/mux-player/src/index.jsx
```diff
@@ -0,0 +1,413 @@
...
+      this.attributeChangedCallback(attrNode.name, null, attrNode.value);
+    });
+
+    if (this.video) this.video.hls.config.maxMaxBufferLength = 2;
```
cjpillsbury, line 362: what's going on here?

packages/mux-player/src/index.jsx
```diff
@@ -0,0 +1,413 @@
...
+  }
+
+  get video() {
+    return this.shadowRoot.querySelector("mux-video");
+  }
+
+  getVideoAttribute(name) {
+    return this.video ? this.video.getAttribute(name) : this.getAttribute(name);
```
cjpillsbury, line 370: this feels a bit brittle. Also since this isn't part of the public api, I'd recommend breaking it out into a separate function, a la:

`const getVideoAttribute = (el, name) => el.video ? el.video.getAttribute(name) : el.getAttribute(name);`

packages/mux-player/src/index.jsx
```diff
@@ -0,0 +1,413 @@
...
+    return this.getVideoAttribute(VideoAttributes.STREAM_TYPE);
+  }
+
+  get envKey() {
+    return this.getVideoAttribute(VideoAttributes.ENV_KEY);
+  }
+
+  attributeChangedCallback(attrName, oldValue, newValue) {
+    if (VideoAttributeNameValues.includes(attrName)) {
+      this.video?.attributeChangedCallback(attrName, oldValue, newValue);
```
cjpillsbury, line 395: I think you want this to be `this.video.setAttribute(attrName, newValue);`

cjpillsbury: Also I think this will have edge cases if you're concerned about this.video not existing.

packages/mux-player/src/index.jsx
```diff
@@ -0,0 +1,413 @@
...
+      </MediaControlBar>
+    </>
+  );
+};
+
+export const LiveChromeLarge = (props) => {
+  const { onAirPlaySelected, hasAirPlay = false } = props;
+  return (
+    <>
+      <MediaControlBar>
```
cjpillsbury, line 162: fine for now but I think these should probably look more like web-component usage (even if they're being rendered w/the light vdom). We can discuss.

packages/mux-video/src/CustomVideoElement.js
```diff
@@ -72,9 +70,10 @@ class CustomVideoElement extends HTMLElement {
         if (mutation.type === "childList") {
           // Child being removed
           mutation.removedNodes.forEach((node) => {
-            this.nativeEl.removeChild(
-              this.nativeEl.querySelector(`track[src="${node.src}"]`)
+            const track = this.nativeEl.querySelector(
+              `track[src="${node.src}"]`
             );
+            if (track) this.nativeEl.removeChild(track);
```
cjpillsbury, line 76: 👍 we'll probably need to account for this overly-presumptuous code better, but the if seems reasonable, given the presumptuousness.

packages/mux-player/src/index.jsx
luwes: this is just temporary to save some bandwidth every time I refresh, hls.js loads 60MB on page load

PR #63, opened 2022-01-26
Add prettier config

luwes:
Adds a prettier config in the root for all packages.

The only difference with media-chrome's is the `singleQuote: false`
https://github.com/muxinc/media-chrome/blob/main/.prettierrc

I guess those are the defaults used by the `pretty-quick` package that runs prettier in the git pre-commit hook.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,8 +1,8 @@
 ---
-name: "🐛 Bug Report"
+name: '🐛 Bug Report'
 about: Report a reproducible bug or regression.
-title: "Bug: "
-labels: "bug, Status: Unconfirmed"
+title: 'Bug: '
+labels: 'bug, Status: Unconfirmed'
 ---
 
 <!--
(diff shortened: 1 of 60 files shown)


luwes:
should we also move to single quotes here? since it will be org wide
@cjpillsbury @mmcc @heff

mmcc:
Definitely single quotes imo, we should align these as closely as possible to the prettier configs we use across projects.

gkatsev:
But is it single quotes always, or single quotes except when you'll be required to escape them? I'd prefer the latter over the former.

cjpillsbury:
When I'm writing code, I tend to use single quotes, but I have 0 investment in lobbying for any of these sorts of details as long as it results in consistency and lends itself to tooling/automation support.

mmcc:
Yeah, but ultimately the beautiful thing here is you can write things however you want in the moment and just let the formatter take care of it on the backend. One of the reasons I push hard for it now is because I've become a total and complete slob when actually writing code assuming that the formatter will fix my indention crimes and such.

I did, however, double check this, and Prettier does indeed use double quotes when there's a single quote contained within ❤️

luwes:
Updated to single quote ⚡

dylanjha:
Are we ready to pretty-ify elements?

cjpillsbury (approved):

PR #64, opened 2022-01-26
improv/playback-core-android-and-tracks: preferMSE for Android for ll…

cjpillsbury:
…-hls content (due to current poor perf natively). Handle async track adds due to timing edge cases in hls.js.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -175,10 +175,16 @@ export const setupHls = (
 
   const canUseNative = !type || (mediaEl?.canPlayType(type) ?? true);
   const hlsSupported = Hls.isSupported();
+  // NOTE: Native HLS playback on Android for LL-HLS has been flaky, so we're prefering
+  // MSE for those conditions for now. (CJP)
+  const userAgentStr = window?.navigator?.userAgent ?? "";
+  const isAndroid = userAgentStr.toLowerCase().indexOf("android") !== -1;
+  const defaultPreferMse = isAndroid && streamType === StreamTypes.LL_LIVE;
 
   // We should use native playback for hls media sources if we a) can use native playback and don't also b) prefer to use MSE/hls.js if/when it's supported
   const shouldUseNative =
-    !hlsType || (canUseNative && !(preferMse && hlsSupported));
+    !hlsType ||
+    (canUseNative && !((preferMse || defaultPreferMse) && hlsSupported));
 
   // 1. if we are trying to play an hls media source create hls if we should be using it "under the hood"
   if (hlsType && !shouldUseNative && hlsSupported) {
@@ -243,7 +249,10 @@ export const setupMux = (
 
 export const loadMedia = (
   props: Partial<
-    Pick<MuxMediaProps, "preferMse" | "src" | "type" | "startTime">
+    Pick<
+      MuxMediaProps,
+      "preferMse" | "src" | "type" | "startTime" | "streamType"
+    >
   >,
   mediaEl?: HTMLMediaElement | null,
   hls?: Pick<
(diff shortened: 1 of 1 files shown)


packages/playback-core/src/index.ts
```diff
@@ -260,16 +269,22 @@ export const loadMedia = (
...
   const type = getType(props);
   const hlsType = type === ExtensionMimeTypeMap.M3U8;
 
   const canUseNative = !type || (mediaEl?.canPlayType(type) ?? true);
   const hlsSupported = Hls.isSupported();
+  const userAgentStr = window?.navigator?.userAgent ?? "";
+  // NOTE: Native HLS playback on Android for LL-HLS has been flaky, so we're prefering
+  // MSE for those conditions for now. (CJP)
+  const isAndroid = userAgentStr.toLowerCase().indexOf("android") !== -1;
+  const defaultPreferMse = isAndroid && streamType === StreamTypes.LL_LIVE;
```
luwes, line 282: maybe in the future we could also look into iPadOS>13 MSE performance vs native HLS?

also wondering if there's a difference between Android tablet vs Android mobile but probably not.

luwes (approved):
LGTM

packages/playback-core/src/index.ts
```diff
@@ -175,10 +175,16 @@ export const setupHls = (
 
   const canUseNative = !type || (mediaEl?.canPlayType(type) ?? true);
   const hlsSupported = Hls.isSupported();
+  // NOTE: Native HLS playback on Android for LL-HLS has been flaky, so we're prefering
```
gkatsev, line 178: Has it been ok for regular HLS? I've always found it to be severely lacking if you're doing more than just hitting play. Or is the idea here to leave it up to user-choice, except when we know things are bad?

packages/playback-core/src/index.ts
gkatsev: There shouldn't be really any difference between android tables and android phones, at least, not beyond regular differences in hardware/manufacturer.

packages/playback-core/src/index.ts
```diff
@@ -318,6 +333,21 @@ export const loadMedia = (
...
+        if (track.getAttribute("label") === "thumbnails") {
+          const thumbnailsSrc = track.getAttribute("src");
+          if (!thumbnailsSrc) return;
+          track.removeAttribute("src");
+          // NOTE: This is super hacky. Need to hunt down if there's a
+          // better event to ensure this happens late enough to not
+          // be stomped on by hls.js (CJP)
+          setTimeout(() => {
+            track.setAttribute("src", thumbnailsSrc);
+          }, 500);
```
gkatsev, line 347: question: Why is this necessary?

packages/playback-core/src/index.ts
cjpillsbury: it's been "reasonable" iirc. Since users can configure `preferMse` via a single attr/prop, I believe we're erring on the side of "use native unless there's strong reasons not to" whenever they don't set that prop.

packages/playback-core/src/index.ts
cjpillsbury: yeah for now we're not forcing non-native by default unless the native playback is janked-up or doesn't support it (aka android for ll-hls, unless/until they fix it).

packages/playback-core/src/index.ts
cjpillsbury: hls really wants to manage exactly one "selected track" and does silly things asynchronously when you provide "out of band" tracks via `<track/>`. It's also especially not great for metadata text tracks (or chapters etc.) unless it "knows about them" and built them itself. tldr - we end up with no `activeCues` because of disable cases and simply re-setting `src` after hls is done with its initial crud (see event deets in comment) was the most reliable way of getting them loaded (moreso than mucking around with the `mode`)

PR #67, opened 2022-01-31
chore/update-release: update release command with --no-private

dylanjha:


cjpillsbury (approved):

dylanjha:
@luwes wow 🥳 

I just published all 5 packages with Lerna and that was a breeze

Just this one small change, adding `--no-private` so that the "packages" in examples/ get ignored (without this flag lerna was attempting to publish the examples)

PR #68, opened 2022-02-03
fix/thumbnails-with-hls-bug: update attempt to make thumbnails play n…

cjpillsbury:
…icely with hls.js to be more tied to specific events.


luwes (approved):
LGTM

PR #69, opened 2022-02-07
Add type to mux-video for playback-core

luwes:
Adds the `type` as a property. 

I ran into an issue with an m3u8 without an extension and tried this attribute but it didn't work because the prop was not on the element.


cjpillsbury (approved):
👌

Issue #66, opened 2022-01-29
DOM property to attribute reflection


luwes:
Noticed this today, native HTML element's string properties that receive a `null` or `undefined` are cast as string and still set as a string to the attribute. The Mux elements behave differently as seen below, removing the attribute. 

https://github.com/muxinc/elements/blob/675fc314751b667e29c25d621bb08cadc5585f58/packages/mux-video/src/index.ts#L97-L107

cjpillsbury:
This was intentional, but maybe it shouldn't have been. To be consistent, it looks like we'd need to rely on `src = ''`. Would need to smoke test to see if there are any "ripple effects"/assumptions elsewhere, including playback-core, but I'm amenable to the refactor. This feels like it's part of the larger conversation of "should `<mux-video/>` be as close as possible to the API/behaviors of `<video/>`, or should we diverge ever?"

PR #72, opened 2022-02-14
Fix removing attributes, add preload support

luwes:
This change fixes the issue of removing attributes from mux-player being forwarded and also removed from mux-video.


packages/mux-player/test/player.test.js
```diff
@@ -82,4 +82,50 @@ describe("<mux-player>", () => {
...
+      player.setAttribute(attrName, attrValue === true ? "" : attrValue);
+      assert(muxVideo.hasAttribute(attrName), `has ${attrName} attr added`);
+
+      if (!["playsinline"].includes(attrName)) {
+        assert.equal(
+          muxVideo[attrName],
+          attrValue,
+          `has true ${attrName} prop`
+        );
+      }
```
gkatsev, line 125: instead of not checking `playsInline` should be check for that in the attrName?

```suggestion
      assert.equal(
        muxVideo[attrName === 'playsinline' ? 'playsInline' : attrName],
        attrValue,
        `has true ${attrName} prop`
      );
```

gkatsev (commented):
This fixes #66, right?

packages/mux-player/test/player.test.js
luwes: it's not #66 actually, it's the issue @cjpillsbury brought up with removing attrs not having any effect.

but good call, maybe it's best to fix @66 in this PR too. this is about that mux-video is removing attributes when it shouldn't. only boolean attributes can be removed by changing the associated property to false. at least if we want to keep the setters / getters behave more like native elements.

luwes: I'm going to do the playsInline a bit different because the `.toLowerCase()` is there. good suggestion!

gkatsev: Ah, cool. Didn't quite get that this change was different from #66. But yeah, it seems related enough to fix it now.

packages/mux-player/test/player.test.js
```diff
@@ -82,4 +82,43 @@ describe("<mux-player>", () => {
...
+      crossOrigin: "",
+      preload: "metadata",
+      poster:
+        "https://image.mux.com/xLGf7y8cRquv7QXoDB02zEe6centwKfVmUOiPSY02JhCE/thumbnail.jpg?time=0",
+      src: "https://stream.mux.com/r4rOE02cc95tbe3I00302nlrHfT023Q3IedFJW029w018KxZA.m3u8",
+    };
+
+    for (let propName in checkProps) {
+      const attrName = propName.toLowerCase();
+      const attrValue = checkProps[attrName];
```
gkatsev, line 114: Should this use `propName` for getting the value?

luwes: wow, good catch. will fix this.

gkatsev (approved):

package.json
```diff
@@ -23,6 +23,7 @@
     "scripts/*"
   ],
   "scripts": {
+    "test": "lerna run test --ignore demo-*",
```
cjpillsbury, line 26: 😎

.github/workflows/ci.yml
```diff
@@ -18,3 +18,4 @@ jobs:
           node-version: ${{ steps.get-versions.outputs.node }}
       - run: yarn install --frozen-lockfile
       - run: yarn build
+      - run: yarn test
```
cjpillsbury, line 21: 😎

packages/mux-player/test/player.test.js
```diff
@@ -82,4 +82,44 @@ describe("<mux-player>", () => {
...
+        "https://image.mux.com/xLGf7y8cRquv7QXoDB02zEe6centwKfVmUOiPSY02JhCE/thumbnail.jpg?time=0",
+      src: "https://stream.mux.com/r4rOE02cc95tbe3I00302nlrHfT023Q3IedFJW029w018KxZA.m3u8",
+    };
+
+    for (let propName in checkProps) {
+      const attrName = propName.toLowerCase();
+      const value = checkProps[propName];
+
+      player.setAttribute(attrName, value === true ? "" : value);
+      assert(muxVideo.hasAttribute(attrName), `has ${attrName} attr added`);
```
cjpillsbury, line 117: non-blocker/fine for now: Generally, it's better to break these up into separate tests, since the first failure will cause the entire test to fail, meaning we won't have visibility on other failures. While that's most likely fine for our current impl, this test structure isn't guaranteed to be resilient to change.

cjpillsbury (changes requested):
One ask (may be fine), otherwise good to ship.

packages/mux-player/src/video-api.ts
```diff
@@ -78,7 +79,11 @@ class VideoApiElement extends HTMLElement {
       AllowedVideoAttributeNames.includes(attrName) &&
       this.video?.getAttribute(attrName) != newValue
     ) {
-      this.video?.setAttribute(attrName, newValue);
+      if (newValue === null) {
+        this.video?.removeAttribute(attrName);
+      } else {
+        this.video?.setAttribute(attrName, newValue);
+      }
```
luwes, line 86: @cjpillsbury you're right, it's here ^^

package.json
cjpillsbury: non-blocking: Per #75 may want to consider switching to `scope` instead of `ignore`

cjpillsbury (approved):
One (non-blocking) callout, otherwise 👍

PR #73, opened 2022-02-14
Add mux-player w/ @github/template-parts

luwes:
(Re-posting from https://github.com/luwes/elements/pull/1)

After discovering this library https://github.com/github/template-parts and then the custom lit-html like implementation on top of the [template-parts spec](https://github.com/WICG/webcomponents/blob/159b1600bab02fe9cd794825440a98537d53b389/proposals/Template-Instantiation.md) which is coded up pretty much all in this file, https://github.com/github/jtml/blob/main/src/html.ts I was curious to see if this would not be a better alternative to the custom DOM utilities I wrote in the mux-player WC. 

I quickly adapted some of the code of [`jtml`](https://github.com/github/jtml/blob/main/src/html.ts) and implemented it in our template code this morning. After using it I think it *is* a better alternative and worth the small overhead of roughly 3kB unzipped.

## Pros

- it's fully declarative, no need for getting any references to the DOM nodes and its subtree and rendering them w/ `RenderableFragment.render()`
- support for rendering attributes after template initialization, no need for node.setAttribute() etc.
- toggling boolean attributes, a value of `false` removes the attribute from the DOM
- our UI implementation code feels more familiar to UI code done with like Lit or React but `@github/template-parts` is the only dependency which is based on a spec which wouldn't change much
- we build the UI lib on top and have control over the `TemplatePart` processors, a `TemplatePart` is either a `NodeTemplatePart` or a `AttributeTemplatePart`. These parts are the holes/expressions of the template literals and can be handled like you wish. Handle iterables, Promises, etc.

## Cons

- it's more code, it adds roughly 3kB unzipped to the bundle of ~520kB. negligible IMO.
- it adds a dependency to [@github/template-parts](https://github.com/github/template-parts)

Feel free to add to the pros / cons list and let me know what you think.


packages/mux-player/src/html.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (value instanceof DocumentFragment && part instanceof NodeTemplatePart) {
+    if (value.childNodes.length) part.replace(...value.childNodes);
+    return true;
+  }
+  return false;
+}
+
+export function processPropertyIdentity(
+  part: TemplatePart,
+  value: unknown
```
cjpillsbury, line 27: non-blocking: We should be able to use `<T = any>processPropertyIdentity(part: TemplatePart, value: T)` to make this a little better with types, especially if we ever get into a case where we want to handle different types differently.

packages/mux-player/src/html.ts
```diff
@@ -0,0 +1,122 @@
...
+): boolean {
+  if (
+    typeof value === "boolean" &&
+    part instanceof AttributeTemplatePart
+    // can't use this because on custom elements the props are always undefined
+    // typeof part.element[part.attributeName as keyof Element] === 'boolean'
+  ) {
+    part.booleanValue = value;
+    return true;
```
cjpillsbury, line 43: non-blocking: We should be able to model these as `is` type predicates (See: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

packages/mux-player/src/template.ts
```diff
@@ -34,62 +33,46 @@ export const template = (props: MuxTemplateProps) => html`
...
-        : ""}
-      ${props.metadata?.viewer_user_id
-        ? `metadata-viewer-user-id="${props.metadata.viewer_user_id}"`
-        : ""}
+      props.streamType === StreamTypes.LL_LIVE}"
+      muted="${props.muted}"
+      debug="${props.debug}"
+      prefer-mse="${props.preferMse}"
+      start-time="${!!props.startTime ? props.startTime : false}"
+      poster="${!!props.poster
```
cjpillsbury, line 43: non-blocking: I think this may be wrong, but I may have a misunderstanding of our poster requirements for alpha/beta. I thought we were trying to support user-provided posters. @dylanjha for visibility.

packages/mux-player/src/template.ts
```diff
@@ -34,62 +33,46 @@ export const template = (props: MuxTemplateProps) => html`
...
+      stream-type="${props.streamType ?? false}"
+      metadata-video-id="${props.metadata?.video_id ?? false}"
+      metadata-video-title="${props.metadata?.video_title ?? false}"
+      metadata-viewer-user-id="${props.metadata?.viewer_user_id ?? false}"
     >
       ${props.playbackId &&
       props.streamType !== StreamTypes.LIVE &&
       props.streamType !== StreamTypes.LL_LIVE
-        ? `<track
+        ? html`<track
```
cjpillsbury, line 56: non-blocking: Here's a more egregious example of what I was discussing re: being able to support template strings that I think we'll definitely want to support longer term.

packages/mux-player/src/template.ts
```diff
@@ -108,66 +91,58 @@ export const ChromeRenderer = (props: MuxTemplateProps) => {
...
-      props
-    )}
-    ${renderable(
-      "airplayButton",
-      ({ supportsAirPlay }: Partial<MuxTemplateProps>) =>
-        supportsAirPlay && MediaAirplayButton(),
-      props
-    )}
-    ${MediaPipButton()}
+    ${props.hasCaptions ? MediaCaptionsButton(props) : ""}
```
cjpillsbury, line 144: non-blocking: This also feels a little clunky, but I think I'm more ok with it than other callouts ("render a Node/set of nodes if some condition is met, otherwise render an empty string" seems reasonable)

packages/mux-video/src/index.ts
```diff
@@ -236,7 +236,11 @@ class MuxVideoElement
   }
 
   get metadata() {
-    return this.__metadata;
+    return Object.assign(this.__metadata, {
+      video_id: this.getAttribute(Attributes.METADATA_VIDEO_ID),
```
cjpillsbury, line 240: blocking: This feels a bit out of scope, but also the refactor feels brittle. I think for each of these we need to minimally check `hasAttribute` so that we don't unintentionally overwrite these key/value pairs in `metadata`

cjpillsbury (changes requested):
Only a couple of small blocking callouts, otherwise 👌

packages/mux-player/src/template.ts
luwes: we could also make it that all `false` values convert to an empty string in a node part. 
the case we need to print `false` is probably very low and then casting it to a string will work.

packages/mux-video/src/index.ts
luwes: good catch, we should do some checks here and determine which has precedence.
or maybe update `this.__metadata[key]` in the attributeChangedCallback, that keeps the order I think.

cjpillsbury (approved):

packages/mux-player/src/template.ts
luwes: oh snap, messed this up. will add a follow up PR!

PR #74, opened 2022-02-14
chore: open-process package to make bin work

gkatsev:
From what I can tell, `bin` scripts are only available when a package is installed globally or when it's installed as a dependency. During development, bin scripts can only be made available via `npm link`. Ideally, `npm link` isn't necessary. So, make `open-process` a dependency that will install locally and make the bin be available for other packages to use.


luwes (approved):
👍

PR #75, opened 2022-02-14
chore: yarn dev shouldn't minify scripts

gkatsev:
When running `yarn dev`, we shouldn't minify. This should decrease built-times but also make debugging easier, whether source maps are enabled or not.
`yarn build` should continue to minify the build output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "scripts/*"
   ],
   "scripts": {
-    "dev": "lerna run dev --parallel --ignore demo-*",
+    "dev": "lerna run dev --parallel",
     "build": "lerna run build",
     "prepare": "husky install && yarn build --ignore demo-*",
     "release": "lerna publish --no-private"
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "scripts/*"
   ],
   "scripts": {
-    "dev": "lerna run dev --parallel --ignore demo-*",
+    "dev": "lerna run dev --parallel --scope '@mux-elements/*'",
     "build": "lerna run build",
     "prepare": "husky install && yarn build --ignore demo-*",
     "release": "lerna publish --no-private"
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "scripts/*"
   ],
   "scripts": {
-    "dev": "lerna run dev --parallel",
+    "dev": "lerna run dev --parallel --ignore demo-*",
     "build": "lerna run build",
     "prepare": "husky install && yarn build --ignore demo-*",
     "release": "lerna publish --no-private"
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "scripts/*"
   ],
   "scripts": {
-    "dev": "lerna run dev --parallel --ignore demo-*",
+    "dev": "lerna run dev --parallel",
```
luwes, line 26: I'm ok with this. @cjpillsbury what was the reason again we ignored the examples?

packages/mux-player/src/template.ts
```diff
@@ -77,6 +77,7 @@ export const template = (props: MuxTemplateProps) => html`
       no-auto-hide
     ></media-loading-indicator>
     ${renderable("chromeRenderer", ChromeRenderer, props)}
+    <div class="foo baz"></div>
```
luwes, line 80: remove when merging

packages/mux-audio-react/package.json
```diff
@@ -16,9 +16,9 @@
     "dev:cjs": "open-process | yarn build:cjs --watch",
     "dev:types": "yarn build:types -w",
     "dev": "npm-run-all --parallel dev:types dev:cjs",
-    "build:cjs": "esbuild src/index.tsx --target=es2019 --minify --bundle --sourcemap --format=cjs --outdir=dist --external:react --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
+    "build:cjs": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --format=cjs --outdir=dist --external:react --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:types": "tsc --declaration --emitDeclarationOnly --outDir './dist/types'",
-    "build": "npm-run-all --parallel build:types build:cjs",
+    "build": "npm-run-all --parallel build:types 'build:cjs --minify'",
```
cjpillsbury, line 21: 👌

package.json
```diff
@@ -23,7 +23,7 @@
     "scripts/*"
   ],
   "scripts": {
-    "dev": "lerna run dev --parallel --ignore demo-*",
+    "dev": "lerna run dev --parallel --scope '@mux-elements/*'",
```
gkatsev, line 26: `--scope` is sort of the opposite of `--ignore` where only matching packages are included. Only the main packages are prefixed with `@mux-elements/*` (and the `/` here is important, otherwise lerna doesn't find them).
I swapped them because it was easier to reason and explain how to include other packages in the README_INTERNAL.md file, since to add demos you would add `--scope 'demo-*'` rather than removing things or swapping things out.

cjpillsbury: Cool I'm down for that as long as it's functionally equivalent with our current setup.

cjpillsbury (approved):

package.json
gkatsev: Yup, it's equivalent!

PR #76, opened 2022-02-15
Feat/mux player react

cjpillsbury:
(Second) First pass impl of MuxPlayer (React)


packages/mux-player-react/src/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  startTime: number;
+  children: never[];
+};
+
+export type MuxPlayerProps = {
+  forwardSeekOffset?: number;
+  backwardSeekOffset?: number;
+  metadataVideoId?: string;
+  metadataVideoTitle?: string;
+  metadataViewerUserId?: string;
```
dylanjha, line 49: In React world, I think it makes sense to handle `metadata` like we did in `<MuxVideo />`:

```jsx
        metadata={{
          video_id: "video-id-123456",
          video_title: "Super Interesting Video",
          viewer_user_id: "user-id-bc-789",
        }}
```

Let me know if that makes sense

packages/mux-player-react/src/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+    children
+  );
+});
+
+const useEventCallbackEffect = (
+  type: string,
+  ref: // | ((instance: EventTarget | null) => void)
+  React.MutableRefObject<EventTarget | null> | null | undefined,
+  callback: EventListener | undefined
+) => {
```
dylanjha, line 89: nice 👍🏼  looks like a clean way to handle the event listener binding / cleanup

dylanjha (commented):
LGTM besides one Mux Data metadata related question. Would appreciate a 2nd review from @luwes

packages/mux-player-react/src/index.tsx
cjpillsbury: Yup definitely does. We can let users let us know if they'd prefer to also have the shorthand and introduce it if they do.

packages/mux-player-react/src/common/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+  propValue: any
+): string | undefined => {
+  if (isKeyOf(propName, ReactPropToAttrNameMap))
+    return ReactPropToAttrNameMap[propName];
+  if (typeof propValue == undefined) return undefined;
+  if (typeof propValue === "boolean" && !propValue) return undefined;
+  if (/[A-Z]/.test(propName)) return toKebabCase(propName);
+  return propName;
+};
+export const toStyleAttr = <T>(x: T) => x;
```
luwes, line 33: is this needed?

packages/mux-player-react/src/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  currentTime: number;
+  volume: number;
+  paused: boolean;
+  src: string | null;
+  poster: string;
+  playbackRate: number;
+  crossOrigin: string;
+  autoPlay: boolean;
+  loop: boolean;
+  muted: boolean;
```
luwes, line 23: should playsinline and preload be added here?

packages/mux-player-react/src/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  useEventCallbackEffect("playerready", ref, onPlayerReady);
+  return [remainingProps];
+};
+
+const MuxPlayer = React.forwardRef<MuxPlayerRefAttributes, MuxPlayerProps>(
+  (props, ref) => {
+    const {
+      /** @TODO Remove these once defaults are added to mux-player (CJP) */
+      forwardSeekOffset = 10,
+      backwardSeekOffset = 10,
```
luwes, line 167: I'll add these asap. yes, can we go with 10s instead of 30s? that feels a bit too much for most videos

gkatsev: completely wild thought: have these as 10s for videos shorter than 5 minutes and 30s for videos longer than 5 minutes?

packages/mux-player-react/src/common/utils.ts
cjpillsbury: I'll re-smoke test but we should be good to drop it (react _**should**_ handle the obj->string conversion 🤞)

packages/mux-player-react/src/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  useObjectPropEffect(
+    "paused",
+    paused,
+    ref,
+    (playerEl: HTMLMediaElement, pausedVal: boolean) => {
+      if (pausedVal) {
+        playerEl.pause();
+      } else {
+        playerEl.play();
+      }
```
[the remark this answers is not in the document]
cjpillsbury: yup we could certainly use `useEffect` directly. Was trying to build something generic for prop updates & event mgmt that we could just use for any/every one of those cases. By using `useObjectPropEffect` we are minimally checking if `el.paused === prop.paused`.

packages/mux-player-react/src/index.tsx
cjpillsbury: For sure. Good callout.

packages/mux-player-react/src/index.tsx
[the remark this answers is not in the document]
cjpillsbury: Since this is a prop, we shouldn't need to "monitor" `playerEl.paused`.

packages/mux-player-react/src/index.tsx
cjpillsbury: actually `preload` should be irrelevant since we rely on hls.js. We should probably remove that from our interface.

packages/mux-player-react/src/index.tsx
cjpillsbury: that would have to be a followup since we'll render the icons before we know the media duration, but cool thought

packages/mux-player-react/src/useElementPropsEffect.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (typeof objVal !== "object") return true;
+  const vKeys = Object.keys(v);
+  const objValKeys = Object.keys(objVal);
+  // NOTE: While very likely fine, it's worth noting that this assumes that all
+  // relevant key/value pairs will be *own* & *enumerable* keys (CJP)
+  if (vKeys.length !== objValKeys.length) return true;
+  return vKeys.some(
+    (vKey) => !objValKeys.includes(vKey) || v[vKey] !== objVal[vKey]
+  );
+};
```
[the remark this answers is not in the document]
luwes: sorry, I missed that it was legacy. I see you found the shared function `shallowEqual` 👍

packages/mux-player-react/src/index.tsx
luwes: not always if native hls / m3u8 is used right?

also a good topic for discussion, I think preload should still behave similarly with hls.js

if the user explicitly sets preload="none" I think it should not start loading any segments in hls.js and we should handle this logic in mux-video / playbackcore with setting hls.js configs.

luwes (approved):
⚡

PR #77, opened 2022-02-15
Add live indicator and fixes to live views

luwes:
Adds the live indicator and fixes some things with the live small and large view. Removed the time display for live.

### DIFF
--- a/packages/mux-player/src/html.ts
+++ b/packages/mux-player/src/html.ts
@@ -46,8 +46,20 @@ export function processBooleanAttribute(
   return false;
 }
 
+export function processBooleanNode(
+  part: TemplatePart,
+  value: unknown
+): boolean {
+  if (value === false && part instanceof NodeTemplatePart) {
+    part.replace("");
+    return true;
+  }
+  return false;
+}
+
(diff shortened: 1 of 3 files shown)


packages/mux-player/src/styles.css
```diff
@@ -89,3 +89,9 @@ media-volume-range {
 media-time-display {
   white-space: nowrap;
 }
+
+.mxp-live-indicator::before {
+  content: "● ";
```
cjpillsbury, line 94: non-blocking: 👍 any sizing weirdness as a result of `::before` instead of including it in the DOM structure?

luwes: not really, did you expect some weirdness with using `::before`?

packages/mux-player/src/html.ts
```diff
@@ -46,8 +46,20 @@ export function processBooleanAttribute(
   return false;
 }
 
+export function processBooleanNode(
```
cjpillsbury, line 49: Assuming this was added for the "false" cases in rendering with the template parts refactor and is unrelated to this PR?

cjpillsbury (approved):
Approved.

PR #78, opened 2022-02-15
feat: add background color to have sufficient contrast on controls

gkatsev:


packages/mux-player/src/template.ts
```diff
@@ -72,6 +72,7 @@ export const template = (props: MuxTemplateProps) => html`
           />`
         : ""}
     </mux-video>
+    <div slot="gestures-chrome" class="mxp-background"></div>
```
gkatsev, line 75: Putting it in the `gestures-chrome` allows it to be in front of the media player but before the controls.

packages/mux-player/src/styles.css
```diff
@@ -30,6 +30,13 @@ media-control-bar [role="switch"] {
   background-color: var(--media-control-background, rgba(20, 20, 30, 0.7));
 }
 
+.mxp-background {
+  height: 100%;
+  width: 100%;
+  position: absolute;
+  background-color: rgba(0, 0, 0, 0.6);
```
gkatsev, line 37: design calls for 0.5, but I realized that we have text in the control bar, the time displays, and at 0.5 our contrast ratio is around 3, where-as with 0.6 we're at 5.74 which is definitely above the 4.5 WCAG requirement for text https://www.w3.org/WAI/WCAG21/Techniques/general/G18.html and the 3 req for icons https://www.w3.org/WAI/WCAG21/Techniques/general/G207.html

packages/mux-player/src/template.ts
gkatsev: went with a css-only solution.

packages/mux-player/src/styles.css
```diff
@@ -30,6 +30,10 @@ media-control-bar [role="switch"] {
   background-color: var(--media-control-background, rgba(20, 20, 30, 0.7));
 }
 
+media-controller:is([media-paused], :not([user-inactive]))::part(vertical-layer) {
```
gkatsev, line 33: target the vertical-layer part, where the controls are, when either the media is paused (and controls are showing) or when we're playing and the user is active and thus the controls are showing.

cjpillsbury: I think we need `centered-layer` too. See: https://github.com/muxinc/media-chrome/blob/main/src/js/media-container.js#L118-L120

cjpillsbury: nevermind, this should be fine! (we don't have a convention on which to use just yet)

cjpillsbury (approved):
small tweak needed I believe

packages/mux-player/src/styles.css
gkatsev: We want this to be as far behind as possible and for now, which should be the vertical layer. Plus, we're not letting mux-player customize controls yet, so, we should end up with both.

luwes:
is there any way this can be animated with the same transition the controls use for hiding?
right now it's an abrupt show / hide, not so smooth

gkatsev:
Good call, let me take a look at what it takes to transition it.

gkatsev:
@luwes updated to transition in and out. The durations were taken from media-chrome's transitions for the controls. Longer term, would probably be nice to make them be css variables or something so that it could be re-used and changed in one place.

packages/mux-player/src/styles.css
```diff
@@ -30,6 +30,15 @@ media-control-bar [role="switch"] {
   background-color: var(--media-control-background, rgba(20, 20, 30, 0.7));
 }
 
+media-controller::part(vertical-layer) {
+  transition: background-color 0.25s;
+}
+
+media-controller:is([media-paused], :not([user-inactive]))::part(vertical-layer) {
+  background-color: rgba(0, 0, 0, 0.6);
+  transition: background-color 1s;
```
cjpillsbury, line 39: 😎  same transition rate as https://github.com/muxinc/media-chrome/blob/main/src/js/media-container.js#L98

cjpillsbury (approved):
re-approved

PR #79, opened 2022-02-15
Fix player poster and default poster

luwes:
Fixes an issue with setting a custom poster image.


cjpillsbury (approved):
Some people/linters feel strongly about chaining ternaries, but I say YOLO. Approved.

PR #81, opened 2022-02-16
Make SVG icons external files

luwes:
This change makes the SVG icons separate files so they can be easily replaced or added in the future. 
Maybe run through SVGO or processed in some way.

They can still receive props via the `template parts` syntax. 
This was still needed for seek forward and backward values until that gets fixed in media-chrome.


cjpillsbury (approved):
🆒

PR #82, opened 2022-02-16
Fix basic css so player doesn't collapse

luwes:
Adds width 100% internally so the player doesn't collapse with no external CSS set.

### DIFF
--- a/packages/mux-player/src/styles.css
+++ b/packages/mux-player/src/styles.css
@@ -2,6 +2,7 @@
   /* default aspect ratio, overwrite w/ e.g. style="aspect-ratio: 16 / 10" */
   aspect-ratio: 16 / 9;
   display: block;
+  width: 100%;
 }
 
 media-controller {


packages/mux-player/src/styles.css
```diff
@@ -2,6 +2,7 @@
   /* default aspect ratio, overwrite w/ e.g. style="aspect-ratio: 16 / 10" */
   aspect-ratio: 16 / 9;
   display: block;
+  width: 100%;
```
cjpillsbury, line 5: non-blocking: not sure that we want the player to automatically grow to 100% of the width of its parent by default, but we can iterate on that.

cjpillsbury (approved):

packages/mux-player/src/styles.css
gkatsev: should we have a min-width instead, then?

luwes: I kinda do prefer it expands like a block element, who's creating a non responsive player nowadays?

PR #84, opened 2022-02-17
Add custom attributes for muted, volume and playbackrate

luwes:
This change adds a `volume` and `playbackrate` attribute that sets the corresponding video property.
Changing the video properties do not reflect back to the attributes.

It also changes the behavior of the `muted` attribute so it reflects to the native video muted property in addition to defaultMuted.


cjpillsbury (approved):
Awesome.

PR #85, opened 2022-02-17
Add default 10s seek offsets

luwes:
Fixes [sc-16003]
Adds the default seek offsets of 10s to the mux-player.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -10,6 +10,7 @@ import {
 } from "./helpers";
 import { template } from "./template";
 import { render } from "./html";
+import { toNumberOrUndefined } from "./utils";
 
 import type { MuxTemplateProps } from "./types";
 import type { Metadata } from "@mux-elements/playback-core";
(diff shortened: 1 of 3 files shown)


packages/mux-player/src/index.ts
```diff
@@ -339,11 +339,11 @@ class MuxPlayerElement extends VideoApiElement {
   }
 
   get forwardSeekOffset() {
-    return this.getAttribute(PlayerAttributes.FORWARD_SEEK_OFFSET);
+    return this.getAttribute(PlayerAttributes.FORWARD_SEEK_OFFSET) ?? 10;
```
cjpillsbury, line 342: nit (non-blocking): we should be consistent with either treating these as strings or numbers. I.e. either:
```js
return this.getAttribute(PlayerAttributes.FORWARD_SEEK_OFFSET) ?? '10';
```
or
```js
if (!this.hasAttribute(PlayerAttributes.FORWARD_SEEK_OFFSET) return 10;
return +this.getAttribute(PlayerAttributes.FORWARD_SEEK_OFFSET);
```
(or similar)

cjpillsbury (approved):
Approved with one callout.

PR #87, opened 2022-02-17
Add ready event to mux-player

luwes:
This adds a `ready` event for the player signaling to the user, the player is ready to accept API requests.

Intentionally doesn't listen to `canplay` or `metadata` (better use those events directly then), 
it's not always sure a `src` is set or video content will be loaded (preload=none).

Maybe one addition could be in case hls.js is used to listen for the manifest loaded event and fire `ready` after that?
But then ideally that would live in playback core if we want to have mux-player not interface directly with the `muxVideo.hls` api.

### DIFF
--- a/packages/mux-player/src/helpers.ts
+++ b/packages/mux-player/src/helpers.ts
@@ -34,10 +34,8 @@ export const hasVolumeSupportAsync = async (
   if (!mediaEl) return false;
   const prevVolume = mediaEl.volume;
   mediaEl.volume = prevVolume / 2 + 0.1;
-  return new Promise<boolean>((resolve, reject) => {
-    setTimeout(() => {
-      resolve(mediaEl.volume !== prevVolume);
-    }, 0);
+  return Promise.resolve().then(() => {
+    return mediaEl.volume !== prevVolume;
   });
 };
 
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -10,7 +10,7 @@ import {
 } from "./helpers";
 import { template } from "./template";
 import { render } from "./html";
-import { toNumberOrUndefined } from "./utils";
+import { toNumberOrUndefined, promisify } from "./utils";
 
 import type { MuxTemplateProps } from "./types";
 import type { Metadata } from "@mux-elements/playback-core";
@@ -59,6 +59,7 @@ class MuxPlayerInternal {
     this._setUpCaptionsButton(el);
     this._setUpAirplayButton(el);
     this._setUpVolumeRange(el);
+    this._setUpPlayerReady(el);
   }
 
   connectedCallback() {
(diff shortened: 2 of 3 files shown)


packages/mux-player/src/helpers.ts
```diff
@@ -34,10 +34,8 @@ export const hasVolumeSupportAsync = async (
   if (!mediaEl) return false;
   const prevVolume = mediaEl.volume;
   mediaEl.volume = prevVolume / 2 + 0.1;
-  return new Promise<boolean>((resolve, reject) => {
-    setTimeout(() => {
-      resolve(mediaEl.volume !== prevVolume);
-    }, 0);
+  return Promise.resolve().then(() => {
```
cjpillsbury, line 37: yup that works, too!

packages/mux-player/src/index.ts
```diff
@@ -217,6 +218,24 @@ class MuxPlayerInternal {
...
+    if (!video) {
+      console.error("The video element is missing!");
+      return;
+    }
+
+    await Promise.all([
+      // If there is no video src on construction don't be waiting on loadstart.
+      // It might look like we need to check video.networkState here also but
+      // the loadstart event is always fired a tick after the video src is set.
+      video.src && promisify(video.addEventListener, video)("loadstart"),
```
cjpillsbury, line 233: nit (non-blocking): We could simplify `promisify` by removing the concept of a `ctx` and simply binding here, e.g.:
```js
video.src && promisify(video.addEventListener.bind(video))("loadstart"),
```
Defer to you on preference.

cjpillsbury, line 233: blocking: What happens here if we set e.g. `playbackId` after initial load/instantiation of the `mux-player`? Will this still work and work as expected?

cjpillsbury (changes requested):
A couple of callouts. There are also a few other "supported"/"available" checks that should probably be here, but those will hopefully be solved in Media Chrome soon and I don't believe should be a blocker for another alpha release, as they only really impact minor cosmetic details for the use cases.

packages/mux-player/src/index.ts
luwes: I think so because there won't be a `src` and the condition will return `false`, 
Promise.all would resolve the `false` immediately.

I'll double check tough to be sure

cjpillsbury: ah do we want playerready to fire immediately when there's no src? I'd assume no.

cjpillsbury:
> Maybe one addition could be in case hls.js is used to listen for the manifest loaded event and fire `ready` after that? But then ideally that would live in playback core if we want to have mux-player not interface directly with the `muxVideo.hls` api.

I agree with all of this.

packages/mux-player/src/index.ts
luwes: yes, I think we should see this more as an API ready event, ready to take commands from the outside.

dylanjha:
Are there some decisions around the details that need to be made for this event?

PR #88, opened 2022-02-18
Fix dialog flashing glitch

luwes:
Fixes an issue where the custom element `<mxp-dialog>` was flashing on player load while it should be hidden.

### DIFF
--- a/packages/mux-player/src/styles.css
+++ b/packages/mux-player/src/styles.css
@@ -5,6 +5,11 @@
   width: 100%;
 }
 
+/* Hide custom elements that are not defined yet */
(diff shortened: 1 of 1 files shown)


packages/mux-player/src/styles.css
```diff
@@ -5,6 +5,11 @@
   width: 100%;
 }
 
+/* Hide custom elements that are not defined yet */
+:not(:defined) {
```
cjpillsbury, line 9: TIL `:defined`! 😎

cjpillsbury (approved):

PR #89, opened 2022-02-18
Inverse controls background transition duration

luwes:
Noticed this was animating inversed compared to the controls. Should come up fast .25s, and fade out slow 1s.


gkatsev (approved):
oops, good catch, thanks!

cjpillsbury (approved):
Good catch.

PR #90, opened 2022-02-18
Add extra small player view w/ only play button

luwes:
Adds an extra small view with just a play button icon.

SCR-20220217-s4a

### DIFF
--- a/packages/mux-player/src/styles.css
+++ b/packages/mux-player/src/styles.css
@@ -25,6 +25,12 @@ media-control-bar [role="switch"] {
   height: 44px;
 }
 
+.size-extra-small media-control-bar [role="button"],
+.size-extra-small media-control-bar [role="switch"] {
+  height: auto;
+  padding: 4.4% 3.2%;
+}
+
 .mxp-spacer {
   flex-grow: 1;
   height: 100%;
@@ -56,7 +62,7 @@ media-controller:is([media-paused], :not([user-inactive]))::part(vertical-layer)
   --media-control-background: transparent;
   --media-control-hover-background: transparent;
   padding: 0;
-  width: min(9%, 90px);
+  width: max(27px, min(9%, 90px));
 }
 
 .mxp-center-controls media-seek-backward-button,
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -226,22 +226,7 @@ export const VodChromeLarge = (props: MuxTemplateProps) => html`
 `;
 
 // prettier-ignore
-export const LiveChromeExtraSmall = (props: MuxTemplateProps) => html`
-  <media-control-bar slot="top-chrome">
-    ${props.hasCaptions && MediaCaptionsButton(props)}
-    ${Spacer()}
-    ${props.supportsAirPlay && MediaAirplayButton()}
-    ${MediaPipButton()}
-  </media-control-bar>
-  <div slot="centered-chrome" class="mxp-center-controls">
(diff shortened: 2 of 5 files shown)


cjpillsbury (commented):
A little worried about loosing mute for this view. Gonna hold off on approval until we can chat with the team. Otherwise, 😎

gkatsev:
Hiding the captions button is also a bit of an issue.
What about putting the mute button top left corner and captions button top right corner?

luwes:
SCR-20220218-nhx
SCR-20220218-ni5
SCR-20220218-nih

packages/mux-player/src/template.ts
```diff
@@ -195,6 +225,24 @@ export const VodChromeLarge = (props: MuxTemplateProps) => html`
   </media-control-bar>
 `;
 
+// prettier-ignore
+export const LiveChromeExtraSmall = (props: MuxTemplateProps) => html`
```
cjpillsbury, line 229: nit (blocking): If these are identical (for now), let's share the template fn as:
`export const LiveChromeExtraSmall = VodChromeExtraSmall;`

packages/mux-player/src/template.ts
```diff
@@ -152,6 +164,24 @@ const MediaFullscreenButton = () => html`
   ${icons.FullscreenExit()}
 </media-fullscreen-button>`;
 
+// prettier-ignore
+export const VodChromeExtraSmall = (props: MuxTemplateProps) => html`
```
cjpillsbury, line 168: non-blocking: PIP/Fullscreen buttons weren't discussed, but I'm cool with adding them and iterating as this hasn't gone through design and is easy enough to update.

cjpillsbury (changes requested):
One small ask, otherwise lgtm.

packages/mux-player/src/utils.ts
```diff
@@ -3,7 +3,7 @@ export function stylePropsToString(props: any) {
   Object.entries(props).forEach(([key, value]) => {
     style += `${kebabCase(key)}: ${value}; `;
   });
-  return style.trim();
+  return style ? style.trim() : undefined;
```
cjpillsbury, line 6: non-blocking: any reason this is yielding `undefined`?

packages/mux-player/src/styles.css
```diff
@@ -56,7 +62,7 @@ media-controller:is([media-paused], :not([user-inactive]))::part(vertical-layer)
   --media-control-background: transparent;
   --media-control-hover-background: transparent;
   padding: 0;
-  width: min(9%, 90px);
+  width: max(27px, min(9%, 90px));
```
cjpillsbury, line 65: non-blocking: This feels brittle. Will this break if the breakpoint changes?

packages/mux-player/src/styles.css
```diff
@@ -25,6 +25,12 @@ media-control-bar [role="switch"] {
   height: 44px;
 }
 
+.size-extra-small media-control-bar [role="button"],
```
cjpillsbury, line 28: interesting solution to "categorize" buttons. Might break down over time, but so would any `:is()` list or similar. 👌

packages/mux-player/src/template.ts
luwes: yes seemed useful if people want to see the tiny video in a larger view

cjpillsbury (approved):

dylanjha:
very nice 😎

PR #93, opened 2022-02-18
fix/aria-hide-inactive-error-dialog: set display to none for dialog w…

cjpillsbury:
…hen closed, in part to keep it from showing up in the AOM.


gkatsev (approved):
Verified locally as well.

luwes:
just putting here for reference that it broke the dialog CSS transition. can re-visit later, might need transitionend events.

PR #94, opened 2022-02-18
fix/error-dialog-autofocus: set focus to showing error dialog iff anc…

cjpillsbury:
…estor mux player has focus.

### DIFF
--- a/packages/mux-player/src/media-chrome/dialog.ts
+++ b/packages/mux-player/src/media-chrome/dialog.ts
@@ -90,7 +90,22 @@ class MediaDialog extends HTMLElement {
 
   show() {
     this.setAttribute("open", "");
-    focus(this);
+    if (document.activeElement?.nodeName.toLowerCase() === "mux-player") {
+      let parentEl: ParentNode | null | undefined = this.parentNode;
+      while (parentEl?.nodeName.toLowerCase() !== "mux-player") {
+        parentEl =
+          parentEl instanceof ShadowRoot ? parentEl.host : parentEl?.parentNode;
+        if (!parentEl || parentEl === document) {
+          console.error(
+            "Not in a mux-player or something has gone terrible wrong!"
+          );
+          break;
+        }
+      }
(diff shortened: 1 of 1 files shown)


packages/mux-player/src/media-chrome/dialog.ts
```diff
@@ -90,7 +90,22 @@ class MediaDialog extends HTMLElement {
...
+        parentEl =
+          parentEl instanceof ShadowRoot ? parentEl.host : parentEl?.parentNode;
+        if (!parentEl || parentEl === document) {
+          console.error(
+            "Not in a mux-player or something has gone terrible wrong!"
+          );
+          break;
+        }
+      }
+      if (parentEl === document.activeElement) {
```
luwes, line 105: what if the focus is on the play button? does this condition include that?

packages/mux-player/src/media-chrome/dialog.ts
```diff
@@ -90,7 +90,22 @@ class MediaDialog extends HTMLElement {
 
   show() {
     this.setAttribute("open", "");
-    focus(this);
+    if (document.activeElement?.nodeName.toLowerCase() === "mux-player") {
```
luwes, line 93: the dialog shouldn't have to know about `mux-player` I think, thinking of when this would be included in media-chrome for example.

this code is based on the native dialog element and that assumes that the parent is the whole `document` and it's okay to steal focus because the dialog would go over the whole page. 

maybe it makes more sense we provide a "parent" for the dialog somehow

cjpillsbury: cool yeah let's put this "on hold" until Tuesday to not rush things. I think a "fine for now" and "fast follow" should/would be fine to get alpha buttoned up, even if we don't think the implementation is what we'd want long term. That said, there's more work to be done here, regardless.

gkatsev:
With #95, this works as expected!

dylanjha:
What's the status on this one?

PR #96, opened 2022-02-22
Fix poster loading multiple times on re-render

luwes:
This change prevents setting the poster attribute if it didn't change. Re-setting the poster attribute on a video elements makes it re-request the poster url and shows a brief flash. This would also happen going from one UI view to another, small to large for example.

### DIFF
--- a/packages/mux-player/src/html.ts
+++ b/packages/mux-player/src/html.ts
@@ -27,6 +27,14 @@ export function processPropertyIdentity(
   part: TemplatePart,
   value: unknown
 ): boolean {
+  if (part instanceof AttributeTemplatePart) {
+    const ns = part.attributeNamespace;
+    const oldValue = part.element.getAttributeNS(ns, part.attributeName);
+    if (String(value) !== oldValue) {
+      part.value = String(value);
+    }
+    return true;
+  }
   part.value = String(value);
   return true;
 }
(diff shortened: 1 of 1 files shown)


packages/mux-player/src/html.ts
```diff
@@ -59,10 +59,11 @@ export function processBooleanNode(
 }
 
 export function processPart(part: TemplatePart, value: unknown): void {
-  processBooleanAttribute(part, value) ||
+  processDirective(part, value) ||
     processBooleanNode(part, value) ||
     processSubTemplate(part, value) ||
     processDocumentFragment(part, value) ||
+    processBooleanAttribute(part, value) ||
```
cjpillsbury, line 66: any reason this order changed?

packages/mux-player/src/template.ts
```diff
@@ -40,11 +40,13 @@ export const template = (props: MuxTemplateProps) => html`
       debug="${props.debug}"
       prefer-mse="${props.preferMse}"
       start-time="${props.startTime != null ? props.startTime : false}"
-      poster="${!!props.poster
-        ? props.poster
-        : props.playbackId
-        ? getPosterURLFromPlaybackId(props.playbackId)
-        : false}"
+      poster="${cache(
```
cjpillsbury, line 43: could this instead be handled by making mux-video smarter about updating its internal state only if the value changes (likely for many attributes)?

packages/mux-player/src/html.ts
luwes: I wanted processing the attributes together so it was easier to see because they're used now below too
https://github.com/muxinc/elements/pull/96/files#diff-41da7eb53a2250a1a0e038f4415e05489b9435fb1eba29153d6d2732d8973a52R181-R182

packages/mux-player/src/template.ts
luwes: I'll try it out

cjpillsbury (approved):
ah ok I see you just changed the architecture to always only set if changed. Seems reasonable to me. 👍

PR #98, opened 2022-02-22
feat: default-hidden-captions to turn off showing captions by default

gkatsev:
Setting `default-hidden-captions` on the WC element or `defaultHiddenCaptions` on the React element will have the captions not turn on by default.


cjpillsbury (changes requested):
nit (blocking): Should add as props in mux-player-react for appropriate typescript typings

cjpillsbury (approved):

PR #99, opened 2022-02-22
feat: only autoplay with the autoplay attribute

gkatsev:
Also, seek to liveSyncPosition on firstplay for live videos

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -143,7 +143,6 @@ class MuxPlayerElement extends VideoApiElement {
     // }
 
     this.#setUpErrors();
-    this.#setUpLiveFirstPlay();
     this.#setUpMutedAutoplay();
     this.#setUpCaptionsButton();
     this.#setUpAirplayButton();
@@ -255,29 +254,6 @@ class MuxPlayerElement extends VideoApiElement {
         );
       };
       this.video.hls.on(Hls.Events.ERROR, onHlsError);
-    }
-  }
-
-  #setUpLiveFirstPlay() {
-    if (this.video?.hls) {
-      const Hls: any = this.video.hls.constructor;
-      if (!this.autoplay) {
-        this.video.hls.on(Hls.Events.LEVEL_LOADED, (e: any, data: any) => {
-          const isLive: boolean = data.details.live;
-
-          if (isLive) {
-            this.video?.addEventListener(
-              "play",
-              () => {
-                if (this.video?.hls?.liveSyncPosition) {
-                  this.video.currentTime = this.video.hls.liveSyncPosition;
-                }
-              },
-              { once: true }
-            );
-          }
-        });
-      }
     }
   }
 
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -262,14 +262,16 @@ class MuxPlayerElement extends VideoApiElement {
     if (this.video?.hls) {
       const Hls: any = this.video.hls.constructor;
       if (!this.autoplay) {
-        this.video.hls.on(Hls.Events.LEVEL_LOADED, (e, data) => {
-          const isLive = data.details.live;
+        this.video.hls.on(Hls.Events.LEVEL_LOADED, (e: any, data: any) => {
+          const isLive: boolean = data.details.live;
 
           if (isLive) {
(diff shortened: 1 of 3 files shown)


packages/mux-player/src/index.ts
```diff
@@ -172,6 +173,27 @@ class MuxPlayerInternal {
...
+              () => {
+                el.video.currentTime = el.video.hls.liveSyncPosition;
+              },
+              { once: true }
+            );
+          }
+        });
+      }
+    }
+  }
```
luwes, line 195: could this already be moved to mux-video? 

this is needed because hls.js starts loading, filling the buffer right away on player load, I wonder if that should happen with live streams if autoplay is not enabled?

gkatsev: Yeah, definitely could be moved to `mux-video`. This is a draft PR, plus, if we wanted to augment the `autoplay` attribute, we'll want to figure out whether we want to do it in mux-player or mux-video, or both, since mux-video currently only accepts a boolean autoplay attribute.

packages/mux-player/src/index.ts
```diff
@@ -55,6 +55,7 @@ class MuxPlayerInternal {
     // }
 
     this._setUpErrors(el);
+    this._setUpLiveFirstPlay(el);
```
cjpillsbury, line 58: Unless I'm missing something, this assumes playback-id has been set on initialization.

packages/mux-player/src/index.ts
```diff
@@ -172,6 +173,29 @@ class MuxPlayerInternal {
     }
   }
 
+  _setUpLiveFirstPlay(el: MuxPlayerElement) {
+    if (el.video?.hls) {
```
cjpillsbury, line 177: `hls` won't exist if there is not a `playback-id`/`src` set, which prompts an `initialize()` in playback-core.

packages/mux-player/src/index.ts
gkatsev: Probably yes. As @luwes mentioned, this might be better served being moved into mux-video as well.
Though, there's also the question if we wanted to support the extended autoplay values.

packages/mux-player/src/index.ts
gkatsev: yeah, this should probably move into playback-core

gkatsev: event types are those that are live but playback should start from the beginning or from EXT-X-START? In which case, we wouldn't want to do this seek. I'll have to see if and how hls.js exposes this info.

For reference, VHS doesn't currently support EXT-X-START or starting at the beginning of a live playlist.

cjpillsbury:
@gkatsev Is this still a draft? I haven't smoke tested, but I believe it's complete for the baseline refactor of autoplay behavior..

gkatsev:
@cjpillsbury at first I was thinking of getting the extended autoplay stuff ready, but I think that should come in a separate PR. Making this available for review.

packages/playback-core/src/index.ts
```diff
@@ -251,18 +253,20 @@ export const loadMedia = (
   props: Partial<
     Pick<
       MuxMediaProps,
-      "preferMse" | "src" | "type" | "startTime" | "streamType"
+      "preferMse" | "src" | "type" | "startTime" | "streamType" | "autoplay"
```
cjpillsbury, line 256: blocking: We need to account for and validate against React for the camelcase of `autoPlay`. Should be able to test against `<MuxVideo/>` & `<MuxAudio/>` (React). As a quick fix we can just allow both here.

cjpillsbury (changes requested):
One callout. Otherwise lgtm.

luwes:
LGTM w/ autoPlay fix

packages/playback-core/src/index.ts
gkatsev: Seems to work without any type errors. Probably by the time it gets here from react, it gets renamed to `autoplay`. Also, locally, both `autoplay` and `autoPlay` properties work in React.
https://elements-demo-nextjs-git-fork-gkatsev-autoplay-shenanigans-mux.vercel.app/MuxVideo

cjpillsbury (approved):
🚢 🇮🇹

PR #104, opened 2022-02-23
chore: fixup minify build with lerna.

gkatsev:
lerna sets the npm_exec_path to pretend it's npm for some scripts. This
causes, `npm-run-all` to use lerna rather than `npm` or yarn and the
`--minify` doesn't get passed automatically.

Instead, by using `--` we can force the `--minify` flag to get passed to
the underlying command.

### DIFF
--- a/packages/mux-player-react/package.json
+++ b/packages/mux-player-react/package.json
@@ -16,9 +16,9 @@
     "dev:cjs": "open-process | yarn build:cjs --watch",
     "dev:types": "yarn build:types -w",
     "dev": "npm-run-all --parallel dev:types dev:cjs",
-    "build:cjs": "esbuild src/index.tsx --target=es2019 --minify --bundle --sourcemap --format=cjs --loader:.css=text --outdir=dist --external:react --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
+    "build:cjs": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --format=cjs --loader:.css=text --outdir=dist --external:react --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:types": "tsc --declaration --emitDeclarationOnly --outDir './dist/types'",
-    "build": "npm-run-all --parallel build:types build:cjs"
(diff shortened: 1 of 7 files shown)


packages/mux-player-react/package.json
```diff
@@ -16,9 +16,9 @@
     "dev:cjs": "open-process | yarn build:cjs --watch",
     "dev:types": "yarn build:types -w",
     "dev": "npm-run-all --parallel dev:types dev:cjs",
-    "build:cjs": "esbuild src/index.tsx --target=es2019 --minify --bundle --sourcemap --format=cjs --loader:.css=text --outdir=dist --external:react --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
```
gkatsev, line 19: this didn't get migrated previously.

luwes:
this looks good, one addition is to do this also for the `prepare` script in the root package.json
`--scope @mux-elements/*` was ignored I believe. at least the demo examples were built while they shouldn't

gkatsev:
@luwes seems to work for me?
```sh
$ yarn run prepare
yarn run v1.22.17
$ husky install && yarn build --scope @mux-elements/*
husky - Git hooks installed
$ lerna run build --scope '@mux-elements/*'
lerna notice cli v4.0.0
lerna info versioning independent
lerna notice filter including "@mux-elements/*"
lerna info filter [ '@mux-elements/*' ]
.
.
.
lerna success run Ran npm script 'build' in 7 packages in 10.1s:
lerna success - @mux-elements/mux-audio-react
lerna success - @mux-elements/mux-audio
lerna success - @mux-elements/mux-player-react
lerna success - @mux-elements/mux-player
lerna success - @mux-elements/mux-video-react
lerna success - @mux-elements/mux-video
lerna success - @mux-elements/playback-core
✨  Done in 11.54s.
```

`yarn build` builds the examples too, though. But I think that's expected.

luwes:
yes but when prepare is called indirectly from the lerna publish command then the scope was not used. 
I saw it this morning when @dylanjha tried releasing.

luwes (approved):
👍

PR #107, opened 2022-02-26
Add signing tokens

luwes:
I didn't add the setters/getters yet for the different tokens with update logic so it passes it down to mux-video.

7d valid tokens:

```
playback-id="bzeU5fRA47S01KDzrObYiiZvzj00j5E00dCVbt3oRzmfF00"
playback-token="eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6Ik96VU90ek1nUWhPbkk2MDJ6SlFQbU52THR4MDBnSjJqTlBxN0tTTzAxQlozelEifQ.eyJleHAiOjE2NDY0Mzg5NjEsImF1ZCI6InYiLCJzdWIiOiJiemVVNWZSQTQ3UzAxS0R6ck9iWWlpWnZ6ajAwajVFMDBkQ1ZidDNvUnptZkYwMCJ9.hWrdcJDa8FJCfVFP19oJ-9FSEVk9eB6DTOCRrucnzsrtUoZbb1OFe7swpQ38Fp3hZNNIt7-LWjdOl90TF4ucu7mhu42qyk3_i054RtmEZyQaj5Qjm3_H4sa2jLO-0QNSnOfp1A9x-fI8M_giGLg-byJPuu_eUqu1MW9bILLly_9gq8m0cNKghUa9xTMJgFmaya4XYudy5Mt2Fu72MiS3csUP3xhKlONVnGHlMRqB-dBVOgAJrayeUquAhaNY346oFBUWVM-EcAZ9G2ARtPakfy4Wpv5BsRKEGtR81P-k7EW8g27U0FKLlrvLkUz3Z-JYu53CRcJUvjkC9sDMrZLcTA"
thumbnail-token="eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6Ik96VU90ek1nUWhPbkk2MDJ6SlFQbU52THR4MDBnSjJqTlBxN0tTTzAxQlozelEifQ.eyJleHAiOjE2NDY0OTMzMTksImF1ZCI6InQiLCJzdWIiOiJiemVVNWZSQTQ3UzAxS0R6ck9iWWlpWnZ6ajAwajVFMDBkQ1ZidDNvUnptZkYwMCJ9.hNBRo1-XDTT1CJMOxf90-8JPJzAygwm-3pVNBj31I7DEukSVRKVgUuhquEJbYXx1xg27xRMu8OVQxVob6jWHdjSwTyygAY040bqdyDxLsRtkDcVxwZ78iiZwtA1eTkxxY-410Ma3HbhNsG0Qjo5AWX46IhD9ARKHL-MPGaKda7FSx8J8jxa3hQ8_M1AKMsx7PrgJYOtW6n0mvkupEAFYRJlqIbkERSBeWChdrjCLYAcXRar5nfdNWlWST2pfllqz8pfJSTWjQRumTonC5BGB89jZUimHnuzkRXm_LeGyXbfZmBKb4d0j9YyGVnTPePqyVPsAQ-bzcfFDU0L67GDgyw"
storyboard-token="eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6Ik96VU90ek1nUWhPbkk2MDJ6SlFQbU52THR4MDBnSjJqTlBxN0tTTzAxQlozelEifQ.eyJleHAiOjE2NDY0OTM0OTgsImF1ZCI6InMiLCJzdWIiOiJiemVVNWZSQTQ3UzAxS0R6ck9iWWlpWnZ6ajAwajVFMDBkQ1ZidDNvUnptZkYwMCJ9.PKEybohVK0JyJGX_3iubRnHZx1ve5OmPmyfZdaKb17N2wVMQCYNltTc-gCUTU7EIKGeTtVOIITCSsIeTgXcI667B2GWJ5juDIErz1h-NQsPIfB-FsUeuWx2rYOap4G3FdwEIjaGc29HPncw-mG0JLcqkMB7jtDxjBY_-YlpjFYJF_z7r-1yIJM7mF3rl8YqeWstojC8oh2Iv2VRkuTyPE31QVI6fQcet5PIRWHudUIGWcNiWM56vwZskJ6qod8UvYpha7K5rhshh0Xdhnvq3Y9b6PXl3fy6VKCZIyszlPVje0IR2bR9iHDXnGbawivUsI65IDm-ZEoJrOzmZctMWAQ"
```

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -259,14 +259,19 @@ const PlayerAttributes = {
   SECONDARY_COLOR: "secondary-color",
   FORWARD_SEEK_OFFSET: "forward-seek-offset",
   BACKWARD_SEEK_OFFSET: "backward-seek-offset",
+  PLAYBACK_TOKEN: "playback-token",
+  THUMBNAIL_TOKEN: "thumbnail-token",
+  STORYBOARD_TOKEN: "storyboard-token",
 };
 
 function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
   return {
+    src: el.src,
     debug: el.debug,
     muted: el.muted,
     envKey: el.envKey,
     playbackId: el.playbackId,
+    tokens: el.tokens,
     poster: el.poster,
     metadata: el.metadata,
     playerSoftwareName: el.playerSoftwareName,
@@ -294,6 +299,8 @@ const playerSoftwareName = "mux-player";
 const internals = new WeakMap();
 
 class MuxPlayerElement extends VideoApiElement {
+  #tokens = {};
+
   static get observedAttributes() {
     return [
       ...(VideoApiElement.observedAttributes ?? []),
@@ -322,7 +329,11 @@ class MuxPlayerElement extends VideoApiElement {
   ) {
     super.attributeChangedCallback(attrName, oldValue, newValue);
 
-    if (MuxVideoAttributeNames.includes(attrName)) {
+    // Prevent forwarding the playback-id, the player sets the src attribute.
+    if (
+      attrName != MuxVideoAttributes.PLAYBACK_ID &&
+      MuxVideoAttributeNames.includes(attrName)
+    ) {
       if (newValue === null) {
         this.video?.removeAttribute(attrName);
       } else {
@@ -384,7 +395,9 @@ class MuxPlayerElement extends VideoApiElement {
    * @return {string?}
    */
   get playbackId() {
-    return getVideoAttribute(this, MuxVideoAttributes.PLAYBACK_ID);
+    // Don't get the mux-video attribute here because it could have the
+    // playback token appended to it.
+    return this.getAttribute(MuxVideoAttributes.PLAYBACK_ID);
   }
 
   /**
@@ -483,6 +496,22 @@ class MuxPlayerElement extends VideoApiElement {
    */
   set metadata(val) {
     if (this.video) this.video.metadata = val;
+  }
+
+  get tokens() {
+    const playback = this.getAttribute(PlayerAttributes.PLAYBACK_TOKEN);
+    const thumbnail = this.getAttribute(PlayerAttributes.THUMBNAIL_TOKEN);
+    const storyboard = this.getAttribute(PlayerAttributes.STORYBOARD_TOKEN);
+    return {
+      ...this.#tokens,
+      ...(playback != null ? { playback } : {}),
+      ...(thumbnail != null ? { thumbnail } : {}),
+      ...(storyboard != null ? { storyboard } : {}),
+    };
+  }
+
+  set tokens(val) {
+    this.#tokens = val ?? {};
   }
 }
 
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -1,6 +1,7 @@
 import "./dialog";
 import {
   getChromeStylesFromProps,
+  getSrcFromPlaybackId,
   getPosterURLFromPlaybackId,
   getStoryboardURLFromPlaybackId,
 } from "./helpers";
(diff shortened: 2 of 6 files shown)


packages/mux-player/src/utils.ts
```diff
@@ -21,3 +21,15 @@ export function toNumberOrUndefined(val: any) {
...
 }
+
+export function toQuery(obj: Record<string, any>) {
+  const params = toParams(obj).toString();
+  return params ? "?" + params : "";
+}
+
+export function toParams(obj: Record<string, any>) {
+  for (let key in obj) {
+    if (obj[key] == null) delete obj[key];
```
cjpillsbury, line 32: blocker: this will mutate the object passed in. If we do want to handle automatically pruning null/undefined (not sure we do, but can see an argument for it), we should probably do it in a non-mutative way, e.g.:
```js
export function toParams(obj: Record<string, any>) {
  const objSansNils = Object.entries(obj).reduce<Record<string, any>>(
    (objSansNils, [k, v]) => {
      if (v != null) {
        objSansNils[k] = v;
      }
      return objSansNils;
    },
    {}
  );
  return new URLSearchParams(objSansNils);
}
```

packages/mux-player/src/index.ts
```diff
@@ -484,6 +497,22 @@ class MuxPlayerElement extends VideoApiElement {
...
     if (this.video) this.video.metadata = val;
   }
+
+  get tokens() {
+    const playback = this.getAttribute(PlayerAttributes.PLAYBACK_TOKEN);
+    const thumbnail = this.getAttribute(PlayerAttributes.THUMBNAIL_TOKEN);
+    const storyboard = this.getAttribute(PlayerAttributes.STORYBOARD_TOKEN);
+    return {
+      ...this.#tokens,
+      ...(playback != null ? { playback } : {}),
```
cjpillsbury, line 507: seems reasonable to me. We could optimize this for the null cases, but I don't think that's a huge priority.

packages/mux-player/src/template.ts
```diff
@@ -44,14 +45,18 @@ export const template = (props: MuxTemplateProps) => html`
       debug="${props.debug}"
       prefer-mse="${props.preferMse}"
       start-time="${props.startTime != null ? props.startTime : false}"
+      src="${!!props.src
```
cjpillsbury, line 48: non-blocking: Any reason we aren't using the supported `playback-id` with token that supported in `mux-video`?

cjpillsbury (changes requested):
A couple of minor asks, otherwise 👌

packages/mux-player/src/index.ts
```diff
@@ -294,6 +299,8 @@ const playerSoftwareName = "mux-player";
 const internals = new WeakMap();
 
 class MuxPlayerElement extends VideoApiElement {
+  #tokens = {};
```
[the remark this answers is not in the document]
luwes: I'm not such a fan of TS private and protected modifiers, they don't actually make the props private I noticed the other day in the compiled JS. Would love to keep the mux-player element clean without `.__privateProp` props.

packages/mux-player/src/template.ts
luwes: yes, it's easier to write with also supporting setting the `src` attribute.
it was also a small issue that mux-video supports adding the token to the playback-id and mux-player does not.

packages/mux-player/test/player.test.js
```diff
@@ -2,17 +2,6 @@ import { fixture, assert, aTimeout } from "@open-wc/testing";
 import "../src/index.ts";
 
 describe("<mux-player>", () => {
-  it("forwards attrs to mux-video", async function () {
```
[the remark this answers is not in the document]
luwes: oh actually the `src` check is done in the tests below.

packages/mux-player/src/utils.ts
luwes: fixed in [`1c0cc70` (#107)](https://github.com/muxinc/elements/pull/107/commits/1c0cc703ca9756f8257c13386ab61c3846d1189c)

packages/mux-player/src/index.ts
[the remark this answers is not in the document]
cjpillsbury: yeah, I hear that. It's just sad that JS doesn't have a concept of "protected" 🙁. We can do an "ask the audience" on prefs here for the team.

packages/mux-player/src/utils.ts
```diff
@@ -21,3 +21,16 @@ export function toNumberOrUndefined(val: any) {
...
+
+export function toQuery(obj: Record<string, any>) {
+  const params = toParams(obj).toString();
+  return params ? "?" + params : "";
+}
+
+export function toParams(obj: Record<string, any>) {
+  let params: Record<string, any> = {};
+  for (let key in obj) {
+    if (obj[key] != null) params[key] = obj[key];
```
cjpillsbury, line 33: works for me!

cjpillsbury (approved):

gkatsev (approved):

packages/mux-player/src/index.ts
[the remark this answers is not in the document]
gkatsev: since we're exposing getters and setters for this value, I don't see a problem with making it private.

PR #108, opened 2022-02-26
Add signing token getters / setters and use private class features

luwes:
This builds on top of the signing-tokens PR #107 

- Adds signing token getter / setters
- Adds colors, seek offset, playbackId, envKey setters
- A small refactor to use private class features instead of the weakmap / extra internal class logic.


packages/mux-player/src/index.ts
```diff
@@ -328,28 +309,48 @@ class MuxPlayerElement extends VideoApiElement {
...
-    if (
-      attrName != MuxVideoAttributes.PLAYBACK_ID &&
-      MuxVideoAttributeNames.includes(attrName)
-    ) {
-      if (newValue === null) {
-        this.video?.removeAttribute(attrName);
-      } else {
-        this.video?.setAttribute(attrName, newValue);
-      }
-    }
```
luwes, line 342: ^^ this is pretty cool, `this.#render();` will now take care of setting and removing the attributes instead of the API calls above here.

packages/mux-player/src/index.ts
```diff
@@ -510,9 +537,54 @@ class MuxPlayerElement extends VideoApiElement {
...
+   * Set the signing tokens for the Mux asset URL's.
+   */
+  set tokens(val: Tokens | undefined) {
     this.#tokens = val ?? {};
   }
+
+  /**
+   * Get the playback token for signing the src URL.
+   */
+  get playbackToken() {
```
gkatsev, line 550: thought: what about interfacing with these tokens via `tokens()`? So, rather than
```js
player.playbackToken = '123'
```
you would
```js
player.tokens.playback = '123'
```

packages/mux-player/src/index.ts
```diff
@@ -456,23 +485,20 @@ class MuxPlayerElement extends VideoApiElement {
 
   /**
    * Set the start time.
-   * @param  {number} val
```
gkatsev, line 459: these aren't necessary because we have the types file?

packages/mux-player/src/index.ts
gkatsev: this is because of template-parts and html.ts?

packages/mux-player/src/index.ts
```diff
@@ -15,34 +15,122 @@ import { toNumberOrUndefined } from "./utils";
...
+function getPlayerSize(el: Element) {
+  const muxPlayerRect = el.getBoundingClientRect();
+  return muxPlayerRect.width < XSMALL_BREAKPOINT
+    ? MediaChromeSizes.XS
+    : muxPlayerRect.width < SMALL_BREAKPOINT
+    ? MediaChromeSizes.SM
+    : MediaChromeSizes.LG;
+}
+
+const MuxVideoAttributes = {
```
gkatsev, line 41: Should these (and `PlayerAttributes`, `getProps` and the Names objects) move into some kind of `props.ts` file?

gkatsev (approved):
Had a couple of non-blocking thoughts or questions. Overall, seems like a good change!

cjpillsbury (approved):
👍

packages/mux-player/src/index.ts
luwes: yup :)

packages/mux-player/src/index.ts
luwes: yes, most of the types are also inferred

packages/mux-player/src/index.ts
luwes: good call, the setting like that does work at the moment. only downside is that it would not trigger the attributeChangedCallback. thinking of this, even setting the full tokens property would not re-render. maybe it should

gkatsev: internally, we could make that when we set `tokens.playback = '123'` we call `setAttribute`. Not sure if that's worth doing over having `player.playbackToken = '123'`

luwes: I will get kinda messy with setters in the plain object etc or we would need to create a new class for Tokens.
maybe lets iterate on it and see if users prefer to use it like that.

packages/mux-player/src/index.ts
luwes: could be nice. we can revisit this.

PR #111, opened 2022-02-28
Improve error handling

luwes:
Just wanted to open up a draft PR to start a discussion because I think we should initially remove the playback-core error handling for some reasons like:

- Currently it's possible to get in an endless loop
- It looks like this sample code shouldn't be taken literally, it requires more work on our part to handle it and check for specific conditions before to actually try and recover. The hls.js demo shows more of this, see here: https://github.com/video-dev/hls.js/blob/2635570ca3ffc6ec2c0f8ef4be84b7da9bf39443/demo/main.js#L1041-L1084
- hls.js does already retry many areas if they fail with an exponential backoff for network errors etc. and retries for appending segments see https://github.com/video-dev/hls.js/blob/master/docs/design.md#error-detection-and-handling
- A fatal network error can be on the master manifest or the playlist which need different recovery; `hls.startLoad();` and `hls.loadSource(manifestURL)` but again this is already retried in hls.js itself.
- The sample code at https://github.com/video-dev/hls.js/commit/c4f6564e7dcf50da295e938a166fda48d02d3fa9#diff-b161966f7254af9c9c5382e859495c69622843e4602a6bb4208e08134e660121R139-R158 was written in 2015.

Some resources for offline / online handling:
- https://github.com/video-dev/hls.js/issues/1878#issuecomment-416752563
- https://github.hubspot.com/offline/docs/welcome/

### DIFF
--- a/scripts/i18n-utils/i18n-utils.js
+++ b/scripts/i18n-utils/i18n-utils.js
@@ -1,0 +1,54 @@
+#!/usr/bin/env node
+import fs from "fs";
+import path from "path";
+
+const srcFile = process.argv[2];
+const langFolder = process.argv[3];
+const baseLangFile = `${langFolder}/en.json`;
+const srcContents = fs.readFileSync(srcFile).toString();
+const baseLangContents = {};
+
+const strings = [];
+const regex = /\b(?:i18n`(.*?)`)/g;
+let result;
+
+while ((result = regex.exec(srcContents)) !== null) {
+  let i = 0;
+  strings.push(
+    result[1].replace(/\$\{(.*?)\}/g, (match, key) => {
+      return "${" + i++ + "}";
+    })
+  );
+}
+
+for (let file of walkSync(langFolder)) {
+  const oldDict = JSON.parse(fs.readFileSync(file).toString());
+  let dict = {};
+  if (file.endsWith("en.json")) {
+    for (let str of strings) {
+      dict[str] = str;
+    }
+  } else {
+    for (let str of strings) {
+      dict[str] = oldDict[str] ?? str;
+    }
+  }
+
+  fs.writeFileSync(file, JSON.stringify(dict, null, "  "));
+  console.log(
+    `${strings.length} strings (${
+      Object.keys(dict).length
+    } unique) written to ${file}`
+  );
+}
+
+function* walkSync(dir) {
+  const files = fs.readdirSync(dir, { withFileTypes: true });
+  for (const file of files) {
+    if (file.isDirectory()) {
+      yield* walkSync(path.join(dir, file.name));
+    } else {
+      yield path.join(dir, file.name);
+    }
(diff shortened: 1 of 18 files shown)


packages/mux-player/src/icons/volume-low.svg
```diff
@@ -1,4 +1,4 @@
-<svg aria-hidden="true" aria-hidden="true" viewBox="0 0 18 18" slot="low">
```
gkatsev, line 1: `aria-hidden="true" aria-hidden="true"`
when you *really* want to make sure that it's hidden from screen readers.

cjpillsbury:
**General Feedback, re: i18n/l10n solution**
I may be ok with this as an interim solution, but it is building out an infrastructure that moves us away from where the standards/specifications are and where they are very likely going. While they do not (yet, more below) have native browser support, Internationalized MessageFormat strings have been around for a long time and are based on ICU standards, and the Intl API is intentionally closely tied to things like the ICU CLDR (see: https://cldr.unicode.org/ & https://github.com/unicode-org/cldr-json).

Though it hasn't yet been updated on the official [tc39 ECMA 402 github stage tracker](https://github.com/tc39/proposals/tree/main/ecma402), it looks like Intl.MessageFormat has recently been accepted to stage 1 (see: https://docs.google.com/presentation/d/1oThTeL_n5-HAfmJTri-i8yU2YtHUvj9AakmWiyRGPlw/edit#slide=id.p & https://github.com/dminor/proposal-intl-messageformat). There is also a suite of polyfills and build tool packages for using MessageFormat as well, found [here](https://github.com/messageformat/messageformat).

While I am not recommending we go "whole hog" and use the polyfills for runtime support of `Intl.MessageFormat`, I do think we should be writing code/an infrastructure that "points in the direction" of assuming that will be where we are aiming. It also opens the door for other tooling, databases, 3rd party infrastructures that conform to these standards.

Again, since you've already provided a solution that will work, I could be convinced to keep it "as is" for now, but I wouldn't want us to continue to invest in more complex use cases if that's going to be moving our codebase farther away from where I think we will want to be.

scripts/i18n-utils/i18n-utils.js
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env node
+import fs from "fs";
+import path from "path";
+
+const srcFile = process.argv[2];
+const langFolder = process.argv[3];
+const baseLangFile = `${langFolder}/en.json`;
```
cjpillsbury, line 7: I think all of these hard-coded/narrow assumptions are fine for now and we can pick it back up in our i18n followup effort.

cjpillsbury (approved):
🚢 🇮🇹  can treat any additional larger discussions (in band on the PR feedback or out of band feature's we've been discussing) as 1+ followup effort(s)/PR(s)

PR #114, opened 2022-03-01
Chore/add test streams

cjpillsbury:


shared/assets/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
luwes, line 1: is there any other place this could live to not add more folders to the root dir?

gkatsev: could go as part of `scripts`?

cjpillsbury: we could add it to scripts, yeah. Would want to hold off on adding it to `examples` in case we want to use these in tests. Normally I think `scripts` or similar is shorthand for build tooling scripts and similar, though.

gkatsev: could rename `scripts` to something else too, like `local-packages` or something

cjpillsbury: cool that works for me.

luwes: yes, `local-packages` would be more confusing I think. it could contain the Mux elements as well

I like `shared` then better instead of scripts, or `./packages/shared/` maybe

PR #115, opened 2022-03-02
Fix new attrs not taking effect

luwes:
This change should make it possible to not use the setAttribute / removeAttribute methods for mux-video.

See [`0c58103` (#114)](https://github.com/muxinc/elements/pull/114/commits/0c58103b64d04a70e8791c6fb7597ed2d82d185a)

Allow for a more declarative view where things are not magically added / removed imperatively in other places in the code base.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -10,7 +10,7 @@ import {
 } from "./helpers";
 import { template } from "./template";
 import { render } from "./html";
-import { toNumberOrUndefined } from "./utils";
+import { toNumberOrUndefined, camelCase } from "./utils";
 
 import type { MuxTemplateProps } from "./types";
 import type { Metadata } from "@mux-elements/playback-core";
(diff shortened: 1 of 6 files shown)


packages/mux-player/src/index.ts
```diff
@@ -309,7 +312,7 @@ class MuxPlayerElement extends VideoApiElement {
     newValue: string
   ) {
     super.attributeChangedCallback(attrName, oldValue, newValue);
-    this.#render();
+    this.#render({ [camelCase(attrName)]: newValue });
```
cjpillsbury, line 315: blocking: let's break this out into a function for translating attribute names -> prop names

cjpillsbury (changes requested):
Just a couple of callouts, otherwise I'm down for this strategy.

cjpillsbury (approved):

gkatsev (approved):

PR #116, opened 2022-03-03
feat: Extended autoplay options

gkatsev:
This change adds a couple of string options for the `autoplay` attribute/`autoPlay` prop. The exiting boolean-attribute nature of `autoplay`/`autoPlay` lives on.
Specifically:
- `muted` - will mute the player and try autoplaying. If autoplaying fails, muted state is restored and player continues as a standard click-to-play player.
- `any` - will try playing with the current options, if it fails it'll behave like `muted` above.
  Happy to get alternatives to the name `any`.

In addition to the above, this switches from relying on native autoplay attribute behavior to calling `play` manually. In terms of behavior, it should be backwards compatible, but will allow us to better control this behavior going forward.

Example usage:
WC
```html
<!-- no autoplay -->
<mux-player ></mux-player>
<!-- autoplay muted, likely to work -->
<mux-player autoplay="muted"></mux-player>
<!-- autoplay with sound, likely to fail -->
<mux-player autoplay></mux-player>
```
React:
```jsx
// autoplay with sound, if possible, then muted if possible
<MuxPlayer autoPlay="any" />
```

### DIFF
--- a/packages/mux-audio-react/src/index.tsx
+++ b/packages/mux-audio-react/src/index.tsx
@@ -54,6 +54,7 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
 
     const mediaElRef = useCombinedRefs(innerMediaElRef, ref);
     const updateAutoplayRef = useRef<UpdateAutoplay | undefined>(undefined);
+    const [updateAutoplay, setUpdateAutoplay] = useState(() => (x: any) => {});
 
     useEffect(() => {
       const src = toMuxVideoURL(playbackId) ?? outerSrc;
@@ -75,16 +76,13 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
         playbackEngineRef.current
       );
       playbackEngineRef.current = nextPlaybackEngineRef;
-      const updateAutoplay = setupAutoplay(
-        mediaElRef.current,
-        autoPlay,
-        playbackEngineRef.current
+      setUpdateAutoplay(() =>
+        setupAutoplay(mediaElRef.current, autoPlay, playbackEngineRef.current)
       );
-      updateAutoplayRef.current = updateAutoplay;
     }, [src]);
 
     useEffect(() => {
-      updateAutoplayRef.current?.(autoPlay);
+      updateAutoplay(autoPlay);
     }, [autoPlay]);
 
     return (
--- a/packages/mux-audio-react/src/index.tsx
+++ b/packages/mux-audio-react/src/index.tsx
@@ -4,6 +4,7 @@ import PropTypes from "prop-types";
 import {
   allMediaTypes,
   initialize,
+  setupAutoplay,
   MuxMediaProps,
   StreamTypes,
   toMuxVideoURL,
@@ -50,6 +51,7 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
     const innerMediaElRef = useRef<HTMLAudioElement>(null);
 
     const mediaElRef = useCombinedRefs(innerMediaElRef, ref);
+    const updateAutoplayRef = useRef(null);
 
     useEffect(() => {
       const src = toMuxVideoURL(playbackId) ?? outerSrc;
@@ -71,7 +73,17 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
         playbackEngineRef.current
       );
       playbackEngineRef.current = nextPlaybackEngineRef;
+      const updateAutoplay = setupAutoplay(
+        mediaElRef.current,
+        autoPlay,
+        playbackEngineRef.current
+      );
+      updateAutoplayRef.current = updateAutoplay;
     }, [src]);
+
+    useEffect(() => {
+      updateAutoplayRef.current(autoPlay);
+    }, [autoPlay]);
 
     return (
       <audio ref={mediaElRef} {...restProps}>
--- a/packages/mux-audio-react/src/index.tsx
+++ b/packages/mux-audio-react/src/index.tsx
@@ -11,9 +11,12 @@ import {
 } from "@mux-elements/playback-core";
 import { getPlayerVersion } from "./env";
 
-export type Props = React.DetailedHTMLProps<
-  React.AudioHTMLAttributes<HTMLAudioElement>,
-  HTMLAudioElement
+export type Props = Omit<
+  React.DetailedHTMLProps<
+    React.AudioHTMLAttributes<HTMLAudioElement>,
+    HTMLAudioElement
+  >,
+  "autoPlay"
 > &
   MuxMediaProps;
 
@@ -33,6 +36,7 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
       startTime,
       src: outerSrc,
       children,
+      autoPlay,
       ...restProps
     } = props;
 
@@ -59,7 +63,7 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
         playerInitTime,
         playerSoftwareName,
         playerSoftwareVersion,
-        autoplay: props.autoPlay,
+        autoplay: autoPlay,
       };
       const nextPlaybackEngineRef = initialize(
         propsWithState,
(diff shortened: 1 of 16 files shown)


packages/mux-audio-react/src/index.tsx
```diff
@@ -11,9 +11,12 @@ import {
...
 
-export type Props = React.DetailedHTMLProps<
-  React.AudioHTMLAttributes<HTMLAudioElement>,
-  HTMLAudioElement
+export type Props = Omit<
+  React.DetailedHTMLProps<
+    React.AudioHTMLAttributes<HTMLAudioElement>,
+    HTMLAudioElement
+  >,
+  "autoPlay"
```
gkatsev, line 19: here, we want to use our definition of `autoPlay` from `MuxMediaProps` rather than the native one in React.

packages/mux-audio-react/src/index.tsx
```diff
@@ -33,6 +36,7 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
       startTime,
       src: outerSrc,
       children,
+      autoPlay,
```
gkatsev, line 39: this removes `autoPlay` from `restProps`, so, it doesn't get added to the `<audio>` element itself.

packages/mux-player-react/src/index.tsx
```diff
@@ -22,7 +22,7 @@ type VideoApiAttributes = {
   playsInline: boolean;
   // preload: string;
   crossOrigin: string;
-  autoPlay: boolean;
+  autoPlay: boolean | string;
```
gkatsev, line 25: this potentially should grab the type from MuxMediaProps.

packages/mux-video-react/src/index.tsx
```diff
@@ -11,9 +11,12 @@ import {
 } from "@mux-elements/playback-core";
 import { getPlayerVersion } from "./env";
 
-export type Props = React.DetailedHTMLProps<
-  React.VideoHTMLAttributes<HTMLVideoElement>,
-  HTMLVideoElement
+export type Props = Omit<
```
gkatsev, line 14: same changes as in `mux-audio-react`. Use our own `autoPlay` type.

packages/mux-video-react/src/index.tsx
```diff
@@ -33,6 +36,7 @@ const MuxVideo = React.forwardRef<HTMLVideoElement | undefined, Partial<Props>>(
       startTime,
       src: outerSrc,
       children,
+      autoPlay,
```
gkatsev, line 39: Like in mux-audio-react, make sure that the prop doesn't get added to the `<video>` element.

packages/playback-core/src/index.ts
```diff
@@ -66,6 +77,8 @@ export type MuxMediaPropTypes = {
     | `${Uppercase<keyof MimeTypeShorthandMap>}`;
   streamType: ValueOf<StreamTypes>;
   startTime: HlsConfig["startPosition"];
+  autoPlay: boolean | ValueOf<AutoplayTypes>;
+  autoplay: boolean | ValueOf<AutoplayTypes>;
```
gkatsev, line 81: I'm not sure if there's a way to combine these or eliminate the duplication here.
It's likely possible if we wanted to use `autoPlay` only and then reference to it as `props.autoPlay` below on lines 319-441

packages/playback-core/src/index.ts
```diff
@@ -74,9 +87,7 @@ declare global {
   }
 }
 
-export type HTMLMediaElementProps = Partial<
-  Pick<HTMLMediaElement, "src" | "autoplay">
```
gkatsev, line 78: `autoplay` isn't needed here, from what I can tell.

packages/playback-core/src/index.ts
cjpillsbury: yeah there probably is (or at least, moving the definition relationships around), but I don't think we need to prioritize that now... may lead to some "lying" hints/false positives, but I can pull on the definitional threads as a followup.

packages/playback-core/src/index.ts
cjpillsbury: hmm... this may be one of the aforementioned relationships/definitional threads. I'm cool with removing it if you confirmed that this won't cause issues of e.g. typescript + `mux-video` _*usage*_ of the autoplay _*prop*_. Given your updated definitions above, I think we should be fine though (I can help verify).

packages/playback-core/src/index.ts
```diff
@@ -263,7 +274,13 @@ export const loadMedia = (
...
     Pick<
       MuxMediaProps,
-      "preferMse" | "src" | "type" | "startTime" | "streamType" | "autoplay"
+      | "preferMse"
+      | "src"
+      | "type"
+      | "startTime"
+      | "streamType"
+      | "autoplay"
+      | "autoPlay"
```
cjpillsbury, line 283: nit: I don't think we need `autoPlay` here.

packages/playback-core/src/index.ts
```diff
@@ -392,6 +409,36 @@ export const loadMedia = (
...
+          // mute the player and then try playing
+          // if that fails, restore muted state
+          case AutoplayTypes.MUTED:
+            mediaEl.muted = true;
+            mediaEl.play().catch(restoreMuted);
+
+            break;
+          // Default or if autoplay is a boolean attribute:
+          // Try playing the video and catch the failed autoplay warning
+          default:
```
cjpillsbury, line 437: nit (non-blocking): Since we're not guaranteed that folks are using typescript and since TS doesn't provide runtime type checking, this means we'd treat any truthy autoplay value that isn't `"any"` or `"muted"` as `autoplay=true`. Probably not a huge deal, but just calling it out.

cjpillsbury (changes requested):
One small nit, but otherwise 👍 assuming sufficient smoke testing has been done.

packages/playback-core/src/index.ts
gkatsev: yeah, I was just thinking that these aren't necessary anymore. I'll try it out tomorrow.

packages/playback-core/src/index.ts
gkatsev: Yeah, I think it isn't necessary, but I'll double-check tomorrow. If I set `autoplay` on `mux-video` in one of the demos, I don't think it complains about this.

packages/playback-core/src/index.ts
gkatsev: I think that would be correct behavior anyway? Since it's a boolean attribute, and we want to maintain backwards compatibility here.

packages/playback-core/src/index.ts
cjpillsbury: yeah it's a "prop" vs. "attribute" case here, so you'd need to confirm in the TypeScript/JS use case of the element instance, but I think we should be fine (🤞)

packages/playback-core/src/index.ts
cjpillsbury: these are instance properties, not attributes.... you may still be right, just want to make sure the two aren't being conflated.

gkatsev: ah, good call. Another thing to check tomorrow, then.

packages/playback-core/src/index.ts
```diff
@@ -392,6 +409,36 @@ export const loadMedia = (
           );
         }
       });
+    } else if (props.autoplay) {
+      hls.once(Hls.Events.MANIFEST_PARSED, () => {
```
luwes, line 413: should the case if hls.js is not used be added?

packages/mux-player/src/index.ts
```diff
@@ -257,27 +256,6 @@ class MuxPlayerElement extends VideoApiElement {
     }
   }
 
-  #setUpMutedAutoplay() {
```
luwes, line 260: 👍

packages/playback-core/src/index.ts
gkatsev: Does playback-core ever not use hls.js?

cjpillsbury: yeah good call/catch @luwes! It doesn't use hls.js when an env supports native by default (unless overridden by `preferMse`

gkatsev: aah, right. Totally forgot about that.

cjpillsbury (approved):
🏄‍♂️

luwes (approved):
👍

PR #118, opened 2022-03-04
Fix preferMse naming bug & fix error dialog transition

luwes:
Fixes some bugs. One was causing preferMse not to work for `mux-video` WC.


packages/mux-player/src/media-chrome/dialog.ts
```diff
@@ -21,16 +21,18 @@ const styles = `
       linear-gradient(to bottom, rgba(20, 20, 30, 0.7) 50%, rgba(20, 20, 30, 0.9))
     );
     /* Needs to use !important to prevent overwrite of media-chrome */
-    transition: var(--media-dialog-transition-open, opacity .2s) !important;
+    transition: var(--media-dialog-transition-open, visibility .2s, opacity .2s) !important;
     transform: var(--media-dialog-transform-open, none) !important;
+    visibility: visible !important;
```
luwes, line 26: double checked this is cool w/ accessibility https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

cjpillsbury (approved):
🏄‍♂️

PR #120, opened 2022-03-07
Time Display: show the ascending time instead of negative time remaining

dylanjha:
This seems like a much more intuitive way to show time in the `<media-time-display>`. This feels like a better default. Maybe in the future we allow this to be configurable.

We don't show time display on live, so this should only impact VOD (and all my screenshots in the docs 😭 )

Thoughts?

# Before

time-remaining_2022-03-07_12-37-41

# After

no-tr_2022-03-07_12-35-55

### DIFF
--- a/examples/vanilla-ts-esm/package.json
+++ b/examples/vanilla-ts-esm/package.json
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "@mux-elements/mux-audio": "^0.2.0",
-    "@mux-elements/mux-player": "^0.1.0-alpha.3",
+    "@mux-elements/mux-player": "^0.1.0-alpha.7",
     "@mux-elements/mux-video": "^0.2.0"
   }
 }
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -1,6 +1,5 @@
 import "media-chrome";
 import "@mux-elements/mux-video";
-import MxpDialog from "./dialog";
 import VideoApiElement from "./video-api";
 import {
   getCcSubTracks,
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -1,4 +1,5 @@
 import "./dialog";
+import "./time-display";
 import {
   getChromeStylesFromProps,
   getSrcFromPlaybackId,
@@ -206,7 +207,7 @@ export const VodChromeSmall = (props: MuxTemplateProps) => html`
   </div>
   <media-control-bar>
     <media-time-range></media-time-range>
-    <media-time-display show-duration></media-time-display>
+    <mpx-time-display></mpx-time-display>
     ${MediaMuteButton()}
     ${props.supportsVolume && html`<media-volume-range></media-volume-range>`}
     <media-playback-rate-button></media-playback-rate-button>
@@ -224,7 +225,7 @@ export const VodChromeLarge = (props: MuxTemplateProps) => html`
     ${MediaSeekBackwardButton(props)}
     ${MediaSeekForwardButton(props)}
     <media-time-range></media-time-range>
-    <media-time-display show-duration></media-time-display>
+    <mxp-time-display></mxp-time-display>
     ${MediaMuteButton()}
     ${props.supportsVolume && html`<media-volume-range></media-volume-range>`}
     <media-playback-rate-button></media-playback-rate-button>
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -207,7 +207,7 @@ export const VodChromeSmall = (props: MuxTemplateProps) => html`
   </div>
   <media-control-bar>
     <media-time-range></media-time-range>
-    <mpx-time-display></mpx-time-display>
+    <mxp-time-display></mxp-time-display>
     ${MediaMuteButton()}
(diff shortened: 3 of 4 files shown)


examples/vanilla-ts-esm/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "@mux-elements/mux-audio": "^0.2.0",
-    "@mux-elements/mux-player": "^0.1.0-alpha.3",
+    "@mux-elements/mux-player": "^0.1.0-alpha.7",
```
dylanjha, line 23: This example code in my repo wasn't correctly linked until I changed this, then it was picking up my changes.

Am I doing something wrong in my local setup? Do I need to run `npm link` or something?

dylanjha:
@gkatsev very interesting idea. Feels like maybe a feature request for Media Chrome. We could probably implement that in Mux Player first, but as a small feature like that maybe it makes sense as an option directly in MC.

Will see what others think about that

gkatsev:
Implementing it here and then moving it into media-chrome sounds good to me. If it's in media-chrome, it probably should be configurable.
I think I first saw something like this in VLC. VLC also apparently gives you a jump-to-time dialog when you double-click on that component.

cjpillsbury:
> Seems like a reasonable, and arguably better, default. One addition, though, is to toggle between remaining or current on click.

Yeah I've def seen that in other players.

cjpillsbury:
@dylanjha overall I don't feel strongly about this either way (feels like a Jason/user feedback question imo).

dylanjha:
@cjpillsbury Yup, I looped i Jason about this

Added functionality for toggling (would love to port this as an option to Media Chrome)

https://user-images.githubusercontent.com/764988/157128097-a81864ab-4309-4a2f-bf14-41d5f225f0c7.mp4

packages/mux-player/src/index.ts
```diff
@@ -1,6 +1,5 @@
 import "media-chrome";
 import "@mux-elements/mux-video";
-import MxpDialog from "./dialog";
```
dylanjha, line 3: @luwes we have this import in the template, so I don't think we need it here. I think it was getting removed during the build process

packages/mux-player/src/time-display.ts
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    this.addEventListener("keydown", (e) => {
+      const { metaKey, altKey, key } = e;
+      if (metaKey || altKey || !ButtonPressedKeys.includes(key)) {
+        this.removeEventListener("keyup", keyUpHandler);
+        return;
+      }
+      this.addEventListener("keyup", keyUpHandler);
+    });
```
dylanjha, line 58: borrowed this keyboard handling code (a11y requirement) from media-chrome-button

https://github.com/muxinc/media-chrome/blob/dd27a2a6620c804f51a6d496156805043edb6a44/src/js/media-chrome-button.js

packages/mux-player/src/time-display.ts
```diff
@@ -0,0 +1,69 @@
...
+    cursor: pointer;
+  }
+`;
+
+const template = document.createElement("template");
+template.innerHTML = `
+  <style>
+    ${styles}
+  </style>
+  <media-time-display show-duration></media-time-display>
```
cjpillsbury, line 12: non-blocking: proposed "extending" `media-time-display` instead of using it as a descendant in the template, but that's not a high priority imo. fwiw I suspect if we wanted this functionality migrated over to Media Chrome, the code would more closely resemble "extending".

cjpillsbury (approved):
😎

packages/mux-player/src/template.ts
```diff
@@ -206,7 +207,7 @@ export const VodChromeSmall = (props: MuxTemplateProps) => html`
   </div>
   <media-control-bar>
     <media-time-range></media-time-range>
-    <media-time-display show-duration remaining></media-time-display>
+    <mpx-time-display></mpx-time-display>
```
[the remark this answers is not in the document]
luwes: or `mx-` 🏍️ 
or `bmx-` 🚴‍♂️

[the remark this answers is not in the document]
gkatsev: I've done the mpx/mxp thing before, too, heh

gkatsev (commented):
Functionality wise, from mouse and keyboard users works well.
From a Screen Reader perspective, it's a bit unclear. There's no indication that you can interact with this control to toggle behavior. Though, given that it reads the same thing to the user regardless of what mode it's in, it probably doesn't matter?

On a separate note, I'm not sure that the media-time-display should be of role progress bar, but that's a media-chrome issue.

luwes (changes requested):
All looks good to me. just the WC renaming since it's visible in the source

packages/mux-player/src/time-display.ts
cjpillsbury: yeah was thinking we shouldn't over architect it if this is expected to make its way into the built in functionality of `media-time-display`

gkatsev: Sure, though, this does beg the question, can/should you extend media-chrome-button, and are the chrome buttons purely for icons or can they be used with text too? I.e., if someone wanted to make a custom button, what's their workflow like? Though, definitely a media-chrome question. I can raise a discussion there.

luwes (approved):

PR #123, opened 2022-03-08
fix/playerInitTime-to-mux-embed: actually define the right property n…

cjpillsbury:
…ame for playerInitTime in mux-audio and mux-video (WC).


dylanjha (approved):
Took me a second to see how this flowed through to playback-core, but I see this gets passed as props to `initialize`

PR #124, opened 2022-03-08
Align more with IntlMessageFormat

luwes:
This change prepares for the potential use of IntlMessageFormat in the future.
See https://formatjs.io/docs/intl-messageformat/

It also fixes eslint for Typescript and adds a tiny eslint plugin so it enforces us not to use template literal expressions (substitutions) in the i18n strings.


scripts/eslint-plugin-i18n/no-substitutions.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  rules: {
+    "no-substitutions": {
+      create: function (context) {
+        return {
+          TaggedTemplateExpression(node) {
+            const tag = node.tag.name;
+            if (tag === "i18n") {
```
cjpillsbury, line 8: rock star.

cjpillsbury (approved):

PR #128, opened 2022-03-09
Run GH action jobs in parallel, add yarn cache, add lint job

luwes:
- Adds `--ignore-scripts` because we don't want an automatic build to happen right after yarn install
- Adds a lint job 
- Separates `build` and `test` so they can run in parallel
- Adds `cache: yarn` 
  https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
           node-version: ${{ steps.get-versions.outputs.node }}
           cache: yarn
       - run: yarn install --frozen-lockfile --ignore-scripts
-      - run: yarn build
+      - run: yarn build --scope @mux-elements/*
 
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,6 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ steps.get-versions.outputs.node }}
-      - run: yarn install --frozen-lockfile
+      - run: yarn install --frozen-lockfile --ignore-scripts
       - run: yarn build
       - run: yarn test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ steps.get-versions.outputs.node }}
+          cache: yarn
       - run: yarn install --frozen-lockfile --ignore-scripts
       - run: yarn build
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -16,6 +16,41 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ steps.get-versions.outputs.node }}
+          cache: yarn
+      - run: yarn install --frozen-lockfile --ignore-scripts
```
cjpillsbury, line 20: so we're installing the same thing multiple times in parallel and also rebuilding multiple times directly and indirectly? I don't think that's what we want…

.github/workflows/ci.yml
```diff
@@ -16,6 +16,41 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ steps.get-versions.outputs.node }}
+          cache: yarn
```
cjpillsbury, line 19: blocking (proposal): I think we should hold off on using yarn cache, at least for now… While it's a nice optimization, it makes builds non-idempotent wrt its dependencies and forcing a refetch of the cached version is difficult at best.

cjpillsbury (changes requested):
Have some questions…

.github/workflows/ci.yml
luwes: installing is mitigated by using the yarn cache, most of the time it's a low cost and it's not like it's an expense for us.
since this runs in parallel, the separate jobs will be completed faster than serially.

what do you mean with rebuilding? for the lint job no build is done, for the test job only the `prepare` script is run which builds the Mux elements. not the examples

.github/workflows/ci.yml
luwes: what would be the issue with using the yarn cache? it speeds up package retrieval
see https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-dependencies
we use this all the time locally

gkatsev: Since we use a lockfile, I don't see an issue with caching.

.github/workflows/ci.yml
```diff
@@ -16,6 +16,41 @@ jobs:
...
+      - run: yarn install --frozen-lockfile --ignore-scripts
+      - run: yarn build --scope @mux-elements/*
```
gkatsev, line 21: I think it would be good to run the build on the example projects too, perhaps in parallel. It has caught type issues for us previously related to props usage etc.

dylanjha (approved):

cjpillsbury (approved):

PR #129, opened 2022-03-09
ci: add codecov to CI

luwes:
lets see what it does, probably won't work without the app in
https://github.com/muxinc/elements/settings/installations


luwes:
@dylanjha the 79% seems accurate. 

the difference with locally is that it includes all  the node_modules packages and then the coverage shows it in the 50%'s. Locally I could get a similar percentage with this web test runner coverageConfig.include property setting it to 

```js
  coverageConfig: {
    report: true,
    include: ["src/**/*"],
  },
```

but then when this was run on GH actions Codecov would not upload correctly and result in an url like
https://codecov.io/github/muxinc/elements/commit/4e38e86fcb8eb1ecd6de26569174ea6402a59d2e

.github/workflows/ci.yml
```diff
@@ -17,5 +17,9 @@ jobs:
         with:
           node-version: ${{ steps.get-versions.outputs.node }}
       - run: yarn install --frozen-lockfile
-      - run: yarn build
+      - run: yarn build --scope @mux-elements/*
```
gkatsev, line 20: there's a benefit to running build on the examples, though, maybe it should be solved as part of the parallel build PR.
The main benefit I see is that it has caught some type of issues that were unnoticed in the main project.

packages/mux-audio/src/index.ts
```diff
@@ -6,13 +6,15 @@ import {
...
   ValueOf,
-  ExtensionMimeTypeMap,
   toMuxVideoURL,
   teardown,
-  PlaybackEngine,
   Metadata,
   mux,
-  type UpdateAutoplay,
+} from "@mux-elements/playback-core";
+import type {
```
gkatsev, line 14: question(non-blocking): why move the type imports into a separate declaration?

packages/mux-audio/package.json
```diff
@@ -12,6 +13,7 @@
   "author": "Mux, Inc.",
   "license": "MIT",
   "scripts": {
+    "test": "web-test-runner **/*test.js --port 8003 --coverage --config test/web-test-runner.config.js",
```
gkatsev, line 16: seems like web-test-runner can't choose a port automatically? That's a shame.

packages/mux-audio/src/index.ts
luwes: web-test-runner was complaining about this. not sure why but this fixed it

gkatsev: wonder if it's something like an esbuild version mismatch and they're using an older esbuild than we do.

packages/mux-audio/test/web-test-runner.config.js
```diff
@@ -0,0 +1,16 @@
+import { esbuildPlugin } from "@web/dev-server-esbuild";
```
gkatsev, line 1: is it possible to have a single config at the top level and then reference it in that location? Since these config files seem to be all the same.

codecov.yml
```diff
@@ -0,0 +1,70 @@
+# Setting coverage targets per flag
+coverage:
+  status:
+    project:
+      default:
+        target: auto #overall project/ repo coverage
+      mux-audio:
```
cjpillsbury, line 7: non-blocking: Looks like this is the recommended way of handling monorepos, but if possible, we should find a way to more easily keep the packages in sync with e.g. our `/packages`

packages/mux-audio/package.json
```diff
@@ -27,8 +30,12 @@
   },
   "devDependencies": {
     "@mux-elements/polyfills": "0.1.0",
+    "@open-wc/testing": "^3.0.3",
+    "@web/dev-server-esbuild": "^0.2.16",
+    "@web/test-runner": "^0.13.26",
     "esbuild": "^0.13.13",
     "npm-run-all": "^4.1.5",
+    "replace": "^1.2.1",
```
cjpillsbury, line 38: non-blocking: We've now got `devDependencies` on `copyfiles` and `replace`. We may want to just consider adding `shx` and using it for our "shell-equivalent use cases"

packages/mux-audio/src/CustomAudioElement.js
```diff
@@ -41,6 +41,7 @@ class CustomAudioElement extends HTMLElement {
   constructor() {
     super();
 
+    this.attachShadow({ mode: "open" });
```
cjpillsbury, line 44: ❤️

packages/mux-audio/src/index.ts
```diff
@@ -6,13 +6,15 @@ import {
...
+  PlaybackEngine,
+  UpdateAutoplay,
+  ExtensionMimeTypeMap,
 } from "@mux-elements/playback-core";
```
cjpillsbury, line 18: ❤️

packages/mux-audio/src/index.ts
```diff
@@ -147,6 +149,36 @@ class MuxAudioElement
...
+  set envKey(val: string | undefined) {
+    // dont' cause an infinite loop
+    if (val === this.envKey) return;
+
+    if (val) {
+      this.setAttribute(Attributes.ENV_KEY, val);
+    } else {
+      this.removeAttribute(Attributes.ENV_KEY);
+    }
+  }
```
cjpillsbury, line 180: ❤️

packages/mux-audio/test/web-test-runner.config.js
cjpillsbury: I'd say we should investigate that as a followup to not scope creep this PR, but I like the idea of in general trying to share more of our dev configs (tsconfig, esbuild, wtr, etc.)

packages/mux-player/package.json
```diff
@@ -14,7 +14,8 @@
   "license": "MIT",
   "scripts": {
     "lint": "eslint src/ --ext .js,.jsx,.ts,.tsx",
-    "test": "web-test-runner **/*test.js --coverage --config test/web-test-runner.config.js",
+    "test": "web-test-runner **/*test.js --port 8001 --coverage --config test/web-test-runner.config.js",
```
cjpillsbury, line 17: non-blocking: assuming this is because they run in parallel on the github action. This'll also be a pain to maintain over time. I suspect we should be able to do something clever in the web-test-runner.config file(s) to detect and/or otherwise programmatically assign port numbers for our packages.

packages/mux-video/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@mux-elements/mux-video",
   "version": "0.3.0",
-  "type": "commonjs",
+  "type": "module",
```
cjpillsbury, line 4: blocking: This is not accurate if our `main` is `dist/index.js`. We've been neglecting confirming different env usages, but we should probably have a followup effort there. Is there a reason you made this change in this PR?

packages/mux-video/src/index.ts
```diff
@@ -6,14 +6,16 @@ import {
...
-  PlaybackEngine,
   mux,
   MediaError,
-  type UpdateAutoplay,
+} from "@mux-elements/playback-core";
+import type {
+  PlaybackEngine,
+  UpdateAutoplay,
+  ExtensionMimeTypeMap,
 } from "@mux-elements/playback-core";
```
cjpillsbury, line 19: ❤️

packages/playback-core/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.2.0",
   "type": "commonjs",
   "description": "Core library for media playback in the browser shared by mux elements",
+  "module": "dist/index.mjs",
```
cjpillsbury, line 6: 👍

cjpillsbury (changes requested):
One blocking ask/confirmation. Otherwise I think we should merge this ASAP since we can always iterate on it.

packages/mux-audio/package.json
luwes: shx looks cool, I'll do on next iteration 👍

packages/mux-video/package.json
luwes: yes it was needed for web-test-runner to run mux-video's tests, otherwise it fails.

SCR-20220314-e28

would the esm bundle cause issues though? I think it's preferred actually

gkatsev: it would potentially be a breaking change, since node/bundlers will start treating the IIFE code as ESM code.

On a related note, should `main` be pointing at the `iife` build or the `cjs` build?

cjpillsbury (approved):

PR #130, opened 2022-03-09
fix: make set autoplay prop same than all str prop

luwes:
Just a tiny PR to test out the auto releasing.

Currently the behavior for the mux-player properties was that it casts them to strings no matter what the value is.
No removal of attributes based on nullish value, see https://github.com/muxinc/elements/issues/66

This change makes it the same than the other props like crossOrigin
https://github.com/muxinc/elements/blob/ef1f004f4a5b736837e78af682be75b3824806e3/packages/mux-player/src/video-api.ts#L244-L246

### DIFF
--- a/packages/mux-player/src/video-api.ts
+++ b/packages/mux-player/src/video-api.ts
@@ -250,15 +250,7 @@ class VideoApiElement extends HTMLElement {
   }
 
   set autoplay(val) {
-    if (val) {
-      this.setAttribute(
-        AllowedVideoAttributes.AUTOPLAY,
-        typeof val === "string" ? val : ""
-      );
-    } else {
-      // Remove boolean attribute if false, 0, '', null, undefined.
-      this.removeAttribute(AllowedVideoAttributes.AUTOPLAY);
-    }
+    this.setAttribute(AllowedVideoAttributes.AUTOPLAY, `${val}`);
   }
 
   get loop() {


packages/mux-player/src/video-api.ts
```diff
@@ -250,15 +250,7 @@ class VideoApiElement extends HTMLElement {
...
-    if (val) {
-      this.setAttribute(
-        AllowedVideoAttributes.AUTOPLAY,
-        typeof val === "string" ? val : ""
-      );
-    } else {
-      // Remove boolean attribute if false, 0, '', null, undefined.
-      this.removeAttribute(AllowedVideoAttributes.AUTOPLAY);
-    }
+    this.setAttribute(AllowedVideoAttributes.AUTOPLAY, `${val}`);
```
cjpillsbury, line 253: blocking: `autoplay` can be either a string or a boolean… I think this isn't quite what we want here.

cjpillsbury (changes requested):
callout string / boolean

packages/mux-player/src/video-api.ts
luwes: a boolean is an empty string and it's up to the user to set that I think. it's more inline with the native HTML elements

gkatsev: as long when in mux-player-react when the prop is set to `null` this attribute will end up getting removed.

cjpillsbury: if I have a `<video/>` tag and use its setter for `autoplay`, I can/would use `videoEl.autoplay = true` or `videoEl.autoplay = false`, so I'd assume that would be more in line with native HTML elements when it comes to setters, right?

luwes: hmm yes, autoplay = false would still enable it. I guess this is an edge case with different types could be used

closing then

PR #131, opened 2022-03-09
Fix/ssr error dialogs

cjpillsbury:
This PR fixes the issue exposed from stream.new and adds additional infrastructure to help smoke test against this. The root cause was another cases of missing polyfills (e.g. CustomEvent added to playback-core). This wasn't exposed in any of our demo apps, including our nextjs one, because it only was surfaced as a result of loading the page/MuxPlayer src using `dynamic`. This PR includes:

* creating a new shared polyfills private package.
* use the polyfills in mux-video, mux-audio, and playback-core
* Add a known unavailable playback-id to our media assets for testing.
* Add a dynamic load example page to our nextjs demo app for testing.


shared/polyfills/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@mux-elements/polyfills",
+  "version": "0.1.0",
+  "main": "./dist/index.js",
```
luwes, line 4: since this is such a lightweight package, can we use `./src/index.ts`	as export and skip all the build scripts?
feels a bit heavy here

shared/polyfills/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "author": "Mux, Inc.",
+  "license": "MIT",
+  "private": true,
+  "scripts": {
+    "dev:cjs": "open-process | yarn build:cjs --watch",
+    "dev:types": "yarn build:types -w",
+    "dev": "npm-run-all --parallel dev:types dev:cjs",
+    "build:cjs": "esbuild src/** --target=es2019 --format=cjs --outdir=dist",
+    "build:types": "tsc --declaration --emitDeclarationOnly --outDir './dist/types'",
+    "build": "npm-run-all --parallel build:types build:cjs"
```
luwes, line 15: scripts could be removed

packages/mux-video/src/CustomVideoElement.js
```diff
@@ -1,4 +1,4 @@
-import "./polyfills/window";
+import "@mux-elements/polyfills/dist/window";
```
luwes, line 1: this could be moved to mux-video's index.ts file to import a `.ts` file instead

luwes (commented):
the build step for this shared package seems not really needed, thoughts?

shared/polyfills/package.json
cjpillsbury: I think if we're worried about ts vs. js, load, we can probably just migrate the polyfill to js. I'd recommend we do that as a followup since this feels like a lower priority.

packages/mux-video/src/CustomVideoElement.js
cjpillsbury: I feel pretty strongly about not adding the polyfills far away from where they're actually needed.

shared/polyfills/package.json
luwes: you're right yes, that makes the most sense. writing the polyfill in plain JS. no need for any extra build steps then

gkatsev: if we're thinking of re-writing into plain JS. Thoughts about having the files at the root of the package only? Then import can be `@mux-elements/polyfills/window` without worrying about `dist/` directories etc.

gkatsev (approved):

PR #135, opened 2022-03-14
Add async to script tag

blairliikala:
When the script tag is being used and placed in the HTML header, the player doesn't load quite right.  I see a pause button, but can't control any UI of the player.  When the tag is below the mux-player element, like in the footer, it loads correctly.  This is specifically with browser caching disabled between page loads, or when loading the player script for the first time.  Since it partially loads it might confuse a first-time user if they were trying to get a quick test or demo going.  Adding async to the script tag also solves the issue if the script tag is above or below the element, and is best practice.


dylanjha:
This should be resolved by https://github.com/muxinc/media-chrome/pull/194 which will be in the `beta.1` release for Mux Player

PR #137, opened 2022-03-16
build: add clean script for all packages and workspace. Clean up dev …

cjpillsbury:
…dependencies. Don't build on install (prepare).

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "predev": "yarn build:packages",
     "build:packages": "lerna run build --scope @mux-elements/*",
     "build": "lerna run build",
+    "prepare": "husky install && yarn build --scope @mux-elements/*",
     "release": "lerna publish --no-private",
     "version:update": "lerna version --no-private"
   }
--- a/package.json
+++ b/package.json
@@ -24,12 +24,12 @@
     "shared/*"
   ],
   "scripts": {
+    "clean": "lerna run clean --parallel --scope @mux-elements/*",
     "format": "prettier --write .",
     "lint": "lerna run lint --scope @mux-elements/*",
     "test": "lerna run test --scope @mux-elements/*",
     "dev": "lerna run dev --parallel --scope @mux-elements/*",
     "build": "lerna run build",
-    "prepare": "husky install && yarn build --scope @mux-elements/*",
     "release": "lerna publish --no-private",
     "version:update": "lerna version --no-private"
   }
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "predev": "yarn build:packages",
     "build:packages": "lerna run build --scope @mux-elements/*",
     "build": "lerna run build",
-    "prepare": "husky install && yarn build --scope @mux-elements/*",
+    "prepare": "husky install && yarn build:packages",
     "release": "lerna publish --no-private",
     "version:update": "lerna version --no-private"
   }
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -24,12 +24,12 @@
     "shared/*"
   ],
   "scripts": {
+    "clean": "lerna run clean --parallel --scope @mux-elements/*",
     "format": "prettier --write .",
     "lint": "lerna run lint --scope @mux-elements/*",
     "test": "lerna run test --scope @mux-elements/*",
     "dev": "lerna run dev --parallel --scope @mux-elements/*",
     "build": "lerna run build",
-    "prepare": "husky install && yarn build --scope @mux-elements/*",
```
gkatsev, line 32: we probably want `yarn build --scope @mux-elements/*` on `prepublishOnly` or something

packages/mux-player/package.json
```diff
@@ -43,6 +45,7 @@
     "eslint": "^7.11.0",
     "npm-run-all": "^4.1.5",
     "replace": "^1.2.1",
+    "shx": "^0.3.4",
```
gkatsev, line 48: Theoretically, we can have `shx` at the top level, and it'll be available for usage in all the workspace packages

package.json
```diff
@@ -24,12 +24,12 @@
     "shared/*"
   ],
   "scripts": {
+    "clean": "lerna run clean --parallel --scope @mux-elements/*",
     "format": "prettier --write .",
     "lint": "lerna run lint --scope @mux-elements/*",
     "test": "lerna run test --scope @mux-elements/*",
     "dev": "lerna run dev --parallel --scope @mux-elements/*",
     "build": "lerna run build",
```
gkatsev, line 32: should we add `"prebuild": "clean"`

packages/playback-core/package.json
```diff
@@ -14,6 +14,7 @@
   "author": "Mux, Inc.",
   "license": "MIT",
   "scripts": {
+    "clean": "shx rm -rf dist/",
     "dev:cjs": "open-process | yarn build:cjs --watch",
     "dev:types": "yarn build:types -w",
     "dev": "npm-run-all --parallel dev:types dev:cjs",
```
gkatsev, line 20: need to add `dev:esm` and add it in here

packages/mux-player-react/package.json
```diff
@@ -34,6 +35,7 @@
     }
   },
   "dependencies": {
+    "@mux-elements/open-process": "0.1.0",
```
luwes, line 38: should be devDependencies

package.json
cjpillsbury: Yeah it's out of scope but I am down. Gonna refactor my last version to just rely on your proposed prebuild as well.

packages/mux-player/package.json
cjpillsbury: True. Philosophically, I prefer to add the dependencies where they are actually expected. It means version syncing/upgrades require some additional TLC but makes tracking and removing dependencies that are no longer used easier, and makes it clearer how dependencies are made available for where they're used.

gkatsev (approved):

packages/mux-player/package.json
cjpillsbury: I'd lean towards using something like https://github.com/JamieMason/syncpack for those parts of housekeeping.

PR #138, opened 2022-03-16
Fix/text tracks after playback id change

cjpillsbury:


luwes:
I agree that this seems the way to go longterm, it's not a crazy amount of code either.

I also like that the track elements will all be visible in the DOM, more in line with the Mux elements and MC vision.

packages/playback-core/src/index.ts
```diff
@@ -242,12 +243,46 @@ export const setupMux = (
...
+  trackEl.label = label;
+  if (lang) {
+    trackEl.srclang = lang;
+  }
+  if (id) {
+    trackEl.id = id;
+  }
+  trackEl.track.mode = 'disabled';
+  mediaEl.appendChild(trackEl);
+  trackEl.addEventListener('cuechange', console.log.bind(null, 'cue change', id));
```
gkatsev, line 267: we probably want to remove this before release

packages/playback-core/src/index.ts
```diff
@@ -323,6 +358,64 @@ export const loadMedia = (
...
+      const textTrack: TextTrack = Array.prototype.find.call(mediaEl.textTracks, (textTrack) => {
+        return textTrack.kind === type && textTrack.id === track;
+      });
+      if (!textTrack) return;
+      const disabled = textTrack.mode === 'disabled';
+      if (disabled) {
+        textTrack.mode = 'hidden';
+      }
+      cues.forEach((cue: VTTCue) => {
+        if (textTrack.cues?.getCueById(cue.id)) return;
```
gkatsev, line 386: Do we need to see if the cue exists already? Any issues with unconditionally adding the cue into the track? It would allow us not to fiddle with the track's mode.

PR #141, opened 2022-03-22
feat: move captions up when controls are showing

gkatsev:
Controls are currently defined as showing when the user is active or the player is paused.

This is more or less the MVP for this, as it only covers moving the bottom captions out of the way of the bottom controls.
The full thing would also probably need to handle the top controls as well as the middle controls.
I am also ignoring some cues depending on how they are positioned, as they are likely not going to be obscured by the bottom controls or thumbnails.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -100,6 +100,7 @@ const playerSoftwareName = 'mux-player';
 
 class MuxPlayerElement extends VideoApiElement {
   #tokens = {};
+  #userInactive: boolean;
   #resizeObserver?: ResizeObserver;
   #state: Partial<MuxTemplateProps> = {
     isDialogOpen: false,
@@ -142,6 +143,9 @@ class MuxPlayerElement extends VideoApiElement {
     this.#setUpCaptionsButton();
     this.#setUpAirplayButton();
     this.#setUpVolumeRange();
+
+    this.#userInactive = this.shadowRoot?.querySelector('media-controller')?.hasAttribute('user-inactive') ?? true;
+    this.#setUpCaptionsMovement();
   }
 
   connectedCallback() {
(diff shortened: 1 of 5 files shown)


packages/mux-player/src/index.ts
```diff
@@ -141,6 +141,98 @@ class MuxPlayerElement extends VideoApiElement {
...
+      selectedTrack?.addEventListener('cuechange', cuechangeHandler);
+      // selectedTrack?.addEventListener('cuechange', () => {mc.pause()}, {once: true});
+    };
+
+    selectTrack();
+    // update the selected track as necessary
+    mc?.media?.textTracks.addEventListener('change', selectTrack);
+    mc?.media?.textTracks.addEventListener('addtrack', selectTrack);
+
+    mc?.addEventListener('userinactivechange', () => {
```
gkatsev, line 225: this depends on a media-chrome release with https://github.com/muxinc/media-chrome/pull/182 and https://github.com/muxinc/media-chrome/pull/190

packages/mux-player/src/index.ts
```diff
@@ -141,6 +142,9 @@ class MuxPlayerElement extends VideoApiElement {
     this.#setUpCaptionsButton();
     this.#setUpAirplayButton();
     this.#setUpVolumeRange();
+
+    this.#userInactive = this.shadowRoot?.querySelector('media-controller')?.hasAttribute('user-inactive') || true;
```
luwes, line 146: does the condition need `??` instead of `||`. right now it will always be true

gkatsev: I tried to, but the build was complaining about it. I can try again.

packages/mux-player/src/index.ts
```diff
@@ -287,6 +291,97 @@ class MuxPlayerElement extends VideoApiElement {
...
+            cuesmap.set(cue, cue.line);
+          }
+
+          // we have to set line to 0 first due to a chrome bug https://crbug.com/1308892
+          cue.line = 0;
+          cue.line = setTo;
+        } else {
+          setTimeout(() => {
+            cue.line = cuesmap.get(cue) || 'auto';
+          }, 500);
```
gkatsev, line 341: this delay is so that the captions won't be partially obscured as the control bar is transitioning into being hidden.

packages/mux-player/src/index.ts
```diff
@@ -99,6 +99,7 @@ const playerSoftwareName = 'mux-player';
 
 class MuxPlayerElement extends VideoApiElement {
   #tokens = {};
+  #userInactive: boolean;
```
gkatsev, line 102: we need to store it for now because Media Chrome can fire multiple events for user inactive changed, without the value actually changing right now, so, we want to work around that.

packages/mux-player/src/index.ts
```diff
@@ -287,6 +291,97 @@ class MuxPlayerElement extends VideoApiElement {
...
+    selectTrack();
+    // update the selected track as necessary
+    mc?.media?.textTracks.addEventListener('change', selectTrack);
+    mc?.media?.textTracks.addEventListener('addtrack', selectTrack);
+
+    mc?.addEventListener('userinactivechange', () => {
+      const newUserInactive = mc?.hasAttribute('user-inactive');
+
+      if (this.#userInactive === newUserInactive) {
+        return;
```
gkatsev, line 376: codecov is saying that this isn't covered by tests. I think that's fine as ideally media-chrome will be fixed not to trigger duplicate events and only trigger when the property has changed.

packages/mux-player/test/player.test.js
```diff
@@ -348,3 +348,188 @@ describe('<mux-player>', () => {
...
+      muted
+    ></mux-player>`);
+
+    const mc = player.shadowRoot.querySelector('media-controller');
+    const media = mc.media;
+
+    // force user-inactive
+    mc.setAttribute('autohide', '0.1');
+    mc.setAttribute('user-inactive', '');
+    mc.dispatchEvent(new Event('userinactivechange'));
```
gkatsev, line 369: this may not be necessary once media-chrome is updated.

PR #144, opened 2022-03-22
build:silence vercel comments on github

dylanjha:


dylanjha:
Well, looks like `github.silent = true`, is the opposite of what we want. This config will prevent the comment, but all the noisy bot activity is still there.

vercel-github-comment

vercel-github-activity

I don't actually think Vercel is in control of this, it's on GitHub side. There's things like this: https://github.com/patridge/github-pull-request-minimizer

Issue #146, opened 2022-03-24
Feature: Support specifying `beaconCollectionDomain`


pchang211:
### Which Mux Elements/Packages would you like a new feature for?

- mux-video
- mux-audio

## 🎥 Describe the feature you'd like to request

Components should be able to specify `beaconCollectionDomain` in the player options so a custom domain can be used for Mux Data collection. The existing `beaconDomain` option will still work but will be deprecated in future versions.

## 💻 Describe the use cases the feature will solve

This will allow Data collection via a non-standard beacon domain.

## ⭐ Describe the solution you'd like to see

Pass `beaconCollectionDomain` through, similar to the legacy `beaconDomain` option.

PR #147, opened 2022-03-24
Add `beaconCollectionDomain` option

pchang211:
Issue: https://github.com/muxinc/elements/issues/146

Add `beaconCollectionDomain` (introduced in mux-embed 4.7.0) option to the Mux Player.


cjpillsbury:
Looks like `mux-player` package got missed. On a quick glance it looks like you'll need to:

- add to the list of attributes per our kebab naming convention (e.g. https://github.com/muxinc/elements/blob/main/packages/mux-player/src/index.ts#L46)
- add a `get`/`set` for the property that follows our convention for `mux-video`-owned props/attrs (e.g. https://github.com/muxinc/elements/blob/main/packages/mux-player/src/index.ts#L448-L457)
- add it to the list of `props` passed to our DOM template via `getProps` (see: https://github.com/muxinc/elements/blob/main/packages/mux-player/src/index.ts#L78)
- pass it as an attribute to the underlying `<mux-video/>` instance via our template (see: https://github.com/muxinc/elements/blob/main/packages/mux-player/src/template.ts#L59)

cjpillsbury:
Additionally, we have *just* added "conventional commits" support for our build/version/CD setup. If you could amend your commit msg(s) to conform to a [new feature](https://www.conventionalcommits.org/en/v1.0.0/#summary), that would be 👌. (tl;dr - msg should read `feat: [brief description of feature]`

packages/mux-audio/src/index.ts
```diff
@@ -23,6 +23,7 @@ type AttributeNames = {
   METADATA_VIDEO_ID: 'metadata-video-id';
   METADATA_VIDEO_TITLE: 'metadata-video-title';
   METADATA_VIEWER_USER_ID: 'metadata-viewer-user-id';
+  BEACON_COLLECTION_DOMAIN: 'beacon-collection-domain',
   BEACON_DOMAIN: 'beacon-domain';
```
dylanjha, line 27: I don't think we need to keep things backwards-compatible here tbh. This is an unannounced, undocumented feature so I think it's just cleaner to drop all references to `beacon-domain`

dylanjha (changes requested):

packages/mux-audio/src/index.ts
cjpillsbury: yeah I thought about that, too. Just figured we could sweep that up in a followup PR.

luwes (approved):
LGTM

packages/mux-video/src/index.ts
```diff
@@ -26,7 +26,7 @@ type AttributeNames = {
   METADATA_VIDEO_ID: 'metadata-video-id';
   METADATA_VIDEO_TITLE: 'metadata-video-title';
   METADATA_VIEWER_USER_ID: 'metadata-viewer-user-id';
-  BEACON_DOMAIN: 'beacon-domain';
+  BEACON_COLLECTION_DOMAIN: 'beacon-collection-domain',
```
cjpillsbury, line 29: nit (blocking): this should be `;`

packages/mux-audio/src/index.ts
```diff
@@ -23,7 +23,7 @@ type AttributeNames = {
   METADATA_VIDEO_ID: 'metadata-video-id';
   METADATA_VIDEO_TITLE: 'metadata-video-title';
   METADATA_VIEWER_USER_ID: 'metadata-viewer-user-id';
-  BEACON_DOMAIN: 'beacon-domain';
+  BEACON_COLLECTION_DOMAIN: 'beacon-collection-domain',
```
cjpillsbury, line 26: nit (blocking): this should be `;`

cjpillsbury (changes requested):
small nits (typescript defs using `,` instead of `;`

cjpillsbury (approved):

PR #148, opened 2022-03-25
fix: initial muted autoplay should autoplay muted

gkatsev:
https://github.com/muxinc/elements/pull/116 added extended autoplay support, but setting the initial attribute to one of the extended values didn't work. This fixes it so that the attribute can be set initially and not just changed after the fact.


cjpillsbury (approved):
😎

luwes (approved):

PR #150, opened 2022-03-28
Feat/seek to live

cjpillsbury:
Updates seek to live component:
- indicator changes from grey (not in live window) to red (in live window)
- is a button that seeks to live when pressed
- is "disabled" when currently in live window

Play button
- seeks to live when unpaused

### DIFF
--- a/packages/mux-player/src/helpers.ts
+++ b/packages/mux-player/src/helpers.ts
@@ -91,6 +91,11 @@ export const seekToLive = (el: MuxPlayerElement) => {
   el.currentTime = liveTime;
 };
 
+export const LL_LIVE_SEGMENT_SECS = 1;
+export const LIVE_SEGMENT_SECS = 2;
+export const DEFAULT_HOLDBACK = 3;
+export const LIVE_HOLDBACK_MOE = 0.5;
+
 export const isInLiveWindow = (el: MuxPlayerElement) => {
   const { streamType } = el;
   const liveTime = getLiveTime(el);
@@ -99,11 +104,13 @@ export const isInLiveWindow = (el: MuxPlayerElement) => {
     return false;
   }
   const delta = liveTime - currentTime;
+  // The live window is based on whether or not the current playhead is within n segment durations (plus a margin of error)
+  // of the live edge (CJP)
   if (streamType === StreamTypes.LL_LIVE) {
-    return delta <= 1 * 3.5;
+    return delta <= LL_LIVE_SEGMENT_SECS * (DEFAULT_HOLDBACK + LIVE_HOLDBACK_MOE);
   }
   if (streamType === StreamTypes.LIVE) {
-    return delta <= 2 * 3.5;
+    return delta <= LIVE_SEGMENT_SECS * (DEFAULT_HOLDBACK + LIVE_HOLDBACK_MOE);
   }
   return false;
 };
--- a/packages/mux-player/src/helpers.ts
+++ b/packages/mux-player/src/helpers.ts
@@ -1,5 +1,6 @@
 import { stylePropsToString, toQuery, camelCase } from './utils';
 import type MuxPlayerElement from '.';
+import { StreamTypes } from './constants';
 
 /* eslint-disable */
 const getEnvPlayerVersion = () => {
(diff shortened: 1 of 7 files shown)


packages/mux-player/src/helpers.ts
```diff
@@ -62,6 +63,39 @@ export function getCcSubTracks(el: MuxPlayerElement) {
...
+export const isInLiveWindow = (el: MuxPlayerElement) => {
+  const { streamType } = el;
+  const liveTime = getLiveTime(el);
+  const currentTime = el.video?.currentTime;
+  if (liveTime == undefined || currentTime == undefined) {
+    return false;
+  }
+  const delta = liveTime - currentTime;
+  if (streamType === StreamTypes.LL_LIVE) {
+    return delta <= 1 * 3.5;
```
cjpillsbury, line 91: lmk if some code comments would be good here. This is based on holdbacks, current mux segment vs. part sizes, and "fudge factor" due to RTT and related.

packages/mux-player/src/template.ts
```diff
@@ -114,8 +109,31 @@ export const ChromeRenderer = (props: MuxTemplateProps) => {
 };
 
 // prettier-ignore
-const MediaPlayButton = () => html`
-  <media-play-button>
+const SeekToLiveButton = (props: Partial<MuxTemplateProps>) => html`
+  <button 
+    aria-disabled="${props.inLiveWindow}"
+    in-live-window="${props.inLiveWindow}"
+    onclick="${function (this: HTMLButtonElement, evt: Event) {
```
cjpillsbury, line 116: doing this inline, but that means a new function will be generated whenever a (re)render occurs. We can "promote" this to props/state if this is a concern.

packages/mux-player/test/player.test.js
```diff
@@ -347,6 +347,88 @@ describe('<mux-player>', () => {
...
+      this.timeout(12000);
+      const playerEl = await fixture(`<mux-player
+        playback-id="v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM"
+        muted
+        stream-type="ll-live"
+      ></mux-player>`);
+
+      const mcPlayEl = playerEl.shadowRoot.querySelector('media-play-button');
+      // NOTE: Need try catch due to bug in play+autoplay behavior (CJP)
+      try {
```
cjpillsbury, line 420: this needs to be ticketed and is related to some of the bad autoplay behavior I noticed during previous efforts.

packages/mux-player/src/helpers.ts
gkatsev: I like having a constant for things like that. Makes it extra obvious what's happening here.

packages/mux-player/src/template.ts
```diff
@@ -114,8 +109,31 @@ export const ChromeRenderer = (props: MuxTemplateProps) => {
 };
 
 // prettier-ignore
-const MediaPlayButton = () => html`
-  <media-play-button>
+const SeekToLiveButton = (props: Partial<MuxTemplateProps>) => html`
+  <button 
+    aria-disabled="${props.inLiveWindow}"
+    in-live-window="${props.inLiveWindow}"
```
luwes, line 115: could `aria-disabled` not be used in the CSS again? 
adding a `in-live-window` attr to a native element is invalid HTML no, maybe use the class attribute?

luwes (dismissed):
use a constant for 3.5 and attribute nitpick

I guess it's fine in HTML5
https://stackoverflow.com/questions/1735230/can-i-add-a-custom-attribute-to-an-html-tag

packages/mux-player/src/template.ts
luwes: why does `this` need to be a function argument? isn't it accessible anyway via this

luwes: why does `this` need to be a function argument? isn't it accessible anyway via this

packages/mux-player/src/template.ts
```diff
@@ -114,8 +109,31 @@ export const ChromeRenderer = (props: MuxTemplateProps) => {
...
+    class="mxp-seek-to-live-button"
+  >
+    Live
+  </button>
+`;
+
+// prettier-ignore
+const MediaPlayButton = (props: Partial<MuxTemplateProps>) => html`
+  <media-play-button 
+    onmediaplayrequest="${props.streamType === StreamTypes.LIVE || props.streamType === StreamTypes.LL_LIVE ? props.onSeekToLive : false}"
```
luwes, line 135: blocking: I think this line can be removed, the prop onSeekToLive can be removed, in the above onclick handler too.

the seekToLive function can be called in the `play` event listener of the media element if needed.

packages/mux-player/src/template.ts
cjpillsbury: short version: "because typescript"

packages/mux-player/src/template.ts
cjpillsbury: cool this has product implications. Chatting w/ @dylanjha about them now. As long as we're ok with the broader assumptions here, I'll refactor.

the seek to live button does not _*necessarily*_ also play. E.g., if content is currently playing but is not in the live window, seek to live will seek but not play.

cjpillsbury: Also not sure we'd want to wait until the `play` event is fired before doing a seek to live. Per MDN:
> The play event is fired when the paused property is changed from true to false, as a result of the play method, or the autoplay attribute.
The current solution will begin the seek concurrent with the "request to play" (and should be earlier than the media-controller logic, due to bubbling). Your proposed solution would entail that we don't even begin to seek until **_after_** the media element has gone from paused to unpaused.

luwes: you're right, I think my initial feeling here was that these changes added more complexity than what we've seen in the template.ts file but realizing the reason is Media Chrome. 

The `play` event is indeed later but it would catch any attempt to resume playing. There's cases on iPhone for example with locking and unlocking where the internal video.pause() is called, this wouldn't go through our UI at all and it would not cause the player to jump to the live window. maybe not the worst until we would support DVR.

packages/mux-player/src/styles.css
```diff
@@ -117,8 +116,24 @@ media-time-display {
   white-space: nowrap;
 }
 
-.mxp-live-indicator::before {
+.mxp-seek-to-live-button {
+  background-color: var(----media-control-background);
```
luwes, line 120: got a little excited here 😄  `----media-control-background`

luwes (changes requested):
Comment or constants for 3.5 and fix up CSS variable

packages/mux-player/src/template.ts
dylanjha: Chiming in here with some comments on the expected behavior (I don't have an opinion on the technical implementation).

First of all, we decided that for `ll-live` and for `live` (non-DVR mode) that tapping the play button will jump to live edge.

**option 1**

- calling `play()` on the mux-player programmatically will also jump to live edge

**option 2**

- calling `play()` on the mux-player programmatically will NOT jump to live edge

Honestly, I don't feel strongly between option 1 and 2 and it feels like a minor detail without a clear expectation of what a developer expects to happen. The good thing is that with option 2 the user will still visually see the "seek to live" button, so nothing would end up in a broken state. For that reason, I'm not too worried about it.

Whichever option is technically simpler is what I think we should go with.

luwes (approved):

PR #157, opened 2022-03-29
build(build-when-needed): Update scripts to not redundantly build.

cjpillsbury:
- Removes build from workspace level `prepare`.
- Removes redundant/unnecessary builds from ci workflows
- Adds `build:packages` to workspace level `pretest` (Can discuss tradeoffs vs. per-package `pretest` vs. independent build in github action scripts)
- No longer `--ignore-scripts` on installs, as this is brittle and may have been working in some cases as a result of previous cached installs
  - `--ignore-scripts`: Do not execute any scripts defined in the project package.json **_and its dependencies_**. (From: https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-ignore-scripts)

After change approval but before merge, we will also need to change our vercel apps setup to `build:packages` on install.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -24,8 +24,6 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - name: Install
         run: yarn install --frozen-lockfile
-      - name: Build packages
-        run: yarn build:packages
       - name: Test
         run: yarn test
       - name: Publish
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -28,8 +28,6 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - name: Install
         run: yarn install --frozen-lockfile
-      - name: Build packages
-        run: yarn build:packages
```
luwes, line 32: 👍  prepublishOnly handles this

gkatsev (approved):

PR #161, opened 2022-03-30
chore: add release notes param to version:update

luwes:
Confirmed here that Lerna only creates release notes and changelogs in the actual version update command
https://github.com/lerna/lerna/blob/a47fc294393a3e9507a8207a5a2f07648a524722/commands/version/index.js#L495-L562

Opened a PR that could make it possible to add changelogs and release notes in CI here:
https://github.com/lerna/lerna/pull/3097

For now the deployer will have to add a GH_TOKEN locally and add the release notes from the `version:update` command.


gkatsev (changes requested):
We should update the release process https://github.com/muxinc/elements/blob/main/README_INTERNAL.md#releasing to mention getting the GH_TOKEN and how to run the command now

gkatsev (approved):

PR #169, opened 2022-04-01
fix: devLog links to GH base URL

dylanjha:
Right now devLog links are 404:

```
https://github.com/muxinc/elements/main/errors/403-playback-id-mismatch.md
```

This corrects the path:

```
https://github.com/muxinc/elements/blob/main/errors/403-playback-id-mismatch.md
```


cjpillsbury (approved):
👌

PR #171, opened 2022-04-05
fix: child custom element mux-video upgraded after mux-player

luwes:
This change adds a fix for a bug in React where the CE children of mux-player were seemingly upgraded later than mux-player.

Also adds a check whether mux-video and media-controller is successfully upgraded and logs an error if not.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -1,6 +1,7 @@
 import '@mux-elements/polyfills/window';
-import 'media-chrome';
-import { MediaError } from '@mux-elements/mux-video';
+// @ts-ignore
+import { MediaController } from 'media-chrome';
+import MuxVideoElement, { MediaError } from '@mux-elements/mux-video';
 import VideoApiElement from './video-api';
 import {
   getCcSubTracks,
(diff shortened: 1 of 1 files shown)


packages/mux-player/src/index.ts
```diff
@@ -145,6 +161,18 @@ class MuxPlayerElement extends VideoApiElement {
...
     //   this.video.hls.config.maxMaxBufferLength = 2;
     // }
+
+    this.#setUpErrors();
+    this.#setUpCaptionsButton();
+    this.#setUpAirplayButton();
+    this.#setUpVolumeRange();
+    this.#monitorLiveWindow();
+    this.#userInactive = this.mediaController?.hasAttribute('user-inactive') ?? true;
+    this.#setUpCaptionsMovement();
```
dylanjha, line 171: Non-blocking: This sort of ties into the conversation we had earlier. Constructors that do a lot of things generally feel like an anti-pattern in web component world. I'd be worried that any of these instance methods are calling `setAttribute` or otherwise making illegal DOM manipulations resulting in an error like: `Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes`

Luckily, none of these methods are doing anything illegal, so that's good, but just wanted to call it out that it's a potential "gotcha".

dylanjha (approved):

PR #172, opened 2022-04-05
fix(mux-video): prevent forward custom attrs

luwes:
This change prevents forwarding custom attributes to the native video element like player-software-name, env-key, etc...

SCR-20220321-iff


packages/mux-video/src/CustomVideoElement.js
```diff
@@ -136,8 +136,9 @@ class CustomVideoElement extends HTMLElement {
         this.nativeEl.removeAttribute(attrName);
       } else {
         // Ignore a few that don't need to be passed through just in case
-        // it creates unexpected behavior.
-        if (['id', 'class'].indexOf(attrName) === -1) {
```
dylanjha, line 140: probably not exhaustive, but a good start and doesn't hurt 👍🏼

dylanjha (approved):

PR #173, opened 2022-04-07
fix: primary color should be also used for the text color on controls

dylanjha:
Fixes a bug where `primaryColor` is not being applied to controls that have text (but it is being applied to controls that have icons via `--media-icon-color`.

Relates to the problem encountered here:

https://github.com/muxinc/media-chrome/discussions/198

# Before

before_2022-04-07_13-19-27

# After

after_2022-04-07_13-19-38


cjpillsbury (approved):
😎

PR #174, opened 2022-04-07
fix: clear some state on playbackId change

luwes:
this change clears the dialog state when changing playbackId

### DIFF
--- a/packages/mux-player/test/player.test.js
+++ b/packages/mux-player/test/player.test.js
@@ -371,6 +371,29 @@ describe('<mux-player> playbackId transitions', () => {
       'https://stream.mux.com/xLGf7y8cRquv7QXoDB02zEe6centwKfVmUOiPSY02JhCE.m3u8?redundant_streams=true'
     );
   });
+
+  it('loads the new playbackId and clears dialog state', async function () {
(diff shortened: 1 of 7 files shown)


packages/mux-player/test/player.test.js
```diff
@@ -347,87 +352,129 @@ describe('<mux-player>', () => {
...
-        playback-id="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"
-        stream-type="vod"
-      ></mux-player>`);
-
-      const mediaControllerEl = playerEl.shadowRoot.querySelector('media-controller');
-      const seekToLiveEl = playerEl.shadowRoot.querySelector('.mxp-seek-to-live-button');
-      assert.exists(mediaControllerEl);
-      assert.notExists(seekToLiveEl);
-    });
+describe('<mux-player> playbackId transitions', () => {
```
cjpillsbury, line 357: non-blocking: We should probably have tests for most or all props/attrs when 1) changing playbackIds and 2) unsetting a playbackId just to have a clearer definition of done for this effort. That said, I think this is still an improvement, so I don't think it should block the merge.

cjpillsbury (approved):
👍

PR #175, opened 2022-04-08
feat: add media-theme-mux as a replaceable element

luwes:
test URL: https://elements-demo-vanilla-git-fork-luwes-media-theme-mux-mux.vercel.app/mux-player.html

> Side note: we might have been too worried about the complexity of the problem of isolating a theme element from mux-player.

This change adds a new element `<media-theme-mux>` under the parent element `<mux-player>` which could be replaced by an other theme. The replacing logic is still a todo which would have to make use of the jtml renderer. I think it shouldn't be too hard. 

The gist of the the changes in this PR are: 
- Moved the theme related files in one folder called `media-theme-mux`: the SVG icons, the majority of styles.css, and most of the HTML in template.ts to a new file called media-theme-mux.ts which creates a new web component.
- Then determined which props were generic enough to keep in the theme and added them as props / attributes

```js
type ThemeMuxTemplateProps = {
  streamType: string;
  playerSize: string;
  defaultHiddenCaptions: boolean;
  hasCaptions: boolean;
  forwardSeekOffset: number;
  backwardSeekOffset: number;
};
```

- Removed some "supported" props and replaced them with Media Chrome's ability to hide based on CSS attribute selector.
- Things that were not added to the theme but via the default slot: mxp-dialog, via a named slot: seek-to-live-button

Below is the class of the theme it simply renders the html template based on the attrs/props.

```js
class MediaThemeMux extends HTMLElement {
  static get observedAttributes() {
    return [
      'stream-type',
      'player-size',
      'default-hidden-captions',
      'has-captions',
      'forward-seek-offset',
      'backward-seek-offset',
    ];
  }

  constructor() {
    super();

    this.attachShadow({ mode: 'open' });
    render(template(getProps(this)), this.shadowRoot as Node);
  }

  attributeChangedCallback() {
    render(template(getProps(this)), this.shadowRoot as Node);
  }
}

if (!customElements.get('media-theme-mux')) {
  customElements.define('media-theme-mux', MediaThemeMux);
}
```

Note that the theme still has a dependency on "jtml" (html.ts) I think that's fine for this theme but we should probably make that html.ts file an isolated package so `<media-theme-mux>` can be a separate package as well


packages/mux-player/src/index.ts
```diff
@@ -209,6 +220,15 @@ class MuxPlayerElement extends VideoApiElement {
...
 
   #monitorLiveWindow() {
+    this.mediaController?.addEventListener('mediaplayrequest', (event) => {
+      if (
+        (event.target as Element)?.localName === 'media-play-button' &&
+        (this.streamType === StreamTypes.LIVE || this.streamType === StreamTypes.LL_LIVE)
+      ) {
+        seekToLive(this);
+      }
+    });
```
luwes, line 230: this would still have a bug where once we switch themes it would require adding the mediaController listeners again. 

this pattern also separates the view and event listener, not sure that is the best way but it was the easiest to get this draft working because the play button is used in several places.

packages/mux-player/src/template.ts
```diff
@@ -71,9 +71,21 @@ export const content = (props: MuxTemplateProps) => html`
...
+        props.onSeekToLive?.(evt);
+        if (props.paused) {
+          const playRequestEvt = new CustomEvent('mediaplayrequest', { composed: true, bubbles: true });
+          (this as HTMLButtonElement).dispatchEvent(playRequestEvt);
+        }
+      }}"
+      class="mxp-seek-to-live-button"
+    >
+      Live
+    </button>
```
luwes, line 87: this seek to live button I chose to add to the theme via a slot and not have it in the theme itself because of the extra required props like onSeekToLive and inLiveWindow. we could also solve this in the same way as the play button by attaching the click event listener in index.ts but it does separate things more and more. 
mixing imperative with declarative.

packages/mux-player/src/index.ts
cjpillsbury: These are all good things to have on our radar. Thanks for the callout.

packages/mux-player/src/styles.css
```diff
@@ -17,110 +17,13 @@ a {
...
-
-media-time-display {
-  white-space: nowrap;
-}
-
-:is(media-time-display, media-text-display, media-playback-rate-button) {
-  color: inherit;
-}
-
 .mxp-seek-to-live-button {
```
cjpillsbury, line 25: Might make sense to move this to the theme's CSS if this is a well-defined slot with a particular purpose. Also, unrelated, looks like the `color` here got missed from Dylan's updates re:`primaryColor` et al.

examples/vanilla-ts-esm/public/mux-player.html
```diff
@@ -67,6 +67,7 @@
     <div class="buttons">
       <button class="dialog-show-btn">Show error dialog</button>
       <button class="dialog-close-btn">Close dialog</button>
+      <button class="switch-theme-btn">Switch Theme</button>
```
cjpillsbury, line 70: blocking: since this current codebase is likely an "intermediary step" with how themes will work, I recommend not merging this example until we're confident in what we want our public-facing API to look like.

packages/mux-player/src/index.ts
cjpillsbury: non-blocking: Per previous conversation, we'll likely want to revisit how this works.

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -0,0 +1,269 @@
...
+    playerSize: el.getAttribute('player-size'),
+    defaultHiddenCaptions: el.hasAttribute('default-hidden-captions'),
+    hasCaptions: el.hasAttribute('has-captions'),
+    forwardSeekOffset: el.getAttribute('forward-seek-offset'),
+    backwardSeekOffset: el.getAttribute('backward-seek-offset'),
+    ...state,
+  };
+}
+
+class MediaThemeMux extends HTMLElement {
```
cjpillsbury, line 241: non-blocking: Per our conversation, we will should refactor this to extend a "`MediaTheme`" that we expect to minimally have a `render()` method that will be invoked on initialization to anticipate integration with Media Chrome.

packages/mux-player/src/styles.css
cjpillsbury: non-blocking: bumping ☝️  as a followup

packages/mux-player/test/template.test.js
```diff
@@ -11,7 +11,7 @@ describe('<mux-player> template render', () => {
     render(content({}), div);
```
cjpillsbury, line 11: non-blocking: If we end up settling on these slots, we should add more tests around how they behave under varying conditions

packages/mux-player/test/template.test.js
```diff
@@ -11,7 +11,7 @@ describe('<mux-player> template render', () => {
...
     assert.equal(
```
cjpillsbury, line 12: non-blocking: If we end up settling on these attributes passed into themes, we should add more tests around how they behave when various values are set/omitted, including over time.

cjpillsbury (changes requested):
Only one minor blocker. Otherwise, 🚢 🇮🇹

examples/vanilla-ts-esm/public/mux-player.html
luwes: good call! I removed this until we have this functioning properly

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
luwes: Media Chrome PR here https://github.com/muxinc/media-chrome/pull/209

cjpillsbury (approved):

PR #177, opened 2022-04-11
fix(mux-player): metadata fields should be set based on 'external'/at…

cjpillsbury:
…tr values in mux-player.

Still needs tests.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -103,6 +103,11 @@ function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
     defaultHiddenCaptions: el.defaultHiddenCaptions,
     playerSize: getPlayerSize(el),
     hasCaptions: !!getCcSubTracks(el).length,
+    // NOTE: In order to guarantee all expected metadata props are set "from the outside" when used
+    // and to guarantee they'll all be set *before* the playback id is set, using attr values here (CJP)
+    metadataVideoId: el.getAttribute(MuxVideoAttributes.METADATA_VIDEO_ID),
+    metadataVideoTitle: el.getAttribute(MuxVideoAttributes.METADATA_VIDEO_TITLE),
+    metadataViewerUserId: el.getAttribute(MuxVideoAttributes.METADATA_VIEWER_USER_ID),
     ...state,
   };
 }
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -43,6 +43,14 @@ export const content = (props: MuxTemplateProps) => html`
       debug="${props.debug ?? false}"
       prefer-mse="${props.preferMse ?? false}"
       start-time="${props.startTime != null ? props.startTime : false}"
+      metadata-video-id="${props.metadataVideoId ?? props.metadata?.video_id ?? false}"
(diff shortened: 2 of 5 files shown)


packages/mux-player/src/template.ts
```diff
@@ -43,6 +43,14 @@ export const content = (props: MuxTemplateProps) => html`
       debug="${props.debug ?? false}"
       prefer-mse="${props.preferMse ?? false}"
       start-time="${props.startTime != null ? props.startTime : false}"
+      metadata-video-id="${props.metadataVideoId ?? props.metadata?.video_id ?? false}"
+      metadata-video-title="${props.metadataVideoTitle ?? props.metadata?.video_title ?? false}"
+      metadata-viewer-user-id="${props.metadataViewerUserId ?? props.metadata?.viewer_user_id ?? false}"
```
luwes, line 48: this looks good but even better I think is using the same pattern we do with `getVideoAttribute` in the `metadata` getter and there wouldn't be a need for extra props in `getProps`. only use the `.metadata` property

packages/mux-player/src/index.ts
```diff
@@ -103,6 +103,11 @@ function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
     defaultHiddenCaptions: el.defaultHiddenCaptions,
     playerSize: getPlayerSize(el),
     hasCaptions: !!getCcSubTracks(el).length,
+    // NOTE: In order to guarantee all expected metadata props are set "from the outside" when used
+    // and to guarantee they'll all be set *before* the playback id is set, using attr values here (CJP)
+    metadataVideoId: el.getAttribute(MuxVideoAttributes.METADATA_VIDEO_ID),
+    metadataVideoTitle: el.getAttribute(MuxVideoAttributes.METADATA_VIDEO_TITLE),
+    metadataViewerUserId: el.getAttribute(MuxVideoAttributes.METADATA_VIEWER_USER_ID),
```
luwes, line 110: this could be removed and in the metadata getter of mux-player something like 

```js
    const video_id = this.getAttribute(MuxVideoAttributes.METADATA_VIDEO_ID);
    const video_title = this.getAttribute(MuxVideoAttributes.METADATA_VIDEO_TITLE);
    const viewer_user_id = this.getAttribute(MuxVideoAttributes.METADATA_VIEWER_USER_ID);

return this.video ? this.video.metadata : {
  video_id,
video_title,
viewer_user_id,
}

Comment (cjpillsbury) [reply packages/mux-player/src/index.ts:110]: Unfortunately that wouldn't work for valid use cases. Can discuss, but tl;dr - need to guarantee that the attributes for all 3 are respected when set "from the outside", even after initial video render.

File: packages/mux-player/src/template.ts (hunk 1 above)
Comment (cjpillsbury) [reply packages/mux-player/src/template.ts:48]: see above comment

File: packages/mux-player/src/index.ts (hunk 2 above)
Comment (luwes) [reply packages/mux-player/src/index.ts:110]: gotcha, yes that's a tricky one. great catch. 

seems because I tried to use the metadata object in template.ts because this line does pass the props `metadataVideoId` etc to the template but nothing is done with them 

```js
this.#render({ [toPropName(attrName)]: newValue });
```

luwes (approved):
LGTM

PR #179, opened 2022-04-12
fix: prevent non-fatal error events propagating

luwes:
this change prevent any non-fatal error events from leaving mux-player.


cjpillsbury (approved):
🏄‍♂️

PR #183, opened 2022-04-13
fix: make mux-video errors more uniform, fix async

luwes:
also fixes an issue with random failing tests

### DIFF
--- a/packages/mux-player/test/errors.test.js
+++ b/packages/mux-player/test/errors.test.js
@@ -45,22 +45,22 @@ describe('errors', () => {
     assert(fired === true, 'the error handler was fired');
   });
 
-  it('default message for MediaError.MEDIA_ERR_ABORTED', async function () {
+  it('default message for MediaError.MEDIA_ERR_ABORTED', function () {
     const error = new MediaError(undefined, MediaError.MEDIA_ERR_ABORTED);
-    const { dialog } = await getErrorLogs(error);
+    const { dialog } = getErrorLogs(error);
     assert.equal(dialog.message, 'You aborted the media playback');
   });
 
-  it('default message for MediaError.MEDIA_ERR_NETWORK', async function () {
+  it('default message for MediaError.MEDIA_ERR_NETWORK', function () {
     const error = new MediaError(undefined, MediaError.MEDIA_ERR_NETWORK);
-    const { dialog } = await getErrorLogs(error);
+    const { dialog } = getErrorLogs(error);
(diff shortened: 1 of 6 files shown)


packages/playback-core/src/index.ts
```diff
@@ -352,6 +351,31 @@ export const loadMedia = (
...
+  // Stop immediate propagation of the native error event, re-dispatch below!
+  event.stopImmediatePropagation();
+
+  const mediaEl = event.target as HTMLMediaElement;
+  const { message, code } = mediaEl?.error ?? {};
+  const error = new MediaError(message, code);
+
+  if (mediaEl.src) {
+    // Attempt to get the response code from the M3U8 src url.
+    const { status } = await fetch(mediaEl.src as RequestInfo);
```
cjpillsbury, line 367: blocking: I think this condition may be too broad? At least in playback-core/mux-video/mux-audio, this may not be an m3u8 and may not be a mux-hosted asset. Also, I think we should be able to check against [the error code](https://developer.mozilla.org/en-US/docs/Web/API/MediaError/code#media_error_code_constants) and only `fetch` for 1 & 2?

packages/mux-player/test/errors.test.js
```diff
@@ -45,70 +45,70 @@ describe('errors', () => {
...
     const error = new MediaError(undefined, MediaError.MEDIA_ERR_DECODE);
     const offline = true;
-    const { dialog } = await getErrorLogs(error, offline);
+    const { dialog } = getErrorLogs(error, offline);
     assert.equal(dialog.title, 'Your device appears to be offline');
     assert.equal(dialog.message, 'Check your internet connection and try reloading this video.');
   });
 
-  it('412 response code message for MediaError.MEDIA_ERR_NETWORK', async function () {
+  it('412 response code message for MediaError.MEDIA_ERR_NETWORK', function () {
```
cjpillsbury, line 91: non-blocking: either in this PR or as a followup, we should write some more full integration tests that will actually request assets and internally throw these errors.

cjpillsbury (changes requested):
one blocking callout and another non-blocking (though we should minimally smoke test). Otherwise, 😎

packages/mux-player/test/errors.test.js
luwes: yup good call! I've locally tested some of these assets

cjpillsbury (approved):
🏄‍♂️

PR #185, opened 2022-04-14
fix: make mux-player size based on video element

luwes:
~Requires this PR in Media Chrome: https://github.com/muxinc/media-chrome/pull/205~

Upgraded media-chrome to 0.6.0

### DIFF
--- a/packages/mux-player/src/styles.css
+++ b/packages/mux-player/src/styles.css
@@ -1,8 +1,5 @@
 :host {
-  /* default aspect ratio, overwrite w/ e.g. style="aspect-ratio: 16 / 10" */
-  aspect-ratio: 16 / 9;
-  display: block;
-  width: 100%;
+  display: inline-block;
 }
 
 /* Hide custom elements that are not defined yet */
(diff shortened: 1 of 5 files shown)


packages/mux-player/src/styles.css
```diff
@@ -21,9 +18,6 @@ media-controller {
   --media-control-hover-background: transparent;
   --media-range-track-background-color: rgba(255, 255, 255, 0.5);
   --media-range-track-border-radius: 3px;
-  --media-aspect-ratio: auto;
-  display: block;
-  width: 100%;
```
gkatsev, line 26: width is still needed as well

gkatsev, line 26: width is still needed as well

gkatsev (changes requested):

gkatsev (approved):

packages/mux-player/src/styles.css
luwes: added, I didn't reproduce anything that required it yet or broke without it but it does make sense to have `width: 100%` as well

gkatsev: I was seeing some issues particularly with portrait videos and trying to set the width, which this resolved.

PR #186, opened 2022-04-15
build: Use downlevel-dts to downgrade type definitions in decls. Upda…

cjpillsbury:
…te package.json. Update unconvertable types in src.

NOTES:
* There is a bug in the version of Terser used by nextjs version 9.5.5 (and a few other versions in between). Can test in a nextjs test app by targeting either `next@9.5.2` or `next@10.1.0`. For more details, see [https://github.com/vercel/next.js/issues/18056](https://github.com/vercel/next.js/issues/18056)
* There is currently a bug with using wildcarding paths for typeVersions with the default export in packages. For more details, see [https://github.com/microsoft/TypeScript/issues/43133](https://github.com/microsoft/TypeScript/issues/43133)
* Since externalizing our dependencies, to locally link, I've had to manually add each @mux-elements dependency via the file protocol and, after installing, manually delete the `node_modules` inside each @mux-elements package. Still need to explore some options that Gary identified that may make this process less onerous/manual.


luwes (approved):
LGTM

PR #191, opened 2022-04-20
fix: point pkgjson#browser at mjs build for webpack 4

gkatsev:


gkatsev:
This PR should be squash merged.

luwes (approved):
LGTM, type is not needed anymore?

gkatsev:
@luwes yeah, we don't really need type, especially since we don't have regular js files that have requires or imports and our main users are still bundlers.

PR #192, opened 2022-04-21
feat: add errorTranslator prop to mux-video 

luwes:
This change adds a new errorTranslator property for customizing error messages sent to Mux data.
The property can be passed to MuxVideoReact too, the logic is all done in playback-core.

It's a shame but I couldn't figure out a way to make it work with the `errorTranslator` mux-embed provides because mux-player requires the actual MediaError object to generate the custom error messages. It requires the error response code of the src url.

mux-embed errorTranslator only provides the error code and the error message.

### DIFF
--- a/packages/mux-player/src/errors.ts
+++ b/packages/mux-player/src/errors.ts
@@ -8,35 +8,48 @@ export function getErrorLogs(
   error: MediaError,
   offline?: boolean,
   playbackId?: string,
-  playbackToken?: string
+  playbackToken?: string,
+  translate?: boolean
 ): { dialog: DialogOptions; devlog: DevlogOptions } {
   let dialog: DialogOptions = {};
   let devlog: DevlogOptions = {};
 
   switch (error.code) {
     case MediaError.MEDIA_ERR_NETWORK: {
-      dialog.title = i18n`Network Error`;
+      dialog.title = i18n(`Network Error`, translate);
       dialog.message = error.message;
 
       switch (error.data?.response.code) {
         case 412: {
-          dialog.title = i18n`Video is not currently available`;
-          dialog.message = i18n`The live stream or video file are not yet ready.`;
-          devlog.message = i18n`This playback-id may belong to a live stream that is not currently active or an asset that is not ready.`;
+          dialog.title = i18n(`Video is not currently available`, translate);
+          dialog.message = i18n(`The live stream or video file are not yet ready.`, translate);
+          devlog.message = i18n(
+            `This playback-id may belong to a live stream that is not currently active or an asset that is not ready.`,
+            translate
+          );
           devlog.file = '412-not-playable.md';
           break;
         }
         case 404: {
-          dialog.title = i18n`Video does not exist`;
+          dialog.title = i18n(`Video does not exist`, translate);
           dialog.message = '';
-          devlog.message = i18n`This playback-id does not exist. You may have used an Asset ID or an ID from a different resource.`;
+          devlog.message = i18n(
+            `This playback-id does not exist. You may have used an Asset ID or an ID from a different resource.`,
+            translate
+          );
           devlog.file = '404-not-found.md';
           break;
         }
         case 403: {
-          dialog.title = i18n`Invalid playback URL`;
-          dialog.message = i18n`The video URL or playback-token are formatted with incorrect or incomplete information.`;
-          devlog.message = i18n`403 error trying to access this playback URL. If this is a signed URL, you might need to provide a playback-token.`;
+          dialog.title = i18n(`Invalid playback URL`, translate);
+          dialog.message = i18n(
+            `The video URL or playback-token are formatted with incorrect or incomplete information.`,
+            translate
+          );
+          devlog.message = i18n(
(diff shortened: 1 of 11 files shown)


packages/mux-player/src/errors.ts
```diff
@@ -108,7 +129,7 @@ export function getErrorLogs(
     case MediaError.MEDIA_ERR_SRC_NOT_SUPPORTED: {
       // If native HLS is used on Safari, M3U8 response errors cause media src not supported errors.
       // If the response returns an error code, fix the MediaError.code and get detailed error logs.
-      const status = error.data?.response?.status;
+      const status = error.data?.response?.code;
```
luwes, line 132: this is a bug actually

packages/playback-core/src/errors.ts
```diff
@@ -1,10 +1,11 @@
 export class MediaError extends Error {
-  static MEDIA_ERR_CUSTOM: number = 0;
   static MEDIA_ERR_ABORTED: number = 1;
   static MEDIA_ERR_NETWORK: number = 2;
   static MEDIA_ERR_DECODE: number = 3;
   static MEDIA_ERR_SRC_NOT_SUPPORTED: number = 4;
   static MEDIA_ERR_ENCRYPTED: number = 5;
+  // @see https://docs.mux.com/guides/data/monitor-html5-video-element#customize-error-tracking-behavior
+  static MEDIA_ERR_CUSTOM: number = 100;
```
cjpillsbury, line 8: nice.

packages/playback-core/src/errors.ts
```diff
@@ -19,11 +20,11 @@ export class MediaError extends Error {
   fatal: boolean;
   data?: any;
 
-  constructor(message?: string, code: number = 0, fatal?: boolean) {
+  constructor(message?: string, code: number = MediaError.MEDIA_ERR_CUSTOM, fatal?: boolean) {
     super(message);
     this.name = 'MediaError';
     this.code = code;
-    this.fatal = fatal ?? code >= MediaError.MEDIA_ERR_NETWORK;
+    this.fatal = fatal ?? (code >= MediaError.MEDIA_ERR_NETWORK && code <= MediaError.MEDIA_ERR_ENCRYPTED);
```
cjpillsbury, line 27: 👌

dylanjha:
Okay, so here's where I landed on this:

1. In general for our level of integration with `mux-embed` it makes sense to flip `automaticErrorTracking` to `false` -- because we want full control over the errors that get sent to Mux Data and we do so with `mux.emit('error')`. This is a perfectly good recommended way of doing things 👌🏼 
2. There is a bug in `mux-embed` when using `automaticErrorTracking: false` with Hls.js then that settings is not respected and errors are automatically tracked (boo -- let's fix this). A workaround is to use mux-embed's `errorTranslator` function and return `false`. This is a hack and purely a workaround. The workaround is a little gnarly because even when we `emit('error')` ourselves, the `errorTranslator` function we provided gets invoked so we have to be sneaky about this and make a decision inside `errorTranslator` to say: is this the `mux-embed` bug, or is this one of the errors we dispatched? -- we do this by checking `typoeof error.player_error_code === 'string'`. This is a MAJOR HACK, but it's the only way we can until the bug gets patched.
3. There is an enhancement feature request with mux-embed so that we could make `automaticErrorTracking: true`, we could rely fully on `errorTranslator` and that would be nice, but that's more of a separate conversation. The way `errorTranslator` functions now, it doesn't give us the full context of the Hls.js error, so that's not going to be possible

Given those constraints, I think we have it right here. We should:

1. Ship this
1. Get the bug fixed in `mux-embed`
1. After that bug is fixed, we no longer have to pass along `errorTranslator` into `mux-embed` at all

dylanjha:
Okay, so here's where I landed on this:

1. In general for our level of integration with `mux-embed` it makes sense to flip `automaticErrorTracking` to `false` -- because we want full control over the errors that get sent to Mux Data and we do so with `mux.emit('error')`. This is a perfectly good recommended way of doing things 👌🏼 
2. There is a bug in `mux-embed` when using `automaticErrorTracking: false` with Hls.js then that settings is not respected and errors are automatically tracked (boo -- let's fix this). A workaround is to use mux-embed's `errorTranslator` function and return `false`. This is a hack and purely a workaround. The workaround is a little gnarly because even when we `emit('error')` ourselves, the `errorTranslator` function we provided gets invoked so we have to be sneaky about this and make a decision inside `errorTranslator` to say: is this the `mux-embed` bug, or is this one of the errors we dispatched? -- we do this by checking `typoeof error.player_error_code === 'string'`. This is a MAJOR HACK, but it's the only way we can until the bug gets patched.
3. There is an enhancement feature request with mux-embed so that we could make `automaticErrorTracking: true`, we could rely fully on `errorTranslator` and that would be nice, but that's more of a separate conversation. The way `errorTranslator` functions now, it doesn't give us the full context of the Hls.js error, so that's not going to be possible

Given those constraints, I think we have it right here. We should:

1. Ship this
1. Get the bug fixed in `mux-embed`
1. After that bug is fixed, we no longer have to pass along `errorTranslator` into `mux-embed` at all

PR #193, opened 2022-04-21
ci: add release notes script and GH action

luwes:
This first pass only adds the separate GH action that can be manually kicked off from Github actions UI.
If it works good we can remove the Lerna option `--create-release github` and save the step of having to get the GH_TOKEN locally

### DIFF
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -1,0 +1,26 @@
+name: Create Release Notes
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  changelog-to-release-notes:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      # extract `engines.node` from package.json and save it to output
+      - name: Get Node.JS version from package.json
+        id: get-versions
+        run: echo ::set-output name=node::$(jq -r .engines.node ./package.json)
+      - name: Use Node.js ${{ steps.get-versions.outputs.node }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ steps.get-versions.outputs.node }}
+          cache: yarn
+      - run: yarn install --frozen-lockfile
+      - name: Create Release
+        run: yarn create-release-notes
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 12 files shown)


scripts/create-release-notes/create-release-notes.js
```diff
@@ -0,0 +1,67 @@
...
+    console.error(err);
+  }
+}
+
+const changelog = await extractLog(process.argv[2]);
+const tag = await getTagName();
+const octokit = new Octokit({ auth: process.env.GH_TOKEN });
+
+try {
+  await octokit.request('POST /repos/{owner}/{repo}/releases', {
```
gkatsev, line 55: It would be nice to set a `discussion_category_name`, so that it starts a discussion for the new version. But that can probably wait till we get to public beta.

package.json
```diff
@@ -36,7 +36,8 @@
     "build": "lerna run build",
     "prepare": "husky install",
     "deploy": "lerna publish from-package --no-private --no-verify-access --conventional-commits --conventional-prerelease=@mux-elements/mux-player,@mux-elements/mux-player-react",
-    "version:update": "lerna version --exact --no-private --conventional-commits --conventional-prerelease=@mux-elements/mux-player,@mux-elements/mux-player-react --create-release github"
+    "version:update": "lerna version --exact --no-private --conventional-commits --conventional-prerelease=@mux-elements/mux-player,@mux-elements/mux-player-react --create-release github",
```
gkatsev, line 39: I think we don't want `--create-release github` anymore?

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,26 @@
+name: Create Release Notes
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
gkatsev, line 5: not a blocker for this PR, and potentially should just be done in a follow-up PR:
would be nice to set this up so that it runs automatically as part of [the Publish commit](https://github.com/muxinc/elements/commit/d6078b8dac3edb45a4f2e625a3936446caa02595).

package.json
luwes: yup, I thought first to try the release-notes GH action out separately. if it works good, I'll remove in next PR

.github/workflows/release-notes.yml
luwes: indeed, we can add it to cd.yml after 1 manual test

package.json
gkatsev: sounds good to me

gkatsev (approved):

PR #195, opened 2022-04-23
feat: add errorTranslator prop & muxVideo.error property

luwes:
This change makes errors more uniform in mux-video by adding a `muxVideo.error` property and now the errorTranslator doesn't require a special `player_error` property because it can access that on the `mux-video` element.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -20,7 +20,7 @@ import { render } from './html';
 import { getErrorLogs } from './errors';
 import { toNumberOrUndefined, i18n, parseJwt, containsComposedNode } from './utils';
 import * as logger from './logger';
-import type { MuxTemplateProps } from './types';
+import type { MuxTemplateProps, ErrorEvent } from './types';
 
 export { MediaError };
 export type Tokens = {
@@ -294,6 +294,25 @@ class MuxPlayerElement extends VideoApiElement {
     // Keep this event listener on mux-player instead of calling onError directly
     // from video.onerror. This allows us to simulate errors from the outside.
     this.addEventListener('error', onError);
+
+    if (this.media) {
+      this.media.errorTranslator = (errorEvent: ErrorEvent = {}) => {
+        const { player_error, player_error_message } = errorEvent;
+        if (!player_error) return errorEvent;
+
+        const { devlog } = getErrorLogs(
+          player_error,
+          !window.navigator.onLine,
+          this.playbackId,
+          this.playbackToken,
+          false
+        );
+        return {
+          player_error_code: player_error.code,
+          player_error_message: devlog.message ? String(devlog.message) : player_error_message,
+        };
+      };
+    }
 
     this.media?.addEventListener('error', (event: Event) => {
       let { detail: error }: { detail: any } = event as CustomEvent;
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -297,19 +297,19 @@ class MuxPlayerElement extends VideoApiElement {
 
     if (this.media) {
       this.media.errorTranslator = (errorEvent: ErrorEvent = {}) => {
-        const { player_error, player_error_message } = errorEvent;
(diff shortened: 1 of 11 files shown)


packages/mux-player/src/index.ts
```diff
@@ -271,22 +271,32 @@ class MuxPlayerElement extends VideoApiElement {
...
+        if (!this.video?.error) return errorEvent;
 
-      // If it is a hls.js error event there will be an error object in the event.
-      // If it is a native video error event there will be no error object.
-      if (!error) {
-        const { message, code } = this.video?.error ?? {};
-        error = new MediaError(message, code);
-      }
+        const { devlog } = getErrorLogs(
+          this.video?.error,
```
luwes, line 279: last error can be accessed here on `this.video.error` similar to 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/error

types/mux.d.ts
```diff
@@ -455,7 +455,7 @@ type EventParamsMap = {
   [events.SEEKED]: void;
   [events.REBUFFER_START]: void;
   [events.REBUFFER_END]: void;
-  [events.ERROR]: void;
+  [events.ERROR]: ErrorEvent;
```
cjpillsbury, line 458: good catch! this is also wrong in the original defs! 🎉

scripts/i18n-utils/i18n-utils.js
```diff
@@ -9,7 +9,8 @@ const srcContents = fs.readFileSync(srcFile).toString();
 const baseLangContents = {};
 
 const strings = [];
-const regex = /\b(?:i18n`(.*?)`)/g;
+// @see https://regexr.com/6jtib
+const regex = /\b(?:i18n\(\s*['"`](.*?)['"`]\s*[,)])/g;
```
cjpillsbury, line 13: 👍. Would be nice if we could avoid having to update this any time an api change happens, but we can let that be future us's problem.

packages/playback-core/src/index.ts
```diff
@@ -384,9 +416,32 @@ async function handleNativeError(event: Event) {
   );
 }
 
+/**
+ * Use a event listener instead of a function call when dispatching the Custom error
+ * event so consumers are still able to disable or intercept this error event.
+ * @param {Event} event
+ */
+function handleInternalError(event: Event) {
```
cjpillsbury, line 424: nit(non-blocking): We should be able to clean up the type being passed here.

packages/playback-core/src/index.ts
```diff
@@ -360,14 +388,18 @@ export const loadMedia = (
...
+  // via EventTarget.dispatchEvent() preventing an infinite loop.
+  if (!event.isTrusted) return;
 
   // Stop immediate propagation of the native error event, re-dispatch below!
   event.stopImmediatePropagation();
 
   const mediaEl = event.target as HTMLMediaElement;
-  const { message, code } = mediaEl?.error ?? {};
+  // Safari sometimes throws an error event with a null error.
+  if (!mediaEl?.error) return;
```
cjpillsbury, line 400: nit: `if (!mediaEl.error) return` should be sufficient.

packages/playback-core/src/errors.ts
```diff
@@ -1,10 +1,11 @@
 export class MediaError extends Error {
-  static MEDIA_ERR_CUSTOM: number = 0;
   static MEDIA_ERR_ABORTED: number = 1;
   static MEDIA_ERR_NETWORK: number = 2;
   static MEDIA_ERR_DECODE: number = 3;
   static MEDIA_ERR_SRC_NOT_SUPPORTED: number = 4;
   static MEDIA_ERR_ENCRYPTED: number = 5;
+  // @see https://docs.mux.com/guides/data/monitor-html5-video-element#customize-error-tracking-behavior
+  static MEDIA_ERR_CUSTOM: number = 100;
```
cjpillsbury, line 8: 😎

gkatsev (approved):

gkatsev (approved):

PR #198, opened 2022-04-25
Expose mux-video as `media` for consistency, prefix _hls

gkatsev:
With this PR, the recommended approach going forward for getting `mux-video` from `mux-player` is via `.media`. `.video` is still available for backwards compatibility.
The `.hls` property got renamed to `_hls` to discourage its use outside of elements.

### DIFF
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -79,6 +79,14 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
     return playerSoftwareVersion;
   }
 
+  /**
+   * @deprecated please use ._hls instead
+   */
+  get hls() {
+    console.warn('<mux-audio>.hls is deprecated, please use ._hls instead');
+    return this._hls;
+  }
+
   get _hls() {
     return this.__hls;
   }
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -79,7 +79,7 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
     return playerSoftwareVersion;
   }
 
-  get hls() {
+  get _hls() {
     return this.__hls;
   }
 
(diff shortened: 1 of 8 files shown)


packages/mux-player/src/video-api.ts
```diff
@@ -129,20 +129,27 @@ class VideoApiElement extends HTMLElement {
...
+  get media(): MuxVideoElement | null | undefined {
+    return this.shadowRoot?.querySelector('mux-video');
+  }
+
+  /**
+   * keep .video for backwards compatibility.
+   */
   get video(): MuxVideoElement | null | undefined {
     return this.shadowRoot?.querySelector('mux-video');
   }
```
gkatsev, line 155: Should we keep this? If so, should we log a warning encouraging folks to move over?

packages/mux-video/src/index.ts
```diff
@@ -94,7 +94,15 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
 
+  /**
+   * @deprecated please use ._hls instead
+   */
   get hls() {
+    console.warn('<mux-player>.hls is deprecated, please use ._hls instead');
+    return this._hls;
+  }
+
+  get _hls() {
```
cjpillsbury, line 105: nit(non-blocking): same proposal as `mux-audio` (something something unofficial something something)

packages/mux-audio/src/index.ts
```diff
@@ -79,7 +79,15 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
...
 
+  /**
+   * @deprecated please use ._hls instead
+   */
   get hls() {
+    console.warn('<mux-player>.hls is deprecated, please use ._hls instead');
+    return this._hls;
+  }
+
+  get _hls() {
```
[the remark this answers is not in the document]
gkatsev: Since we're using this property, it'll mean that the default example will log this warning as well. Ideally, there shouldn't be any such warnings by default.

cjpillsbury (approved):
🏄‍♂️

PR #199, opened 2022-04-25
Add parts to video and audio, run all builds on dev

### LINKED ISSUE #125: Bug: No way to style `<video>` element inside `<mux-video>`

- mux-video

- Chrome

Mac

`<script async src="https://unpkg.com/@mux-elements/mux-video"></script>`.

### Describe the bug

I want to style the`<video>` element inside of the `<mux-video>` element. In particular, I'd like to add `object-fit: cover`.

Pasted_Image_3_8_22__2_30_PM

## The current behavior

As far as I can tell, there is no way to style this element.

## The expected behavior

I want to be able to style it. A simple solution would be to put a `part="video"` attribute on the `<video>` element. That way one could style it as described in https://developer.mozilla.org/en-US/docs/Web/CSS/::part. I've tested this out by adding the `part` attribute in Chrome Devtools and it works.

image

gkatsev:
The part attributes for `mux-video` and `mux-audio` fix #125.


cjpillsbury (approved):
😎

packages/mux-video/src/CustomVideoElement.js
```diff
@@ -33,7 +33,7 @@ template.innerHTML = `
 
 </style>
 
-<video crossorigin></video>
+<video part="video" crossorigin></video>
```
luwes, line 36: awesome 👍

luwes (approved):
🚢

balexand:
Thanks!

PR #201, opened 2022-04-26
fix(playback-core): implement track/cue management in playback-core

luwes:
- This change fixes the issue of removing text tracks when switching playback-id's 
- Fixes a weird bug in Chrome that happens when toggling the textTrack disabled -> showing and captions are not rendered anymore. Re-adding the cues to the textTrack does the trick.

~Hls.js also had strange behavior after toggling to disabled the getters `hls.subtitleTrack` and `hls.subtiteTracks` were not correct anymore, they changed to `-1` and an empty array respectively. Added a fix for that by using the events Hls.js provides and keeping track of state in playback-core.~

^^ was caused by the change event handler not being cleaned up. this is fixed now.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -185,6 +185,7 @@ export const setupHls = (
       debug,
       startPosition,
       ...streamTypeConfig,
+      renderTextTracksNatively: false,
     });
 
     return hls;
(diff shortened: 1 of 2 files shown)


gkatsev (approved):
Seems to be working without any issues for me. I think that this is likely a better direction to go in compared to #197 and friends, as it's less magical.

PR #202, opened 2022-04-26
fix: make preload attribute work with hls.js

gkatsev:


cjpillsbury (approved):
😎

luwes (approved):
Sweet! 

potential improvement is adding logic in the preload setter, but can iterate on that

PR #203, opened 2022-04-27
fix: custom video events handling

### LINKED ISSUE #60: mux-video script tag doesn't consistently fire native video events
### Checklist

- [X] Verified whether your issue exists in the latest version Mux Element(s) release: `npm install @mux-elements/[element_name]` or `yarn @mux-elements/[element_name]`.
  - (NOTE: If the issue is not occurring in the latest version, we may not resolve it.)
- [X] Confirmed that this issue hasn't [already been filed](https://github.com/muxinc/elements/issues?q=is%3Aopen+is%3Aissue+label%3Abug).

- mux-video

- Chrome
- Safari
- Firefox

macOS

- script tag

### Describe the bug

Native video events e.g. `loadedmetadata` are inconsistently fired across different browsers.

## Steps To Reproduce

in a svelte project:
1. add `mux-video` script tag to the root html e.g. `app.html`
2. create a `/vid` route and the following snippet
3. serve your app and visit `~/vid` with your devtool open
4. you'll observe different behaviors on the following browsers:
   - chrome: `loadedmetadata` never fires. You can get the metadata on component mount though
      ```ts
      onMount(() => {
         if (muxVideoEl) {
             videoWidth = muxVideoEl.videoWidth
             videoHeight = muxVideoEl.videoHeight
             duration = muxVideoEl.duration
         }
      })
      ``` 
   - safari: `loadedmetadata` fires quite right but `videoWidth = 0` and `videoHeight = 0`. Attempting to read the values on component mount returns `0` for all 3 variables.
   - firefox: same as `chrome`.

```svelte
<script>
let src = 'https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8'
let duration
let videoWidth
let videoHeight
let muxVideoEl

const onloadedmetadata = () => {
  if (muxVideoEl) {
    duration = muxVideoEl.duration
    videoWidth = muxVideoEl.videoWidth
    videoHeight = muxVideoEl.videoHeight

    console.log('onloadedmetadata', { duration, videoWidth, videoHeight })
  }
}
</script>

<mux-video
  bind:this="{muxVideoEl}"
  src="{src}"
  controls
  preload="metadata"
  autoplay
  on:loadedmetadata="{onloadedmetadata}">
</mux-video>

<style>
mux-video {
  width: 100%;
  height: 100%;
  margin: 0 auto;
}
</style>
```

I should mention that the event fires right with the `npm package`.

I'm providing 2 links to code examples so that you can run it across many browsers as possible - for example safari is not compatible with stackblitz but works with svelte REPL. On svelte REPL, I provided 2 files `Bug.svelte` and `Workaround.svelte`.

Link to code example:
1. [Stackblitz](https://stackblitz.com/edit/sveltekit-1ir3xo?file=src%2Froutes%2F[vid].svelte)
2. [SvelteKit REPL](https://svelte.dev/repl/e7af803cb7ff491f87f76cbfbeac189b?version=3.46.1)

## The current behavior

Native video events don't fire as expected when using the script tag - focusing only on the `loadedmetadata` event for this issue. The workaround is waiting for the script to get `loaded` before rendering the `mux-video` tag. This means that one would need to load the script using javascript or any method that allows for access to the `loaded` event

## The expected behavior

All native video events should fire usually and as expected without needing to hook to `loaded` event.

luwes:
Related https://github.com/muxinc/custom-video-element/issues/7
Fix #60 

This fixes an issue where one couldn't add event listeners until the custom element was upgraded because only after the upgrade did we forward addEventListener calls to the native video element.

This solution works by not overriding the custom element addEventListener and removeEventListener so the events actually get attached to the element on which they are called. Then later the video events are dispatched on the custom element.

This fixes another hidden issue where before adding a video event listener to the document with capture: true would not work because the video element in the shadow DOM was the one that did the dispatching and those events don't leave the shadow DOM.


cjpillsbury (changes requested):
Looks great. Will need equivalent adds for `CustomAudioElement` and should probably add tests, esp. since these should be relatively easy to add.

packages/mux-audio/package.json
```diff
@@ -40,7 +40,7 @@
     "build:esm-module": "esbuild src/index.ts --target=es2019 --bundle --sourcemap --metafile=./dist/module.json --format=esm --outfile=./dist/mux-audio.mjs --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:cjs": "esbuild src/index.ts --target=es2019 --bundle --sourcemap --metafile=./dist/cjs.json --format=cjs --outdir=dist --out-extension:.js=.cjs --external:@mux-elements/* --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:iife": "esbuild src/index.ts --target=es2019 --bundle --sourcemap --metafile=./dist/iife.json --format=iife --outfile=./dist/mux-audio.js --define:PLAYER_VERSION=\"'$npm_package_version'\"",
-    "build:types": "tsc --declaration --emitDeclarationOnly --outDir './dist/types' && downlevel-dts ./dist/types ./dist/types-ts3.4",
+    "build:types": "tsc --declaration --emitDeclarationOnly --outDir './dist/types' && shx cp -r \"src/**/*.d.ts\" dist/types && downlevel-dts ./dist/types ./dist/types-ts3.4",
```
gkatsev, line 43: non-blocking: this made me realize that `dev:types` broke when `downlevel-dts` was added. We should fix it in a separate PR.

packages/mux-audio/test/player.test.js
```diff
@@ -19,4 +19,42 @@ describe('<mux-audio>', () => {
...
+        assert.equal(e.target, player);
+        eventMap[e.type] = true;
+      });
+    });
+
+    player.volume = 0.5;
+
+    // Seems only <audio> is throwing in my tests:
+    //
+    //   NotAllowedError: play() failed because the user didn't interact
```
gkatsev, line 42: I'm surprised this was an issue, even though it was muted. Maybe an assumption that `<audio>` element shouldn't be muted?

gkatsev (approved):

packages/mux-audio/test/player.test.js
luwes: yes, could be. I was surprised too

packages/mux-video/test/player.test.js
```diff
@@ -20,6 +20,45 @@ describe('<mux-video>', () => {
...
+      loadedmetadata: true,
+      loadstart: true,
+      play: true,
+      playing: true,
+      timeupdate: true,
+      volumechange: true,
+      waiting: true,
+      resize: true,
+    });
+  });
```
gkatsev, line 60: We can let the test runner know that we are done by returning a resolved promise like so
```suggestion
    });
    
    return Promise.resolve();
  });
```

luwes: @gkatsev an async function always returns Promise.resolve() with the value it returns when the function body ends. in this case it would be return Promise.resolve(undefined). no need to be explicit

luwes: oh didn't notice the test fail right away, it's unrelated to resolving the async function. random timeout issue

gkatsev: Ah, that's true.
Though, it seemed to help with the timeout errors in https://github.com/muxinc/elements/commit/c71767c873afce33884815ceae16be824a0cea84

gkatsev: I guess since we're trying to play, we should increase the timeout for this test at least to 5000, if not 12000 (I also noticed that I accidentally kept a really high timeout increase in https://github.com/muxinc/elements/commit/c71767c873afce33884815ceae16be824a0cea84#diff-358b4f843e79b6ab7642f6054352ab817b25c7b9a408090f5abe1b5475b494ecR420 and forgot to revert it back to 12000)

luwes: oh actually this is because `AbortError: The play() request was interrupted by a new load request.` 
hmm, a new load request

PR #205, opened 2022-05-03
fix: fix preload=metadata failing to load video

gkatsev:


gkatsev:
This still loads data. The issue is related to preload=metadata and autoplay and there's some timing shenanigans between them.

PR #206, opened 2022-05-03
fix: preload=metadata with autoplay

gkatsev:
With autoplay, it is possible to get the play event before
MANIFEST_LOADED happens. In those cases, we don't want to call
`hls.stopLoad` as we got a signal to continue playback.

### DIFF
--- a/packages/playback-core/src/autoplay.ts
+++ b/packages/playback-core/src/autoplay.ts
@@ -91,18 +91,34 @@ export const setupAutoplay = (
   // This will seek first play event of *any* live video including event-type,
   // which probably shouldn't seek
   if (!autoplay) {
+    const handleSeek = () => {
+      // don't seek if we're not live
+      if (!isLive) {
+        return;
+      }
+      // seek to either hls.js's liveSyncPosition or the native seekable end
+      if (hls?.liveSyncPosition) {
+        mediaEl.currentTime = hls.liveSyncPosition;
+      } else {
+        mediaEl.currentTime = mediaEl.seekable.end(0);
+      }
+    };
     mediaEl.addEventListener(
       'play',
       () => {
-        // don't seek if we're not live
-        if (!isLive) {
-          return;
-        }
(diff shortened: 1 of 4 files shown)


packages/playback-core/src/index.ts
```diff
@@ -373,8 +373,14 @@ export const loadMedia = (
...
-        // when preload is metadata, stop loading when the manifest is loaded
-        hls.once(Hls.Events.MANIFEST_LOADED, () => hls.stopLoad());
+        // when preload is metadata, stop loading when the manifest is parsed,
+        hls.once(Hls.Events.MANIFEST_LOADED, () => {
+          // unless we're no longer paused, which can happen with autoplay,
+          // where play will happen before MANIFEST_LOADED ends up triggering
+          if (mediaEl.paused) {
+            hls.stopLoad();
+          }
+        });
```
luwes, line 383: 👍

packages/playback-core/src/autoplay.ts
```diff
@@ -91,18 +91,34 @@ export const setupAutoplay = (
...
+        if (mediaEl.preload === 'metadata') {
+          if (hls) {
+            hls.once(Hls.Events.LEVEL_UPDATED, handleSeek);
+          } else if (mediaEl.seekable.end(0) === Infinity) {
+            const interval = setInterval(() => {
+              if (mediaEl.seekable.end(0) !== Infinity) {
+                clearInterval(interval);
+                handleSeek();
+              }
+            }, 100);
```
luwes, line 118: could we use an event here instead of a setInterval?

gkatsev: Let me see if there's something reliable here, as that would definitely be better.

gkatsev:
Follow-up:
- fix the test so that we don't skip it anymore
- look into an event that can be used instead of the set interval

cjpillsbury (approved):
👍

PR #208, opened 2022-05-03
fix: fullscreen controls layout

luwes:
Fixes an issue where the controls layout could be incorrect in fullscreen if the mux-player embed was small enough.

The controls mode was dependent on the mux-player dimensions when it should be dependent on the media-controller because this is the element that actually goes fullscreen.


cjpillsbury (approved):
makes perfect sense. 👌

PR #209, opened 2022-05-03
Preload autoplay followup

gkatsev:


packages/mux-player/src/video-api.ts
```diff
@@ -128,13 +128,7 @@ class VideoApiElement extends HTMLElement {
   }
 
   get hasPlayed() {
-    const mc = this.shadowRoot?.querySelector('media-controller');
```
gkatsev, line 131: with the theme change, this no longer works, so, we should grab it via `this.mediaController` instead.

packages/mux-player/test/player.test.js
```diff
@@ -476,23 +476,22 @@ describe('seek to live behaviors', function () {
...
     const playerEl = await fixture(`<mux-player
       playback-id="v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM"
       muted
       stream-type="ll-live"
       preload="auto"
     ></mux-player>`);
 
-    const mcPlayEl = playerEl.theme.shadowRoot.querySelector('media-play-button');
-    // NOTE: Need try catch due to bug in play+autoplay behavior (CJP)
-    try {
```
gkatsev, line 489: I'm not seeing the need for this try/catch, but will re-add if it fails in CI.

packages/mux-player/test/player.test.js
```diff
@@ -476,23 +476,22 @@ describe('seek to live behaviors', function () {
...
-      await playerEl.play();
-    } catch (_e) {}
+    await playerEl.play();
+
     await waitUntil(() => !playerEl.paused, 'play() failed');
     await waitUntil(() => playerEl.inLiveWindow, 'playback did not start inLiveWindow');
     playerEl.pause();
     await waitUntil(() => !playerEl.inLiveWindow, 'still inLiveWindow after long pause', { timeout: 7500 });
+
+    const mcPlayEl = playerEl.theme.shadowRoot.querySelector('media-play-button');
```
gkatsev, line 494: it seems that if we keep this line above, the mcPlayEl and the media-play-button that's actually in the player end up being two separate elements and so the next line doesn't do anything.

packages/playback-core/src/autoplay.ts
```diff
@@ -110,12 +110,15 @@ export const setupAutoplay = (
...
-                clearInterval(interval);
-                handleSeek();
-              }
-            }, 100);
+            mediaEl.addEventListener(
+              'canplay',
+              () => {
+                // we need a setTimeout because seeking immediately on canplay
+                // can cause the player to stop get into a non-playing state
+                setTimeout(handleSeek, 100);
```
gkatsev, line 118: 100ms seems to be the best and making the player not freeze after the seek, though, it isn't ideal either.

packages/mux-player/test/player.test.js
cjpillsbury: cool may have been resolved or at least side-stepped for the permutation with the preload+autoplay updates.

packages/mux-player/test/player.test.js
cjpillsbury: cool that makes sense. We may need to revisit the template parts architecture or how we're using it to make sure it's not unnecessarily creating entirely new DOM elements when it could just narrowly update props.

packages/playback-core/src/autoplay.ts
cjpillsbury: blocking: Is 100ms a "magic number"? Is deferring until the next event loop/frame a la `Promise.resolve().then(...)` or `setTimeout(...,0)` or `requestAnimationFrame(...)` not sufficient? If it is a necessary magic number, how confident are you that this will reliably work? Have you tested across envs?

cjpillsbury (changes requested):
May not need changes, but minimally, confirmation of setTimeout would be good.

packages/playback-core/src/autoplay.ts
gkatsev: Yeah, it's a bit of a magic number playing around in desktop safari. `setTimeout(0)` is *not* sufficient.
I'm not positive that this change is inherently better than what we had previously, unfortunately. But let me double check how it compares to the setInterval and how it works on ipad.

cjpillsbury (approved):
🎉

PR #212, opened 2022-05-05
fix: playback-core initialize() running 3 times

luwes:
on player load instead of 1.

I'm pretty sure attributeChangedCallback is called when a custom element is upgraded so the loop in CustomVideoElement is not needed. I'm a bit less sure about the manual call in connectedCallback. cc @cjpillsbury 

We might have to investigate a little deeper but this fixes the issue.


packages/mux-audio/src/CustomAudioElement.js
```diff
@@ -67,19 +67,15 @@ class CustomAudioElement extends HTMLElement {
...
-    });
+    const nativeEl = (this.nativeEl = this.shadowRoot.querySelector('audio'));
 
     // Neither Chrome or Firefox support setting the muted attribute
     // after using document.createElement.
     // One way to get around this would be to build the native tag as a string.
     // But just fixing it manually for now.
     // Apparently this may also be an issue with <input checked> for buttons
-    if (nativeEl.defaultMuted) {
+    if (this.hasAttribute('muted')) {
```
luwes, line 78: this is needed because we aren't running the above attributes loop which would set `nativeEl.defaultMuted` to the correct value. 

`attributeChangedCallback` is always called automatically when the custom element is upgraded and there are some attributes set so the loop is really not required.

PR #213, opened 2022-05-06
fix: fix 3x init of playback-core if `src` used

luwes:
Fixes an issue where playback-core was init 3 times on player load.

### DIFF
--- a/packages/mux-video/src/index.ts
+++ b/packages/mux-video/src/index.ts
@@ -359,7 +359,8 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
         if (newValue === oldValue) {
           break;
         }
-        this.__updateAutoplay?.(newValue);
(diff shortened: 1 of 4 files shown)


packages/mux-video/src/index.ts
```diff
@@ -359,7 +359,8 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
         if (newValue === oldValue) {
           break;
         }
-        this.__updateAutoplay?.(newValue);
+        /** In case newValue is an empty string or null, use this.autoplay which translates to booleans (WL) */
+        this.__updateAutoplay?.(this.autoplay);
```
luwes, line 363: FYI `__updateAutoplay()` didn't handle the case when newValue was an empty string

cjpillsbury (approved):

gkatsev (approved):
Confirmed that an event handler that previously got called 3 times only gets called once!

PR #218, opened 2022-05-10
fix(mux-player): Ensure that the externally set poster attr is used i…

cjpillsbury:
…ff it's set.


gkatsev (approved):
Verified fix locally.

PR #225, opened 2022-05-18
Upgrade to CRA 5, rename cjs to .cjs.js

gkatsev:
Supersedes #210, inspired by #224


luwes (approved):
🚢 !

cjpillsbury (changes requested):
blocker: has this been tested against CRA 4/older webpack? I noticed all cjs bundles were removed from the react packages, and I **_thought_** the assumption was cjs was expected for those cases.

gkatsev:
@cjpillsbury it's because esm builds were added to the react packages. We can remove those, if we wanted.

Good call on trying it out with current CRA, let me try it out locally.

gkatsev:
And I need to resolve the merge conflict again

gkatsev:
CRA4 with the two changes: https://github.com/muxinc/elements/pull/227
Has the following error:
```
Can't import the named export 'StreamTypes' from non EcmaScript module (only default export is available)
```

gkatsev:
I've removed the addition of the esm builds for the react packages because they seem to break CRA4.

gkatsev:
The CRA build of vercel is now passing in https://github.com/muxinc/elements/pull/227: https://elements-demo-create-react-app-git-fork-gkatsev-cjs-798162-mux.vercel.app/

cjpillsbury (approved):
non-blocking nit: looks like I/you generated the updated CRA app w/npm instead of yarn so the docs changed. We should probably still recommend folks use yarn, but this is def lo pri.

Otherwise, looking 😎, tested across different apps.

PR #227, opened 2022-05-19
don't merge

gkatsev:
This is a PR for the changes from https://github.com/muxinc/elements/pull/225 without the CRA update to verify that CRA4 still works as expected.


gkatsev:
esm build fails because webpack/CRA enforces strictness for es modules and doesn't allow importing from the react package as a default and named properties.

PR #228, opened 2022-05-19
feat(playback-core): Update config values for hls to be more consiste…

cjpillsbury:
…nt with native playback and to reasonable trim the backbuffer during playback.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -139,10 +139,19 @@ export const getType = (props: Partial<Pick<MuxMediaProps, 'type' | 'src'>>) => 
 };
 
 export const getStreamTypeConfig = (streamType?: ValueOf<StreamTypes>) => {
-  if (streamType === StreamTypes.LL_LIVE) {
-    return {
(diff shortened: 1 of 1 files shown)


packages/playback-core/src/index.ts
```diff
@@ -139,12 +139,27 @@ export const getType = (props: Partial<Pick<MuxMediaProps, 'type' | 'src'>>) =>
 };
 
 export const getStreamTypeConfig = (streamType?: ValueOf<StreamTypes>) => {
-  if (streamType === StreamTypes.LL_LIVE) {
-    return {
-      maxFragLookUpTolerance: 0.001,
+  const defaultConfig = {
```
gkatsev, line 142: seems a bit weird that the default config is part of `getStreamTypeConfig`. I'd expect the section below where we instantiate hlsjs to be more like:
```js
new Hls({
  ...defaultConfig,
  debug,
  startPosition,
  ...streamTypeConfig
});
```

cjpillsbury: cool I'm down.

gkatsev (approved):

PR #229, opened 2022-05-20
fix: tweak when and how much to shift captions when user is active

gkatsev:


packages/mux-player/src/index.ts
```diff
@@ -361,6 +361,17 @@ class MuxPlayerElement extends VideoApiElement {
 
     // toggles activeCues for a particular track depending on whether the user is active or not
     const toggleLines = (track: TextTrack, userInactive: boolean) => {
+      // if we're live, exit early from here because our controls design means captions don't need to shift
+      // unless secondary color is set, in which case the entire bottom control bar may be opaque
+      if (
+        this.streamType &&
+        ['live', 'll-live'].includes(this.streamType) &&
```
cjpillsbury, line 368: nit (blocking): Let's use the constants from playback-core here (have a followup to de-dup all the stuff we no longer need now that you gave us external modules)

packages/mux-player/src/index.ts
```diff
@@ -378,7 +389,12 @@ class MuxPlayerElement extends VideoApiElement {
           // for cues that have more than one line, we want to push the cue further up
           const lines = cue.text.split('\n').length;
           // start at -3 to account for thumbnails as well.
-          const setTo = -3 - lines;
+          let offset = -3;
+          if (this.streamType && ['live', 'll-live'].includes(this.streamType)) {
```
cjpillsbury, line 393: nit (blocking): same as above.

cjpillsbury (dismissed):
nits nits.

packages/mux-player/src/index.ts
```diff
@@ -361,6 +361,17 @@ class MuxPlayerElement extends VideoApiElement {
 
     // toggles activeCues for a particular track depending on whether the user is active or not
     const toggleLines = (track: TextTrack, userInactive: boolean) => {
+      // if we're live, exit early from here because our controls design means captions don't need to shift
+      // unless secondary color is set, in which case the entire bottom control bar may be opaque
+      if (
+        this.streamType &&
+        ['live', 'll-live'].includes(this.streamType) &&
+        !this.secondaryColor &&
+        this.offsetWidth >= 800
```
luwes, line 370: seems a very specific condition to me, is it worth to disable instead of keeping it uniform? 
I'm leaning to keep it uniform unless the experience is a lot better this way

gkatsev: it is a bit. If the player is smaller than ~800px there's a higher chance that the volume slider will overlap captions but when it's larger it becomes less likely.

PR #233, opened 2022-05-25
update mux-player signed url docs, add a quick check for it

### LINKED ISSUE #232: Docs/Bug: mux-player signed playback creates malformed URL
# Which documentation could use some work?

mux-player

# Is this a discrepancy or an improvement?

discrepancy

## 🖊 Describe what you'd like to see change in the documentation

The mux-player docs on [signed URLs](https://github.com/muxinc/elements/tree/main/packages/mux-player#signed-urls-and-other-playback-query-params) shows you can pass query params, like the token, with the playback ID. But according to the [attributes](https://github.com/muxinc/elements/tree/main/packages/mux-player#attributes) you must seperate the `playback-id` and `playback-token`.
Passing the token with the `playback-id` results in the malformed URL `https://stream.mux.com/playbackid?token=123?redundant_streams=true`
https://github.com/muxinc/elements/blob/3fdf72d0e1322571aa18e8de74649f0dce5ae7b9/packages/mux-player/src/helpers.ts#L29

## 💡 Describe how you think this change will help the documentation
The currently provided documentation on this is incorrect. You can't pass extra query parameters when using signed URLs according to the [mux docs](https://docs.mux.com/guides/video/play-your-videos#add-delivery-redundancy-with-redundant-streams)

Just have to make it clear that you are required to seperate them if using signed URLs (or don't append the `redundant_streams=true` if a token is passed with `playback-id`)

gkatsev:
Fixes #232.

### DIFF
--- a/packages/mux-player/README.md
+++ b/packages/mux-player/README.md
@@ -218,13 +218,3 @@ If you prefer to use the in-code MSE-based engine (currently hls.js) whenever po
 >
 </mux-player>
 ```
-
-### Signed URLs and other playback query params
-
-Mux supports a number of query parameters on HLS playback URLs. Most commonly is the `token=` param [used for signed URLs](https://docs.mux.com/guides/video/secure-video-playback).
-
-In order to use `token=` -- or any other query params, pass them through with the `playback-id` attribute:
-
-```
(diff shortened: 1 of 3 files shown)


dylanjha (approved):
LGTM!

PR #237, opened 2022-05-27
feat: add casting logic without cast-button

luwes:
For testing purposes there are some temporary cast buttons in the vanilla example that use the player API.
https://elements-demo-vanilla-git-fork-luwes-chromecast-mux.vercel.app/mux-player.html

- [x] ~handle loading the Google cast framework script~ decided this is up to the user to load the cast script 
- [x] upgrade Media Chrome to get the cast-button


packages/mux-player/src/template.ts
```diff
@@ -62,14 +63,25 @@ export const content = (props: MuxTemplateProps) => html`
...
     >
-      ${props.playbackId && props.streamType !== StreamTypes.LIVE && props.streamType !== StreamTypes.LL_LIVE
-        ? html`<track
-            label="thumbnails"
-            default
-            kind="metadata"
-            src="${getStoryboardURLFromPlaybackId(props.playbackId, props.tokens.storyboard)}"
-          />`
-        : html``}
+      <video
```
cjpillsbury, line 66: blocking: How would you feel about CustomVideoElement just baking in using `is="castable-video"` per the discussion w/Heff on the `<media-chrome/>` side? Feels lower risk in terms of refactors and makes ChromeCast support also easier for folks using `<mux-video/>` + `<media-chrome/>` (per e.g. our TMI presentations).

packages/mux-video/src/CustomVideoElement.js
```diff
@@ -69,7 +69,26 @@ class CustomVideoElement extends HTMLElement {
...
     this.shadowRoot.appendChild(template.content.cloneNode(true));
 
-    const nativeEl = (this.nativeEl = this.shadowRoot.querySelector('video'));
+    const media = this.shadowRoot.querySelector('[name=media]');
+    const nativeEl = (this.nativeEl = media.assignedElements({ flatten: true })[0]);
+
+    if (nativeEl.hasAttribute('is')) {
+      customElements.whenDefined(nativeEl.getAttribute('is')).then(() => {
+        // If nativeEl is a custom built-in forward new methods and props.
+        const props = Object.getOwnPropertyNames(Object.getPrototypeOf(nativeEl));
```
cjpillsbury, line 78: non-blocking/followup: this is probably going to be a bit brittle, since it assumes the relevant props are on:

1. the prototype of the element and not the instance (safer for native HTMLMediaElement and related)
2. the immediate prototype of the element and not somewhere in the protochain (a bigger concern if we're anticipating e.g. mixins or other inheritance chains)

packages/mux-video/src/CustomVideoElement.js
```diff
@@ -218,33 +239,38 @@ for (
...
-      },
-    };
-
-    if (prop !== prop.toUpperCase()) {
-      // Setter (not a CONSTANT)
-      config.set = function (val) {
-        this.nativeEl[prop] = val;
+forwardProps(nativeElTest, nativeElProps);
+
+function forwardProps(nativeEl, props) {
```
cjpillsbury, line 244: non-blocking: this will handle properties (with the concerns mentioned above). What about attributes? Should we e.g. check the `observedAttributes` of the element to ensure proper forwarding?

packages/mux-video/src/CustomVideoElement.js
```diff
@@ -218,33 +239,38 @@ for (
...
+        get() {
+          return this.nativeEl[prop];
+        },
       };
-    }
 
-    Object.defineProperty(CustomVideoElement.prototype, prop, config);
-  }
-});
+      if (prop !== prop.toUpperCase()) {
```
cjpillsbury, line 263: non-blocking: fine for now, but we should probably make this less presumptuous.

shared/polyfills/window.js
```diff
@@ -52,6 +52,14 @@ if (!globalThis.HTMLElement) {
   globalThis.HTMLElement = HTMLElement;
 }
 
+if (!globalThis.HTMLVideoElement) {
+  class HTMLVideoElement extends EventTarget {}
```
cjpillsbury, line 56: blocking: I think we're going to need to do some thorough testing to make sure we aren't breaking SSR polyfills here. If we are, this might mean we need to bump up prioritizing how we're handling polyfills on the Open Elements side of the fence.

cjpillsbury:
blocking: It looks like this release would expose "casting logic with controls" unless I'm missing something (which is certainly possible).

cjpillsbury (changes requested):
a handful of questions/callouts, with a smaller number being blockers.

packages/mux-video/src/CustomVideoElement.js
luwes: good call, will check this out.

shared/polyfills/window.js
gkatsev: we're already polyfilling related things like HTMLElement and EventTarget above, I'm not sure that this addition will cause any issues.
Currently, the polyfills are run before the rest of mux-video and mux-player.

packages/mux-player/src/template.ts
luwes: I'm fine with that but to me the current solution feels a bit easier than adding castable-video and castable-audio in each of 4 the Mux elements WC and React...

Maybe we could add Casting support later to those elements with this solution first

cjpillsbury: ah I just meant literally updating the template of CustomVideoElement from:

```html
<slot name="media"><video part="video" crossorigin></video></slot>
```

to:
```html
<slot name="media"><video part="video" is="castable-video" crossorigin></video></slot>
```
(and importing it into the module)

cjpillsbury: def cool with holding off as a followup if you think this is the path of lower risk

shared/polyfills/window.js
cjpillsbury: cool then marking as resolved for now. We've already run into some issues, but now that we're discussing those may have all been `document`-rlated.

cjpillsbury (approved):
before merging, make sure to remove the cast button (maybe comment out to make followups easier?)

packages/mux-video/src/CustomVideoElement.js
```diff
@@ -137,13 +162,19 @@ class CustomVideoElement extends HTMLElement {
...
 
-  // We need to handle sub-class custom attributes differently from
-  // attrs meant to be passed to the internal native el.
   attributeChangedCallback(attrName, oldValue, newValue) {
-    this.forwardAttribute(attrName, oldValue, newValue);
+    // Initialize the attributes right after construction when they become available.
+    if (!this.#hasAttrCallback && !this.isConnected) {
+      this.#hasAttrCallback = true;
+      this.#init();
+    }
```
luwes, line 170: this logic is here for a bug issued here https://github.com/muxinc/custom-video-element/issues/11

shared/castable-video/index.mjs
```diff
@@ -0,0 +1,3 @@
+// this is a ecmascript module entry for castable-video to remove the need to build for this package.
+// Since the castable-video will automatically bind to globalThis, no export is necessary.
+import './castable-video.js';
```
cjpillsbury, line 3: nit (non-blocking): fine for now, but we should probably export this longer term.

cjpillsbury (approved):
Good first pass!

PR #238, opened 2022-05-31
Feat: mux-uploader and mux-uploader-drop

clearlyTHUYDOAN:
## Description

Getting this up so I can get OSS eyes on the code. Based off [this TDD](https://www.notion.so/mux/TDD-Mux-Uploader-9273ce97b52749dc91910d3a1f84f7c7).

## To-Dos (Blocking Merge)
- [x] Initial README

## To-Dos (Fast Follows)
- [ ] Tests
  - Blocked by `upchunk` not playing well with the testing environment
- [ ] Accessibility (screen reader)


examples/vanilla-ts-esm/public/mux-uploader-full.html
```diff
@@ -0,0 +1,122 @@
...
+        <main>
+            <header>
+              <h1>Add a video.</h1>
+              <h1>Get a shareable link to stream it.</h1>
+            </header>
+            <h2>Enter your upload GCS url:</h1>
+            <input type="text" class="direct-upload-url" placeholder="https://storage.googleapis.com/..." />
+            <div class="button-container">
+              <mux-uploader type="bar" id="uploader-controller">
+                <mux-uploader-drop slot="dropzone" text="Upload to stream.new" fullScreen enableOverlay enableStatus></mux-uploader-drop>
```
luwes, line 106: (blocking): for attributes prefer to use lowercase;
- `fullscreen`
- `enable-overlay` or just `overlay` since it's a boolean attribute
- `enable-status` or `status`

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+      r="52"
+      cx="60"
+      cy="60"
+    />
+  <svg>
+</div>
+
+<slot name="dropzone">
+  <mux-uploader-drop></mux-uploader-drop>
+</slot>
```
luwes, line 203: 👍

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+
+class MuxUploaderElement extends HTMLElement {
+  hiddenFileInput: HTMLInputElement | null | undefined;
+  filePickerButton: HTMLButtonElement | null | undefined;
+  svgCircle: SVGCircleElement | null | undefined;
+  progressBar: HTMLElement | null | undefined;
+  uploadPercentage: HTMLElement | null | undefined;
+  statusMessage: HTMLElement | null | undefined;
+  retryMessage: HTMLElement | null | undefined;
+  _dropHandler: Function;
```
luwes, line 220: nit (non-blocking): we started using real private props in Mux player to keeping the internals as private as possible and have a unpolluted API. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

cool thing is that esbuild will automatically transform these to use WeakMaps for browsers that don't support them yet.

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+    this.retryMessage = this.shadowRoot?.getElementById('retry-message');
+
+    this._dropHandler = this.handleUpload.bind(this);
+  }
+
+  connectedCallback() {
+    this.setDefaultType();
+    this.setupFilePickerButton();
+    this.setupRetry();
+    this.setupDropHandler();
```
luwes, line 244: these could also possibly be real private methods

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+    this.removeEventListener('mux-drop', this._dropHandler, false);
+  }
+
+  get radius() {
+    return Number(this.svgCircle?.getAttribute('r'));
+  }
+
+  get circumference() {
+    return this.radius * 2 * Math.PI;
+  }
```
luwes, line 272: helper functions like these better are defined out of the class with the class instance as an argument so they're not exposed on the public API. 
e.g. `const getRadius = (el) => Number(el.svgCircle?.getAttribute('r'));`

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+    this.filePickerButton?.addEventListener('click', () => {
+      // TO-DO: Allows user to reattempt uploading the same file after an error.
+      // Note: Apparently Chrome and Firefox do not allow changing an indexed property on FileList...(TD).
+      // Source: https://stackoverflow.com/a/46689013
+
+      this.hiddenFileInput?.click();
+    });
+
+    this.hiddenFileInput?.addEventListener('change', (evt) => {
+      const file = this.hiddenFileInput?.files && this.hiddenFileInput?.files[0];
```
luwes, line 325:
```suggestion
      const file = this.hiddenFileInput?.files?.[0];
```

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+    this.setAttribute('role', 'progressbar');
+    this.setAttribute('aria-label', 'progress bar');
+    this.setAttribute('aria-live', 'polite');
+
+    // TO-DO: Might want to standardize if we prefer to have users disable or enable things. (TD).
+    // Edge case: User wants to use the uploader without drag.
+    // Because we slot a default mux-uploader-drop if they don't slot one,
+    // in order to disable drop, the user has to pass disableDrop to mux-uploader and mux-uploader-drop
+    // must apply it in order to disable the drop. It currently requires you to explicit pass "true".
+    if (this.getAttribute('disableDrop')) {
```
luwes, line 255:
```suggestion
    if (this.getAttribute('disable-drop')) {
```

packages/mux-uploader/src/mux-uploader-drop.ts
```diff
@@ -0,0 +1,168 @@
...
+    left: 0;
+    right: 0;
+    bottom: 0;
+    height: 100%;
+    width: 100%;
+  }
+
+  /* TO-DO: Make default dropzone hover more apparent. (TD).*/
+
+  :host([disableDrop]) .dropzone {
```
luwes, line 20:
```suggestion
  :host([disable-drop]) .dropzone {
```

examples/vanilla-ts-esm/public/mux-uploader-widget.html
```diff
@@ -0,0 +1,119 @@
...
+  <body>
+    <div class="container">
+      <header>
+        <h2>Enter your upload GCS url:</h1>
+        <input type="text" class="direct-upload-url" placeholder="https://storage.googleapis.com/..." />
+      </header>
+      <div class="dashboard">
+        <div class="widget"></div>
+        <div class="widget">
+          <mux-uploader type="bar" id="uploader-controller" disableDrop="true">
```
luwes, line 99:
```suggestion
          <mux-uploader type="bar" id="uploader-controller" disable-drop>
```

an HTML attribute without a value implies a boolean true

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+      const muxUploaderDrop = this.shadowRoot?.querySelector('mux-uploader-drop');
+      muxUploaderDrop?.setAttribute('disableDrop', '');
```
luwes, line 257:
```suggestion
      muxUploaderDrop?.setAttribute('disable-drop', '');
```

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+          const offset = this.circumference - (percent / 100) * this.circumference;
+
+          this.svgCircle.style.strokeDashoffset = offset.toString();
+        }
+      }
+    }
+  }
+
+  handleUpload(evt: CustomEvent) {
+    const url = this.getAttribute('url');
```
luwes, line 361: (blocking): missing the `url` setter and getter, this seems like an attribute that should also be settable via an element property

packages/mux-uploader/src/mux-uploader-drop.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      const uploaderController = this.muxUploader ?? this;
+
+      uploaderController.dispatchEvent(
+        new CustomEvent('mux-drop', {
+          composed: true,
+          bubbles: true,
+          detail: file,
+        })
+      );
```
luwes, line 161: I think this should be only `this.dispatchEvent()` because the event will bubble up the DOM tree.
it also better there's no need for the controller instance here

examples/vanilla-ts-esm/public/mux-uploader-widget.html
clearlyTHUYDOAN: yeah that's what I thought too so I was surprised when the logic wouldn't recognize the attribute value as truthy without explicit `true` 🤔  maybe I imagined it

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+  globalThis.customElements.define('mux-uploader', MuxUploaderElement);
+  /** @TODO consider externalizing this (breaks standard modularity) */
+  globalThis.MuxUploaderElement = MuxUploaderElement;
+}
+
+if (!globalThis.customElements.get('mux-uploader-drop')) {
+  globalThis.customElements.define('mux-uploader-drop', MuxUploaderDropElement);
+  //@ts-ignore
+  globalThis.MuxUploaderDropElement = MuxUploaderDropElement;
+}
```
gkatsev, line 417: should this piece be moved into mux-uploader-drop?

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+  position: relative;
+}
+
+.text-container {
+  flex-wrap: nowrap;
+  justify-content: space-between;
+  padding-bottom: 16px;
+}
+
+:host([draggable]) .upload-instruction{
```
gkatsev, line 79: nit:
```suggestion
:host([draggable]) .upload-instruction {
```

packages/mux-uploader/src/index.ts
gkatsev: This might depend on how we want to expose mux-uploader-drop directly to users of this package.

packages/mux-uploader/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@mux-elements/mux-uploader",
```
gkatsev, line 2: should we call this `@mux-elmeents/uploader`? One less word to write.

packages/mux-uploader/src/index.ts
cjpillsbury: Agree this should probably be defined in the mux-uploader-drop module.

examples/vanilla-ts-esm/public/mux-uploader-widget.html
cjpillsbury: for boolean attributes, you should be able to rely on `myFancyCustomElement.hasAttribute('my-special-boolean')`

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+      if (this.svgCircle) {
+        // strokeDasharray is the size of dashes used to draw the circle with the size of gaps in between.
+        // If the dash number is the same as the gap number, no gap is visible: a full circle.
+        // strokeDashoffset defines where along our circle the dashes (in our case, a dash as long as the
+        // circumference of our circle) begins. The larger the offset, the farther into the circle you're
+        // starting the "dash". In the beginning, offset is the same as the circumference. Meaning, the visible
+        // dash starts at the end so we don't see the full circle. Instead we see a gap the size of the circle.
+        // When the percentage is 100%, offset is 0 meaning the dash starts at the beginning so we can see the circle. (TD).
+
+        this.svgCircle.style.strokeDasharray = `${this.circumference} ${this.circumference}`;
```
cjpillsbury, line 291: followup(non-blocking): Following the code thread here, we may be able to do this entirely through css (🤞)

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+      }
+    }
+  }
+
+  handleUpload(evt: CustomEvent) {
+    const url = this.getAttribute('url');
+
+    if (!url) {
+      if (this.statusMessage) this.statusMessage.innerHTML = 'No url attribute specified -- cannot handleUpload';
+      throw Error('No url attribute specified -- cannot handleUpload');
```
cjpillsbury, line 365: blocking: Let's replace `throw Error` with `console.error` as an initial implementation. Throwing an error will create an unrecoverable state.

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,419 @@
...
+      file: evt.detail,
+    });
+
+    upload.on('error', (err) => {
+      this.setAttribute('upload-error', '');
+
+      if (this.statusMessage && this.uploadPercentage) {
+        this.statusMessage.innerHTML = 'An error has occurred';
+      }
+      throw Error(err.detail.message);
```
cjpillsbury, line 388: Like above, let's make this a `console.error` for now.

packages/mux-uploader/src/mux-uploader-drop.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    this.overlay = shadowRoot.getElementById('overlay');
+    this.overlayText = shadowRoot.getElementById('overlay-text');
+  }
+
+  connectedCallback() {
+    this.setupDragEvents();
+
+    //@ts-ignore
+    this.shadowRoot.getElementById('overlay-text').innerHTML = this.getAttribute('text');
```
cjpillsbury, line 93: nit (blocking): This should be in `attributeChangedCallback` instead of `connectedCallback` so it updates if/when the attribute changes.
nit (non-blocking): Can use `this.overlayText?.innerHTML` that you defined ☝️

cjpillsbury (changes requested):
A few small asks, otherwise 🚢 🇮🇹

packages/mux-uploader/src/index.ts
clearlyTHUYDOAN: TIL!

clearlyTHUYDOAN:
@luwes @cjpillsbury Addressed all blocking comments and most fast-follows.

packages/mux-uploader/src/mux-uploader-drop.ts
```diff
@@ -0,0 +1,168 @@
+const template = document.createElement('template');
+
+template.innerHTML = `
+<style>
+  /* These styles simulate a user passing these via props. Until
+     that's implemented, we just hardcode it for non-full-screen demo purposes. (TD).
+  */
+  .dropzone {
```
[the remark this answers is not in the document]
cjpillsbury: Yeah I wouldn't worry too much about solutioning right now, since we probably want to strike a balance between trying to anticipate use cases vs. users informing us about pain in their use cases. Also, fwiw, the `stylePropsToString` is more for handling the `React` -> Web Component translation, so folks can do things in a fairly "idiomatic to React" way even though, "under the hood," we are frequently using Web Components.

packages/mux-uploader/src/index.ts
```diff
@@ -0,0 +1,418 @@
...
+      if (this.svgCircle) {
+        // strokeDasharray is the size of dashes used to draw the circle with the size of gaps in between.
+        // If the dash number is the same as the gap number, no gap is visible: a full circle.
+        // strokeDashoffset defines where along our circle the dashes (in our case, a dash as long as the
+        // circumference of our circle) begins. The larger the offset, the farther into the circle you're
+        // starting the "dash". In the beginning, offset is the same as the circumference. Meaning, the visible
+        // dash starts at the end so we don't see the full circle. Instead we see a gap the size of the circle.
+        // When the percentage is 100%, offset is 0 meaning the dash starts at the beginning so we can see the circle. (TD).
+        this.svgCircle.style.strokeDasharray = `${getCircumference(this)} ${getCircumference(this)}`;
+        this.svgCircle.style.strokeDashoffset = `${getCircumference(this)}`;
```
luwes, line 295: 👍

clearlyTHUYDOAN:
@cjpillsbury @luwes Updated with last blocking item: README

Christian will look into pipeline failures 👼

cjpillsbury (approved):
🚢 🇮🇹

PR #239, opened 2022-06-02
feat: subtitles for M3U8 playlists, resume session, add cast-button

luwes:
This change adds support for text tracks loaded from M3U8 and simplifies the logic around text tracks.

The trackId's are now always retrieved from the remote player and the remote tracks matched with the local tracks.

- [x] tested changing sources (playbackId's) while casting and works
- [x] test wifi on/off toggling
- [x] fix resuming the cast session when the session persists after browser refresh
- [x] just noticed a bug when resuming a session the loading spinner keeps spinning! not locally, only on Vercel

### Non-blocking enhancements

- [ ] POLISH - ideally the mouseover shadow overlay stays visible while casting AND maybe the cast receiver name can be shown while casting
- [ ] currently when starting to cast playback-core (hls.js) will still start loading segments, should it be disabled while casting?
- [ ] the possible edge case @gkatsev mentioned that no local `<track>` corresponds with the remote tracks and toggling tracks off and on would break

### DIFF
--- a/packages/mux-player/src/helpers.ts
+++ b/packages/mux-player/src/helpers.ts
@@ -80,9 +80,10 @@ export function getCcSubTracks(el: MuxPlayerElement) {
 
 export const getLiveTime = (el: MuxPlayerElement) => {
   const { media } = el;
-  return media?._hls?.liveSyncPosition ?? media?.seekable.length
-    ? media?.seekable.end(media.seekable.length - 1)
-    : undefined;
+  return (
+    media?._hls?.liveSyncPosition ??
+    (media?.seekable.length ? media?.seekable.end(media.seekable.length - 1) : undefined)
+  );
 };
 
 export const seekToLive = (el: MuxPlayerElement) => {
--- a/shared/castable-video/castable-video.js
+++ b/shared/castable-video/castable-video.js
@@ -1,4 +1,4 @@
-/* global chrome, cast */
+/* global globalThis, chrome, cast */
 
 const CastableVideoMixin = (superclass) =>
   class CastableVideo extends superclass {
@@ -71,13 +71,17 @@ const CastableVideoMixin = (superclass) =>
       return CastableVideo.#castContext?.getCurrentSession();
     }
 
-    #castAvailable = false;
+    #editTracksInfo(request) {
+      return new Promise((resolve, reject) => {
+        CastableVideo.#currentSession?.getSessionObj().media[0].editTracksInfo(request, resolve, reject);
+      });
+    }
+
+    #isInit = false;
     #localState = { paused: false };
     #remoteState = { paused: false, currentTime: 0, muted: false };
     #remotePlayer;
     #remoteListeners = [];
-    #textTrackState = new Map();
-    #textTrackIdCount = 0;
     #enterCastCallback;
     #leaveCastCallback;
     #castChangeCallback;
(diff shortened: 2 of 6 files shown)


shared/castable-video/castable-video.js
```diff
@@ -414,20 +345,67 @@ const CastableVideoMixin = (superclass) =>
...
+      const remoteTracks = this.#remotePlayer.mediaInfo?.tracks ?? [];
+      const remoteSubtitles = remoteTracks.filter(({ type }) => type === chrome.cast.media.TrackType.TEXT);
+
+      const localSubtitles = [...this.textTracks].filter(({ kind }) => kind === 'subtitles' || kind === 'captions');
+
+      // Create a new array from the local subs w/ the trackId's from the remote subs.
+      const subtitles = remoteSubtitles
+        .map(({ language, name, trackId }) => {
+          // Find the corresponding local text track and assign the trackId.
+          const { mode } = localSubtitles.find((local) => local.language === language && local.label === name) ?? {};
```
gkatsev, line 372: Is it possible that the labels won't match due to different HLS parsing libraries?

gkatsev (commented):
Just the one question, really.

shared/castable-video/castable-video.js
luwes: it's possible yes. 

1 solution to this I had in mind was to temporarily add <track>'s with specific "cast-xyz" id's, basically replace the local ones with remote ones so Media Chrome can work with the remote ones while casting.

I briefly entertained the idea of overriding `video.textTracks`, I saw videojs did a similar thing, but for this use case it would be lots of extra code. 

(we'll have to do this soon enough though for `video.videoTracks` and video renditions)

gkatsev: For now, we can probably ignore this potential issue until we run into it and then fix it at that point.

Re-implementing Tracks would be a pain. If we do need to, we should consider whether it's worth spending the time extracting Video.js's Tracks into a separate package that we can use.

gkatsev (approved):

packages/mux-player/src/helpers.ts
```diff
@@ -80,9 +80,10 @@ export function getCcSubTracks(el: MuxPlayerElement) {
 
 export const getLiveTime = (el: MuxPlayerElement) => {
   const { media } = el;
-  return media?._hls?.liveSyncPosition ?? media?.seekable.length
-    ? media?.seekable.end(media.seekable.length - 1)
-    : undefined;
+  return (
+    media?._hls?.liveSyncPosition ??
+    (media?.seekable.length ? media?.seekable.end(media.seekable.length - 1) : undefined)
+  );
```
luwes, line 86: cc @cjpillsbury was a bug where `media?.seekable.end(media.seekable.length - 1)` was always tried to get because of condition priority. resulted in this error sometimes
SCR-20220602-qqa

shared/castable-video/package.json
```diff
@@ -12,5 +12,23 @@
   "repository": "https://github.com/muxinc/elements",
   "author": "Mux, Inc.",
   "license": "MIT",
-  "private": true
+  "private": true,
+  "eslintConfig": {
```
cjpillsbury, line 16: 😎

shared/castable-video/castable-video.js
cjpillsbury: Agree that this shouldn't be prioritized just yet. FWIW depending on what we need to do with textTracks, we can also always revisit the proxy impl we abandoned for circumventing the textTrack rendering bugs.

cjpillsbury (approved):

PR #240, opened 2022-06-03
fix: attrs not available in constructor

luwes:
This fixes the issue w/ `stream-type` warning, `start-time` not set etc. when the custom element is defined before parsing the HTML.

This unveiled another bug:

- [x] fix media-chrome bug where `media-gesture-receiver` is not associated as a state receiver
	https://github.com/muxinc/media-chrome/pull/238

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -140,6 +140,7 @@ const initialState = {
 };
 
 class MuxPlayerElement extends VideoApiElement {
+  #isInit = false;
   #tokens = {};
   #userInactive = true;
   #resizeObserver?: ResizeObserver;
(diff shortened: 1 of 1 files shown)


packages/mux-player/src/index.ts
```diff
@@ -163,6 +164,20 @@ class MuxPlayerElement extends VideoApiElement {
     super();
 
     this.attachShadow({ mode: 'open' });
+
+    // If the custom element is defined before the <mux-player> HTML is parsed
+    // no attributes will be available in the constructor (construction process).
+    // Wait until initializing attributes in the attributeChangedCallback.
+    // If this element is connected to the DOM, the attributes will be available.
+    if (this.isConnected) {
```
gkatsev, line 172: Ah, I thought it was going to be something like that. Forgot that `isConnected` is a thing.

gkatsev (approved):
Thanks!

PR #244, opened 2022-06-13
fix(mux-player): Update theme css to better account for audio vs. non…

cjpillsbury:
…-audio styling.


gkatsev (approved):
Tested, looks good.

PR #246, opened 2022-06-13
ci: add SHA hash to canary versions

luwes:
didn't end up using `lerna exec` because I could not get the `LERNA_PACKAGE_NAME` to expand properly in a bash command substitution.

this adds the SHA hash to the canary versions. 
also filters out beta versions when getting the latests versions because that would mess up the counter.

### DIFF
--- a/packages/mux-audio/package.json
+++ b/packages/mux-audio/package.json
@@ -44,7 +44,7 @@
     "build": "npm-run-all --parallel build:types 'build:esm -- --minify' 'build:iife -- --minify' 'build:cjs -- --minify' 'build:esm-module -- --minify'",
     "prebuild": "yarn clean",
     "prepublishOnly": "yarn build",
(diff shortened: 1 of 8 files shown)


packages/mux-audio/package.json
```diff
@@ -44,6 +44,7 @@
     "build": "npm-run-all --parallel build:types 'build:esm -- --minify' 'build:iife -- --minify' 'build:cjs -- --minify' 'build:esm-module -- --minify'",
     "prebuild": "yarn clean",
     "prepublishOnly": "yarn build",
+    "deploy:canary": "npm version $(semver $(npm view @mux-elements/mux-audio versions --json | jq -r '. - map(select(contains(\"beta\"))) | last' | sed -r 's/-[a-z0-9]{7}$//g') -i prerelease --preid canary)-$(git rev-parse --short HEAD) --git-tag-version false && npm publish --tag canary",
```
gkatsev, line 47: so, the jq command filters all the versions that contain "beta" version and then removes them from the list. Should we add `alpha` in here too?

luwes: yes, it shouldn't semver increase based on other prereleases. 
yes, if we would use `alpha` it should be added, will add to be safe

PR #249, opened 2022-06-14
fix(mux-player): Add cc btn to Vod XS chrome.

cjpillsbury:


luwes (approved):

gkatsev:
nit: "cc button" generally refers to captions button, not cast button.

PR #252, opened 2022-06-14
feat: add defaultMuted, defaultPlaybackRate props

luwes:


packages/mux-player/src/video-api.ts
```diff
@@ -217,6 +217,16 @@ class VideoApiElement extends HTMLElement {
     }
   }
 
+  get defaultPlaybackRate() {
+    return this.media?.defaultPlaybackRate ?? 1;
```
cjpillsbury, line 221: issue(blocker) since this is deferring to `media.defaultPlaybackRate`, we would need to update `mux-player-react` for functional parity. I can see two options (but am open to others):
1. Make `defaultPlaybackRate` work like `defaultMuted`, where we "funnel it through" the "`playbackrate`" attribute
1. Add prop and use `useObjectPropEffect` callback to do the mapping. This would mean folks would need to know the difference between the two.

cjpillsbury (changes requested):
Add some kind of support for `defaultPlaybackRate` to `mux-player-react`

cjpillsbury (approved):

PR #255, opened 2022-06-16
ci: try out turborepo

luwes:
trying out Turborepo for faster builds, tests. also caches remotely for the Vercel example apps.

---

We can fully replace Lerna with a publish flow like we use for canaries if we like.


cjpillsbury (approved):
🧑‍🔬 🔬 let's find out.

PR #261, opened 2022-06-17
fix: preload=metadata now triggers loadedmetadata

gkatsev:
Previously, we stopped loading on MANIFEST_LOADED, but that would
prevent enough data to be loaded for the video element to trigger
`loadedmetadata`. Instead, we should limit how much data hls.js is
allowed to load to the bare minimum. This means that only the first
segment should get loaded before playback begins. Then, when playback
begins, we can reset these values to the defaults and let hls.js load
data as normal.


luwes (approved):
LGTM

PR #264, opened 2022-06-21
fix(mux-uploader): Make progress, errors, completion accessible to VoiceOver

clearlyTHUYDOAN:
## Description
`mux-uploader` is currently not accessible to screen readers particularly around upload states 😢 

This PR adds initial screen reader functionality for VoiceOver 🌈 

- Announces upload completion regardless if progress bar is in focus ✅ 
- Announces upload error regardless if progress bar is in focus ✅ 
- Announces current progress when progress bar is in focus ✅ 
- Ticks when progress bar is in focus and is uploading ✅ 

## Notes
- Briefly tested with ChromeVox but the browser keeps crashing. Not sure why yet.
- Normally, when contents change within a live region, it can be announced. However, it seems like when changes to contents happen within different upload events, such as progress and completion, one won't be read.
   - Currently dealing with this using a combination of `innerHTML` and `sr-only` 
 - The "Try again" element is currently a span with `role=button`. We could use the native button and style it to look like text. Though, since it was previously made a span, it was faster to just throw the right role on it for the same access. 

## Demos

**Short upload to completion**
https://user-images.githubusercontent.com/22087604/174873899-4d151e7a-5ab8-443f-99c9-b8b828403fc5.mp4

**Short upload with error**
https://user-images.githubusercontent.com/22087604/174873919-c7922e79-bfab-4798-8700-252aea977294.mp4

**Medium-sized upload to completion**
https://user-images.githubusercontent.com/22087604/174873943-92d9e25d-4159-43ba-b5db-95cc905be686.mp4

### DIFF
--- a/packages/mux-uploader/src/index.ts
+++ b/packages/mux-uploader/src/index.ts
@@ -8,6 +8,15 @@ const styles = `
   background-color: var(--uploader-background-color, inherit);
 }
 
+.sr-only {
+  position:absolute;
+  left:-10000px;
+  top:auto;
+  width:1px;
+  height:1px;
+  overflow:hidden;
+  }
+
 p {
   color: black;
 }
@@ -50,7 +59,7 @@ button:active {
   width: 100%;
 }
 
-.radial-type, .bar-type, .upload-status, .retry-message, .text-container {
+.radial-type, .bar-type, .upload-status, .retry-button, .text-container {
   display: none;
 }
 
@@ -62,7 +71,7 @@ button:active {
   display: none;
 }
 
-.retry-message {
+.retry-button {
   color: #e22c3e;
   text-decoration-line: underline;
   cursor: pointer;
@@ -108,7 +117,7 @@ button:active {
   display: none;
 }
 
-:host([upload-error]) .retry-message {
+:host([upload-error]) .retry-button {
   display: inline-block;
 }
 
@@ -168,9 +177,11 @@ template.innerHTML = `
 
 <p class="upload-instruction" id="upload-instruction">Drop file to upload</p>
 
+<div class="sr-only" id="sr-only" aria-live="polite"></div>
+
 <div class=text-container>
-  <span class="status-message" id="status-message"></span>
-  <span class="retry-message" id="retry-message">Try again</span>
+  <span class="status-message" id="status-message" aria-live="polite"></span>
+  <span class="retry-button" id="retry-button" role="button" tabindex="0">Try again</span>
 </div>
 
 <input type="file" />
@@ -178,7 +189,8 @@ template.innerHTML = `
 <slot name="custom-progress"><p class="upload-status" id="upload-status"></p></slot>
 
 <div class="bar-type">
-  <div class="progress-bar" id="progress-bar"></div>
+  <div role="progressbar" aria-description="A bounded progress bar from 0 to 100" aria-valuemin="0"
+  aria-valuemax="100" class="progress-bar" id="progress-bar" tabindex="0"></div>
 </div>
 <div class="radial-type">
   <svg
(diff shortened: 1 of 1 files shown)


clearlyTHUYDOAN:
I checked that coverage report and it doesn't seem to be related to my direct code changes 🤔

gkatsev:
@clearlyTHUYDOAN I would ignore codecov if it goes down by less than 0.1%

packages/mux-uploader/src/index.ts
```diff
@@ -337,8 +365,7 @@ class MuxUploaderElement extends HTMLElement {
 
   setProgress(percent: number) {
     if (this.uploadPercentage) this.uploadPercentage.innerHTML = `${Math.floor(percent)}%`;
-    this.setAttribute('aria-valuenow', `${Math.floor(percent)}`);
-    this.setAttribute('aria-valuetext', `${Math.floor(percent)} percent`);
+    this.progressBar?.setAttribute('aria-valuenow', `${Math.floor(percent)}`);
```
cjpillsbury, line 368: non-blocking: why remove `aria-valuetext`?

packages/mux-uploader/src/index.ts
```diff
@@ -168,17 +177,20 @@ template.innerHTML = `
 
 <p class="upload-instruction" id="upload-instruction">Drop file to upload</p>
 
+<div class="sr-only" id="sr-only" aria-live="polite"></div>
+
 <div class=text-container>
-  <span class="status-message" id="status-message"></span>
-  <span class="retry-message" id="retry-message">Try again</span>
+  <span class="status-message" id="status-message" aria-live="polite"></span>
```
cjpillsbury, line 183: non-blocking: I think you may want to give this a [`role="status"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role) to use it how you're intending? See how we've implemented [media-loading-indicator](https://github.com/muxinc/media-chrome/blob/main/src/js/media-loading-indicator.js#L65)

packages/mux-uploader/src/index.ts
```diff
@@ -372,6 +400,7 @@ class MuxUploaderElement extends HTMLElement {
     }
 
     this.setAttribute('upload-in-progress', '');
+    this.progressBar?.focus();
```
cjpillsbury, line 403: 😎

packages/mux-uploader/src/index.ts
```diff
@@ -394,7 +426,19 @@ class MuxUploaderElement extends HTMLElement {
...
 
     upload.on('success', () => {
-      console.info('Upload complete!');
+      const successMessage = 'Upload complete!';
+
+      if (this.statusMessage) {
+        this.statusMessage.innerHTML = successMessage;
+      }
+
+      // TO-DO: It seems like statusMessage cannot be updated within two different events. (TD).
```
cjpillsbury, line 435: This should be fine, though could be a timing issue. One way you could pin that down would be to break out the `srOnlyText.innerHTML` updates into its own function or method.

cjpillsbury (changes requested):
A few small-ish asks, but pumped about this.

packages/mux-uploader/src/index.ts
```diff
@@ -168,17 +177,20 @@ template.innerHTML = `
...
+  <span class="retry-button" id="retry-button" role="button" tabindex="0">Try again</span>
 </div>
 
 <input type="file" />
 <slot name="custom-button"><button type="button">Upload video</button></slot>
 <slot name="custom-progress"><p class="upload-status" id="upload-status"></p></slot>
 
 <div class="bar-type">
-  <div class="progress-bar" id="progress-bar"></div>
+  <div role="progressbar" aria-description="A bounded progress bar from 0 to 100" aria-valuemin="0"
```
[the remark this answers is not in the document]
clearlyTHUYDOAN: I did feel like this description was too long while using a screen reader. Will change to a concise description.

packages/mux-uploader/src/index.ts
cjpillsbury: Cool I think that's a good enough reason to prefer `aria-valuenow` for now. We can always revisit from a UX perspective which vocalizations would be preferred by users (where these tradeoffs bottom out imo, esp if the dev overhead is minimal)

cjpillsbury (approved):
😎 lgtm!

PR #265, opened 2022-06-21
docs(mux-uploader): Update README to replace copy pasta from mux-player.

clearlyTHUYDOAN:
## Description

Some text from copying over `mux-player`'s markdown table was leftover 🍝 

Updated text to be relevant to `mux-uploader` 🌈

### DIFF
--- a/packages/mux-uploader/README.md
+++ b/packages/mux-uploader/README.md
@@ -59,10 +59,36 @@ If you are using ECMAScript modules, you can also load the `mux-uploader.mjs` fi
 
 ```html
 <body>
-  <p></p>
+  <!-- Upload button by itself with default drag an drop scoped to the space it takes up. Displays upload progress in text as percentage.-->
+  <mux-uploader url="authenticated-url" type="bar" status></mux-uploader>
 
-  <mux-uploader url="authenticated-url" id="uploader"></mux-uploader>
+  <!-- Upload button by itself with drag an drop disabled. Does not display text percentage.-->
+  <mux-uploader url="authenticated-url" type="bar" disable-drop></mux-uploader>
+
+  <!-- Upload button with access to additional drag and drop features via slots i.e. fullscreen drag and drop with text overlay (work-in-progress).-->
+  <mux-uploader url="authenticated-url">
+    <mux-uploader-drop slot="dropzone" text="Upload to stream.new" fullscreen overlay></mux-uploader-drop>
+  </mux-uploader>
 </body>
+```
+
+## Drag and Drop
+
+The `mux-uploader`, whether you use `mux-uploader-drop` and its additional features or not i.e. fullscreen and text overlay, has basic drag and drop functionality available to it by default. If you'd rather shop your own drag and drop solution, you can disable the default drag and drop on `mux-uploader` and dispatch a custom `file-ready` event when you need to upload. `mux-uploader` will handle the upload upon receiving the event.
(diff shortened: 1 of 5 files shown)


packages/mux-uploader/README.md
```diff
@@ -61,10 +61,7 @@ If you are using ECMAScript modules, you can also load the `mux-uploader.mjs` fi
 <body>
   <p></p>
 
-  <mux-uploader
-    url="authenticated-url"
-    id="uploader"
-  ></mux-player>
+  <mux-uploader url="authenticated-url" id="uploader"></mux-uploader>
 </body>
```
dylanjha, line 65: can we expand on this example, or add another example about how to use `<mux-uploader-drop>`? Based on these docs I'm not sure how to use that other component

packages/mux-uploader/README.md
```diff
@@ -77,9 +74,9 @@ If you are using ECMAScript modules, you can also load the `mux-uploader.mjs` fi
 | `url`                | `string`  | The authenticated URL that your file will be uploaded to. Check out the [direct uploads docs](https://docs.mux.com/guides/video/upload-files-directly#1-create-an-authenticated-mux-url) for how to create one. Required. | `undefined` |
 | `id`                 | `string`  | An ID that allows `mux-uploader-drop` to locate `mux-uploader`. Required.                                                                                                                                                 | N/A         |
```
dylanjha, line 75: is `id` *always* required? even if I'm not using `mux-uploader-drop`?

packages/mux-uploader/README.md
```diff
@@ -77,9 +74,9 @@ If you are using ECMAScript modules, you can also load the `mux-uploader.mjs` fi
...
-| `status`             | `boolean` | This is an arbitrary title for your video that will be passed in as metadata into Mux Data. Adding a title will give you useful context in your Mux Data dashboard. (optional, but encouraged)                            | false       |
+| `upload-in-progress` | `boolean` | Toggles visual status of progress bar while upload is n progress.                                                                                                                                                         | false       |
+| `upload-error`       | `boolean` | Toggles visual status of progress bar when upload encounters an error.                                                                                                                                                    | false       |
+| `status`             | `boolean` | Toggles text status of progress bar, which is a percentage by default.                                                                                                                                                    | false       |
```
dylanjha, line 79: looks better 👍🏼 

Can you clarify how someone would use these attributes? As a developer what is the use case for when I would want to show or hide the progress bar or error?

Or are these attributes that get sent internally by the web component and I'm not supposed to be manipulating them from the outside?

I'm not quite following `status` which says is `boolean` but then says it is a percentage by default. What is the alternative to percentage?

dylanjha (changes requested):

packages/mux-uploader/README.md
clearlyTHUYDOAN: Will work on clarifying! Thanks for providing insight as a person totally new to the docs.

clearlyTHUYDOAN:
@dylanjha Added some more examples, clarified certain attributes, found some non-blocking TO-DO items, and renamed a custom event to be more relevant to its use cases i.e. `mux-drop` => `file-ready` (also used for click events).

dylanjha (approved):
Love it, much better 👏🏼 

I may have some suggestions for future iterations, but for now this is a big improvement!

PR #269, opened 2022-06-24
feat: add ability to choose a Media Theme via attribute

luwes:
this change adds the ability to use a different theme via a simple `theme="media-theme-goldshine"` attribute ✨ 

it required some changes to the html tagged template renderer. 
added support for unsafe static values that can make part of the HTML string to be process by the template parser.

SCR-20220624-kk4


types/media-chrome.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module 'media-chrome' {
```
cjpillsbury, line 1: 😎

cjpillsbury (approved):
👍 lgtm

PR #271, opened 2022-06-27
build:change all references from @mux-elements npm scope to @mux npm …

cjpillsbury:
…scope.


gkatsev (approved):
When we're ready for the change, 👍🏻

PR #283, opened 2022-07-05
test: make events dispatch test more reliable

gkatsev:
Removing emptied and waiting increase the reliability of the test since
these two events are less likely to occur on any given playback compared
to the other events that are triggered in that test.


cjpillsbury (approved):
😎

PR #286, opened 2022-07-07
Mux elements scope deprecation

gkatsev:
When we are ready to take the plunge, we can push this branch upstream, which should allow us to then dispatch the `cd.yml` workflow on this branch to publish the `console.warn` update and then run the `deprecate.yml` workflow to deprecate the old packages on npm.

In the packages, I used `console.warn` instead of `console.error` because a `console.error` could potentially fail a build. In addition, it's logged only in the following packages to minimize the chance that there are multiple logs:
- `mux-audio`
- `mux-video`
- `mux-audio-react`
- `mux-video-react`
- `mux-uploader`

### DIFF
--- a/.github/workflows/deprecate.yml
+++ b/.github/workflows/deprecate.yml
@@ -21,13 +21,13 @@ jobs:
           # this line is required for the setup-node action to be able to run the npm publish below.
           registry-url: 'https://registry.npmjs.org'
           cache: yarn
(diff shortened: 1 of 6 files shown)


gkatsev:
rebased against the [build/mux-elements-scope-cutoff branch](https://github.com/muxinc/elements/tree/build/mux-elements-scope-cutoff)

luwes (approved):
LGTM 👍

PR #291, opened 2022-07-12
feat(playback-core, mux-player): Add support to removeTextTrack. Remo…

cjpillsbury:
…ve all identified tracks on hls destroy. Add methods to mux-player.

Adding for forward-looking use cases with timed metadata.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -4,7 +4,7 @@ import '@mux/playback-core';
 import { MediaController } from 'media-chrome';
 import MediaThemeMux from './media-theme-mux/media-theme-mux';
 import MuxVideoElement, { MediaError } from '@mux/mux-video';
-import { Metadata, StreamTypes } from '@mux/playback-core';
+import { Metadata, StreamTypes, addTextTrack, removeTextTrack } from '@mux/playback-core';
 import VideoApiElement, { initVideoApi } from './video-api';
 import { getPlayerVersion, isInLiveWindow, seekToLive, toPropName } from './helpers';
 import { template } from './template';
@@ -907,6 +907,18 @@ class MuxPlayerElement extends VideoApiElement {
    */
   set storyboardToken(val) {
     this.setAttribute(PlayerAttributes.STORYBOARD_TOKEN, `${val}`);
+  }
+
+  addTextTrack(kind: TextTrackKind, label: string, lang?: string, id?: string) {
+    const mediaEl = this.media?.nativeEl;
+    if (!mediaEl) return;
+    return addTextTrack(mediaEl, kind, label, lang, id);
+  }
+
+  removeTextTrack(track: TextTrack) {
+    const mediaEl = this.media?.nativeEl;
+    if (!mediaEl) return;
+    return removeTextTrack(mediaEl, track);
   }
 }
 
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -920,6 +920,10 @@ class MuxPlayerElement extends VideoApiElement {
     if (!mediaEl) return;
     return removeTextTrack(mediaEl, track);
   }
+
+  get textTracks() {
+    return this.media?.textTracks;
+  }
 }
 
 export function getVideoAttribute(el: MuxPlayerElement, name: string) {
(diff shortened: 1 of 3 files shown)


gkatsev (commented):
Since we're adding convenience methods to add and remove tracks, should we also provide a way to get a list of tracks?

cjpillsbury:
> Since we're adding convenience methods to add and remove tracks, should we also provide a way to get a list of tracks?

Yeah good call. I'll add that.

packages/mux-player/src/index.ts
```diff
@@ -908,6 +908,18 @@ class MuxPlayerElement extends VideoApiElement {
   set storyboardToken(val) {
     this.setAttribute(PlayerAttributes.STORYBOARD_TOKEN, `${val}`);
   }
+
+  addTextTrack(kind: TextTrackKind, label: string, lang?: string, id?: string) {
```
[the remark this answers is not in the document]
cjpillsbury: I think as long as we're "extending" the API, this should be fine (aka the additional parameterization comes at the end and is optional).

packages/playback-core/src/tracks.ts
```diff
@@ -127,6 +124,19 @@ function createTextTrack(
     trackEl.id = id;
   }
   trackEl.track.mode = 'disabled';
+  // Add data attribute to identify tracks that should be removed when switching sources/destroying hls.js instance.
+  trackEl.setAttribute('data-removeondestroy', '');
```
luwes, line 128: can we remember these tracks without setting any attributes? would a weakmap work?

cjpillsbury: > can we remember these tracks without setting any attributes? would a weakmap work?

@luwes We can, but I actually prefer using the attribute for:
1. Simplicity (we only add it on creation and use it via query selectors to identify items to remove)
2. Visibility (programmatic and devs trying to e.g. debug)
3. Avoiding module level state (harder to test/debug at a unit level since we now have modules whose state changes over time once loaded)
4. Applying a valid use case for data attributes (a data-driven identifier associated with an element instance)
5. Versatility (advanced users can easily add the attribute if they want to guarantee the track gets removed when the src changes without having to build any architecture)

If you feel strongly about the change and see particular advantages to it, we can get a gut check from the rest of the team and implement in whichever way the team as a whole prefers.

luwes: cool I'm ok with it, my preference for keeping the data- attribute out was more for not adding special HTML to the document.

I would argue against point 5 and best to keep it out of their control because it can already be done with the API methods.

packages/playback-core/src/tracks.ts
```diff
@@ -73,11 +72,9 @@ export function setupTracks(
...
     mediaEl.textTracks.removeEventListener('change', changeHandler);
 
-    const trackEls = mediaEl.querySelectorAll('track');
+    // Use data attribute to identify tracks that should be removed when switching sources/destroying hls.js instance.
+    const trackEls: NodeListOf<HTMLTrackElement> = mediaEl.querySelectorAll('track[data-removeondestroy]');
     trackEls.forEach((trackEl) => {
-      if (!(trackEl.id && ['subtitles', 'captions'].includes(trackEl.kind))) return;
-      if (!hls.subtitleTracks.some(({ type }, idx) => trackEl.id === `${type.toLowerCase()}${idx}`)) return;
-
       mediaEl.removeChild(trackEl);
```
luwes, line 78: nit (non-blocking): `trackEl?.remove()`

packages/playback-core/src/tracks.ts
```diff
@@ -127,6 +124,19 @@ function createTextTrack(
...
+
+export function removeTextTrack(mediaEl: HTMLMediaElement, track: TextTrack) {
+  const trackEl: HTMLTrackElement = Array.prototype.find.call(
+    mediaEl.querySelectorAll('track'),
+    (trackEl: HTMLTrackElement) => trackEl.track === track
+  );
+  if (!trackEl) {
+    return;
+  }
+  mediaEl.removeChild(trackEl);
```
luwes, line 141: nit (non-blocking): `trackEl?.remove()`

luwes (approved):
LGTM

PR #293, opened 2022-07-12
feat(mux-uploader, mux-uploader-react): Reset state to allow users to…

cjpillsbury:
… re-select same file when retrying a failed upload. Add error messaging for when url is unset.


clearlyTHUYDOAN (approved):
Smoke tested it on the CRA preview. LGTM :shipit:

PR #296, opened 2022-07-18
shorten prefer-mse explanation

tomByrer:
Puts benefit up front


luwes:
thanks for the contribution @tomByrer ! looks good to me, I'll let others chime in too

PR #298, opened 2022-07-18
fix: upgrade MC v0.8.1

luwes:
https://github.com/muxinc/media-chrome/pull/271
fixes jumpy time range with playbackRate other than 1


cjpillsbury (approved):
🎉

PR #300, opened 2022-07-20
Fix/metadata cases

cjpillsbury:
This should fix for mux-player and mux-player react. We had some assumptions in the way metadata was passed down to the underlying `mux-video` instance (and eventually playback-core). This makes sure that:
* we pass _**all**_ metadata down to the `mux-video` instance
* we _**always**_ set metadata directly on the `mux-video` instance (via the force upgrade/`init()`)
* we don't conflate `metadata` values with the corresponding attributes (since that is managed/defined by the underlying `mux-video` instance)

This also adds a simple function to derive some metadata key/value pairs from our `mediaAsset` in the NextJS app for `MuxPlayer` (react) to facilitate testing. You should be able to inspect network traffic of the `litix` events to see expected changes in `vid`/`vtt` (`video_id` and `video_title`, respectively) when changing the selected `mediaAsset`.


luwes (approved):
LGTM 👍

PR #305, opened 2022-07-21
sc-35838 add codeowners

akemner:
sc-35838


cjpillsbury (changes requested):
question(blocker): Holding off until we know the answer to https://github.com/github/docs/issues/18353 and can assess.

PR #316, opened 2022-07-28
fix: CustomVideo.nativeEl being null in some cases

### LINKED ISSUE #314: Bug: TypeError: Invalid value used as weak map key
### Checklist

- [x] Verified whether your issue exists in the latest version Mux Element(s) release: `npm install @mux/[element_name]` or `yarn @mux/[element_name]`.
  - (NOTE: If the issue is not occurring in the latest version, we may not resolve it.)
- [x] Confirmed that this issue hasn't [already been filed](https://github.com/muxinc/elements/issues?q=is%3Aopen+is%3Aissue+label%3Abug).

- mux-video

- Chrome

(For example: macOS, Windows, iOS, Android)
- macOS

(For example: npm package, `<script>` tag)

vue 2 with a component:
```
      <mux-video
          :playback-id="video.playbackId"
      ></mux-video>
```

### Describe the bug

I am getting the follow error in the console, i can see a thumbnail working correctly and the controls show but its not loading the video in and i cannot press play:

**_Error: Uncaught TypeError: Invalid value used as weak map key_**

## Steps To Reproduce

1. Just loading a component
2. Worked previously

Link to code example:

## The current behavior
Video will not play

## The expected behavior
Video will play

luwes:
fix #314

this change fixes a bug where `customVideo.nativeEl` or `customAudio.nativeEl` would be undefined in the playback core `initialize` function causing the set of the weakmap key to error.

nativeEl was undefined because the init flow of the custom media elements was not called yet.
moving the super.attributeChangedCallback to the top fixed this issue

### DIFF
--- a/examples/nextjs-with-typescript/pages/index.tsx
+++ b/examples/nextjs-with-typescript/pages/index.tsx
@@ -47,6 +47,11 @@ function HomePage() {
         </li>
         <li>
           <h3>
+            <Link href="/mux-audio">Mux Audio (Web Component) Demo</Link>
+          </h3>
+        </li>
+        <li>
+          <h3>
             <Link href="/mux-player">Mux Player (Web Component) Demo</Link>
           </h3>
         </li>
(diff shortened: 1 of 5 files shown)


cjpillsbury (approved):
LGTM!

PR #317, opened 2022-07-29
chore: use Node 16 LTS

luwes:
update Node to version 16.

The extracting in GH action is removed because the 2 versions, I think it's also not really needed to be automated. if we upgrade again I'm pretty sure we will not miss updating it in the GH actions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@mux/elements",
   "engines": {
-    "node": "^14.17.0",
+    "node": "^14.17.0 || ^16.13.0",
     "npm": "^7.17.0"
   },
   "version": "0.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@mux/elements",
   "engines": {
-    "node": "^14.17.0",
+    "node": "^14.17.0 || ^16.13.0",
     "npm": "^7.17.0"
```
gkatsev, line 5: non-blocking: we should also update this npm version to be 8.x. It's what's ships with 16.x.

gkatsev (approved):

luwes:
SCR-20220801-f24

gkatsev (approved):

PR #318, opened 2022-08-01
Media Chrome Update and hotkeys support

gkatsev:


packages/mux-player/src/index.ts
```diff
@@ -533,6 +535,10 @@ class MuxPlayerElement extends VideoApiElement {
         }
         break;
       }
+      case PlayerAttributes.NOHOTKEYS: {
+        this.mediaController?.setAttribute(PlayerAttributes.NOHOTKEYS, newValue);
```
cjpillsbury, line 539: question(blocker): Any reason why this isn't using the template to update (See: https://github.com/muxinc/elements/blob/main/packages/mux-player/src/media-theme-mux/media-theme-mux.ts#L77)? If not, should refactor to prefer template-based updates.

cjpillsbury (changes requested):
One small callout, otherwise 👌

packages/mux-player/src/index.ts
luwes: agree here, avoid direct mutations to the DOM if possible in player

gkatsev: Yeah, that's what I was trying to do at first, but wasn't able to get it working. Have one more thought, but may need a second set of eyes for this.

gkatsev: Looks like it does eventually get through without this change. However, it doesn't happen until I resize the player. Thoughts on how to make sure it gets dispatched immediately?

cjpillsbury (approved):
👍

PR #321, opened 2022-08-04
feat: add `controlslist` blocklist

luwes:
this change adds a `controlslist` attribute and `controlsList` property (similar to DOMTokenList type) to hide specific controls. This is the list that is currently added and these areas in the player can be hidden:

- `notop`
- `nobottom`
- `nocenter`
- `nocenterplay`
- `nocenterseekbackward`
- `nocenterseekforward`
- `noplay`
- `noseekbackward`
- `noseekforward`
- `nomute`
- `nocaptions`
- `noairplay`
- `nopip`
- `nocast`
- `nofullscreen`
- `noplaybackrate`
- `novolumerange`
- `notimerange`
- `notimedisplay`
- `noremoteplayback`
- `noseeklive`


packages/mux-player/src/helpers.ts
```diff
@@ -112,3 +112,91 @@ export const isInLiveWindow = (el: MuxPlayerElement) => {
...
+      if (!this.contains(t)) this.#tokens.push(t);
+    });
+    this.#el.setAttribute(this.#attr, `${this.value}`);
+  }
+
+  remove(...tokens: string[]) {
+    tokens.forEach((t) => {
+      this.#tokens.splice(this.#tokens.indexOf(t), 1);
+    });
+    this.#el.setAttribute(this.#attr, `${this.value}`);
```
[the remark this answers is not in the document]
gkatsev: Interesting, just tried this with `class` and `classList` and the `class` attribute sticks around when you `.remove()` the last item, but calling `removeAttribute('class')` does work, which currently doesn't work with this implementation.

Either way, I don't think it's a blocker, but would be nice to have.

gkatsev (commented):
This is great! Love the `controlsList` property being similar to `classList`.

Had one question around the attribute sticking around.

gkatsev:
Oh, we also want to document this in the README

packages/mux-player/src/helpers.ts
```diff
@@ -112,3 +112,91 @@ export const isInLiveWindow = (el: MuxPlayerElement) => {
   }
   return false;
 };
+
+export class AttributeTokenList {
```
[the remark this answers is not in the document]
luwes: yup DOMTokenList which is what `controlsList` is as well
https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList

examples/vanilla-ts-esm/public/mux-player.html
```diff
@@ -90,6 +90,7 @@
      -->
 
     <mux-player
+      controlslist="noplay nofullscreen noremoteplayback noplaybackrate"
```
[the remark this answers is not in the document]
luwes: yup scusi, will remove soon

packages/mux-player/README.md
```diff
@@ -117,6 +117,50 @@ Enable the [Google Cast](https://developers.google.com/cast) button in the contr
...
+```html
+<style>
+  mux-player::part(center play button) {
+    display: none;
+  }
+</style>
+<mux-player playback-id="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"></mux-player>
+```
+
+Supported **parts**:
```
cjpillsbury, line 137: suggestion(non-blocking): This is a good first pass, but either now or as a followup we should discuss how these parts relate to the controls and "regions" of the UI

packages/mux-player/src/helpers.ts
```diff
@@ -112,3 +112,99 @@ export const isInLiveWindow = (el: MuxPlayerElement) => {
...
```
cjpillsbury, line 116: suggestion(non-blocker): Consider moving this into its own module, since it's an implementation of a generic class.

cjpillsbury, line 116: question(non-blocker): Any reason we didn't build on top of `DOMTokenList`? (https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList)

packages/mux-player/src/index.ts
```diff
@@ -568,6 +562,18 @@ class MuxPlayerElement extends VideoApiElement {
         break;
       }
     }
+
+    const shouldClearState = [
```
cjpillsbury, line 566: question(blocker): Why was this change necessary? Feels out of scope. (or is this is rebase diff issue?)

packages/mux-player/README.md
```diff
@@ -201,6 +246,7 @@ If you prefer to use the in-code MSE-based engine (currently hls.js) whenever po
...
 | `volume`                                      | Is a double indicating the audio volume, from 0.0 (silent) to 1.0 (loudest).                                                                                                                                                                                                                                        | `1`         |
 | `customDomain`                                | Is a `String` that assigns a custom domain to be used for Mux Video.                                                                                                                                                                                                                                                | `undefined` |
+| `controlsList`                                | Is a `DOMTokenList` similar to `classList` which offers a way to hide the controls elements/buttons that are being shown by the player. For more, see the section on [`controlslist`](#controlslist)                                                                                                                | `''`        |
```
cjpillsbury, line 249: chore(blocker): If this is going to be read-only by design, we should mention that here somewhere.

packages/mux-player/src/helpers.ts
gkatsev: It isn't possible to instantiate the built-in DOMTokenList, unfortunately.

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -155,92 +155,95 @@ const ChromeRenderer = (props: ThemeMuxTemplateProps) => {
...
-`;
+  </media-play-button>`;
 
+type SeekBackwardButtonProps = { backwardSeekOffset: string; part?: string };
 // prettier-ignore
-type SeekBackwardButtonProps = { backwardSeekOffset: string, part?: string };
-const MediaSeekBackwardButton = ({ backwardSeekOffset, part = parts.seekBackward }: SeekBackwardButtonProps) => html`
-  <media-seek-backward-button seek-offset="${backwardSeekOffset}" part="${part}">
+const MediaSeekBackwardButton = ({ backwardSeekOffset, part = parts.bottom }: SeekBackwardButtonProps) => html`
+  <media-seek-backward-button seek-offset="${backwardSeekOffset}" part="${part} ${parts.seekBackward}">
```
cjpillsbury, line 167: 👌 this feels clean for part concatenation.

packages/mux-player/src/helpers.ts
gkatsev: this seems like it'll be useful when we'd want a `hotkeys` list like controlslist

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -309,15 +312,15 @@ export const VodChromeExtraSmall = (props: ThemeMuxTemplateProps) => html`
 // prettier-ignore
 export const VodChromeSmall = (props: ThemeMuxTemplateProps) => html`
   <media-control-bar slot="top-chrome" style="justify-content: flex-end;">
-    ${MediaCaptionsButton(props)}
-    ${MediaAirplayButton()}
-    ${MediaCastButton()}
-    ${MediaPipButton()}
+    ${MediaCaptionsButton({ ...props, part: parts.top })}
```
cjpillsbury, line 315: 👌 thought(non-blocker): If these will be used primarily to identify "region", we may want to rename the property

packages/mux-player/src/template.ts
```diff
@@ -14,16 +14,54 @@ import { i18n, stylePropsToString } from './utils';
 import type { MuxTemplateProps } from './types';
 import { StreamTypes } from '@mux/playback-core';
 
+const controlsListStyles: Record<string, string> = {
```
cjpillsbury, line 17: thought(non-blocker): I believe we should be able to make this non-dynamic via CSS + list-inclusion attribute selectors (`[controlslist~=notop]`). This shouldn't block the merge, though.

packages/mux-player/src/template.ts
```diff
@@ -87,12 +125,10 @@ export const content = (props: MuxTemplateProps) => html`
           ? getSrcFromPlaybackId(props.playbackId, { domain: props.customDomain, token: props.tokens.playback })
           : false
       }"
-      cast-stream-type="${[StreamTypes.LIVE, StreamTypes.LL_LIVE].includes(props.streamType as any) ? 'live' : false}"
+      cast-stream-type="${isLive(props) ? 'live' : false}"
```
cjpillsbury, line 128: thought(non-blocker): This looks functionally equivalent, so no worries on removing even though it's out of scope.

packages/mux-player/src/template.ts
```diff
@@ -14,16 +14,54 @@ import { i18n, stylePropsToString } from './utils';
...
   </style>
   ${content(props)}
 `;
 
 const flatCSSParts = Object.values(parts).flatMap((group) => group.split(' '));
 const forwardUniqueCSSParts = [...new Set(flatCSSParts)].join(', ');
 
+const isLive = (props: MuxTemplateProps) => [StreamTypes.LIVE, StreamTypes.LL_LIVE].includes(props.streamType as any);
+
+const isLiveOrDVR = (props: MuxTemplateProps) =>
```
cjpillsbury, line 62: suggestion(non-blocking): If we're going to use these functions, we should probably defer the "live" half to `isLive` here. i.e.:

```js
isLive(props) || [StreamTypes.DVR, StreamTypes.LL_DVR].includes(props.streamType as any);
```

packages/mux-player/src/template.ts
```diff
@@ -105,21 +141,24 @@ export const content = (props: MuxTemplateProps) => html`
...
-        props.onSeekToLive?.(evt);
-        if (props.paused) {
-          const playRequestEvt = new CustomEvent('mediaplayrequest', { composed: true, bubbles: true });
-          (this as HTMLButtonElement).dispatchEvent(playRequestEvt);
-        }
-      }}"
-    >
-      Live
-    </button>
+    ${
```
cjpillsbury, line 144: question(non-blocker): Why was this refactored? Was slotting this causing problems (It shouldn't have)? If not, I'd lean towards always slotting, since it's less presumptuous.

cjpillsbury (changes requested):
a few minor callouts, otherwise this is 😎

packages/mux-player/src/helpers.ts
cjpillsbury: well that's simultaneously a good reason and a bad reason.

gkatsev: yeah, I even just tried if I could extend it, but it doesn't let you because DOMTokenList is only an interface, unfortuantely.

cjpillsbury: ah if that's the case (I again don't think this should be a blocker) but @luwes might be worth trying to add `implements DOMTokenList` and at least see if there are any discrepancies between your impl and the "official" interface

Here's the def:
```ts
interface DOMTokenList {
    [Symbol.iterator](): IterableIterator<string>;
    entries(): IterableIterator<[number, string]>;
    keys(): IterableIterator<number>;
    values(): IterableIterator<string>;
}
```

packages/mux-player/src/index.ts
luwes: I moved this code block from above to below the switch block because once we `render()` we want the `controlsList` to be updated with the new value. it's used in template.ts to render the styles

```js
    ${[...props.controlsList]
      .map((token) => controlsListStyles[token])
      .filter(Boolean)
      .join('\n')}
```

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
luwes: I considered this but region feels so much geo related to me, not sure if this is the same for others.
I'll leave it for now

packages/mux-player/src/template.ts
luwes: `part="${isLive(props) ? 'top' : 'bottom'} seek-live button"` this is the needed change and I noticed this is only required for live like streams so it made sense to me to not render this for VOD. no big changes in here though, just added 2 conditionals

cjpillsbury: hrm. Did you not change the rendering logic to conditionally render based on live vs. not? Currently, that shouldn't have any impact either way, but it doesn't look like it was necessary.

cjpillsbury: aka
* before: always rendered the seek live `<button>` and let the theme figure out if it "cared" about the slotted element "under the hood"
* after: conditionally render the seek live `<button>` "from the outside"

luwes: yes, in the case of VOD we won't ever need a seek to live button so it shouldn't be rendered at all. 
I should've done this from the start but I didn't realize it at the time.

cjpillsbury: not going to push back here as a blocker to this PR, but there are some benefits to always passing in slotted elements as we start having robust support of themes, since that lets theme implementors decide whether or not they want to use the slotted element and under what conditions.

cjpillsbury (approved):

gkatsev (approved):

PR #323, opened 2022-08-05
feat(mux-uploader): Add support for directly setting file reference.

cjpillsbury:
Completely untested and would want to move some bits and bobs around, but this should work for direct file ref support.
Also need to add prop to react component, but that should be an easy lift I can hit later this evening.


luwes:
closing as it's stale and has conflicts

PR #331, opened 2022-08-10
fix(mux-uploader / mux-uploader-react): Error event collisions

clearlyTHUYDOAN:
## Description

A user was getting this Typescript error when using `onError`:

Screen Shot 2022-08-09 at 3 53 04 PM

After some digging, it turns out...

- [React elements](https://reactjs.org/docs/events.html) have an event handler called `onError` 💡 
- [HTML elements](https://developer.mozilla.org/en-US/docs/Web/API/Element/error_event) have an event handler called `error` 💡 

This PR renames both...
- `mux-uploader`'s custom event dispatch name from `error` to `uploaderror`; and ↔️ 
- `MuxUploader`'s error handler from `onError` to `onUploadError` ↔️ 
- Updates related `README`s and example(s) ✅ 

...which _should_ resolve the Typescript error and prevent collisions 🌈 

## Testing

- [x] Locally on the `next.js` app ✨


cjpillsbury (approved):
tl;dr - Fine with moving forward with this solution, but feel free explore a refactor that uses `ErrorEvent` instead of `CustomEvent` instead.

PR #332, opened 2022-08-11
Feat/expose control props

cjpillsbury:
Adding props & attributes to mux player react to support configurable:

- [x] playback rates
- [x] hiding duration
- [ ] defaulting to show remaining time

This is based on current goals for control customization.


packages/mux-player/src/template.ts
```diff
@@ -80,6 +80,8 @@ export const content = (props: MuxTemplateProps) => html`
     default-hidden-captions="${props.defaultHiddenCaptions}"
     forward-seek-offset="${props.forwardSeekOffset}"
     backward-seek-offset="${props.backwardSeekOffset}"
+    playbackrates="${props.playbackRates ?? false}"
+    hide-duration="${props.hideDuration ?? false}"
```
cjpillsbury, line 84: note: This was another case of attributes vs. props from `getProps` stepping on toes, where, even for a boolean, I have to handle this with nullish coalescing.

packages/mux-player/src/index.ts
```diff
@@ -674,6 +680,55 @@ class MuxPlayerElement extends VideoApiElement {
...
+      this.removeAttribute(PlayerAttributes.HIDE_DURATION);
+    } else {
+      this.setAttribute(PlayerAttributes.HIDE_DURATION, '');
+    }
+  }
+
+  /**
+   * Get the playback rates applied to the playback rate control.
+   */
+  get playbackRates() {
```
gkatsev, line 710: should we make playbackRates an AttributeTokenList like controlslist?

gkatsev (commented):
One thought

packages/mux-player/src/index.ts
cjpillsbury: iiinteresting. I'd lean towards no, at least for now, just because I think that would be clunky for the `set` side of the fence. With this, it's just "set an array". If we wanted to go that route, we'd probably want to at least figure out what that would look like and how it would relate to e.g. https://github.com/muxinc/media-chrome/blob/main/src/js/media-playback-rate-button.js#L33

cjpillsbury: This is particularly relevant for `mux-player-react` fwiw, though we could hypothetically implement the react prop and the custom element prop differently.

packages/mux-player/src/index.ts
```diff
@@ -674,6 +680,55 @@ class MuxPlayerElement extends VideoApiElement {
...
+    } else {
+      this.setAttribute(PlayerAttributes.DEFAULT_SHOW_REMAINING_TIME, '');
+    }
+  }
+
+  get hideDuration() {
+    return this.hasAttribute(PlayerAttributes.HIDE_DURATION);
+  }
+
+  set hideDuration(val: boolean | undefined) {
```
luwes, line 699: wouldn't this be better suited as a controlsList token? `noduration` 
I'd prefer this to keep the API surface as compact as possible 

this also reminded me of these web component guidelines
https://www.w3.org/2001/tag/doc/webcomponents-design-guidelines/#native-html-elements

> Don’t use custom attributes for styling, like bgcolor. Instead use [Custom Properties](https://drafts.csswg.org/css-variables) or [CSS Shadow Parts](https://drafts.csswg.org/css-shadow-parts/).

I know it's more 1 to 1 with how Media Chrome does it but the controlsList API seems like a better fit without needing to introduce another attribute and property.

packages/mux-player/src/media-chrome/time-display.ts
```diff
@@ -60,6 +62,23 @@ class MxpTimeDisplay extends HTMLElement {
 
     this.addEventListener('click', this.toggleTimeDisplay);
   }
+
+  attributeChangedCallback(attrName: string, _oldValue: string | null, newValue: string | null) {
+    if (attrName === 'hide-duration') {
+      if (typeof newValue === 'string') {
```
luwes, line 68: is there a particular reason this check is based on typeof? I thought we've been using `newValue == null` to see whether it was true or false for a boolean attribute

packages/mux-player/src/index.ts
```diff
@@ -674,6 +680,55 @@ class MuxPlayerElement extends VideoApiElement {
     this.setAttribute(PlayerAttributes.SECONDARY_COLOR, `${val}`);
   }
 
+  get defaultShowRemainingTime() {
+    return this.hasAttribute(PlayerAttributes.DEFAULT_SHOW_REMAINING_TIME);
+  }
+
+  set defaultShowRemainingTime(val: boolean | undefined) {
```
luwes, line 687: might feel as a stretch but here also I feel this could be part of the controlsList API.

it would require what @gkatsev brought up to have a part that is hidden by default `remainingtime` and the controlsList allowlist could make it show. e.g. `controlslist="noduration remainingtime"`

in Media Chrome these should also be able to be shown / hidden by a CSS var or part I think how they explain in the web component guidelines

packages/mux-player/src/media-chrome/time-display.ts
cjpillsbury: nope. Hadn't noticed we'd settled on a soft rule/convention for attributeChangedCallback, but I'll change it for consistency.

packages/mux-player/src/index.ts
cjpillsbury: I'm down for that idea. I'll go ahead and refactor, but just want to mention as food for thought that this will be the one controlslist element without a corresponding part. (not a big deal in my opinion)

packages/mux-player/src/index.ts
cjpillsbury: I'd lean away from using this one with the controls list, not just because it's different from the others with the block vs. allow syntax, but also because clicking on the element toggles between the two display versions.

packages/mux-player/src/template.ts
```diff
@@ -80,6 +83,11 @@ export const content = (props: MuxTemplateProps) => html`
     default-hidden-captions="${props.defaultHiddenCaptions}"
     forward-seek-offset="${props.forwardSeekOffset}"
     backward-seek-offset="${props.backwardSeekOffset}"
+    playbackrates="${props.playbackRates ?? false}"
+    hide-duration="${[...props.controlsList.values()].some(
+      (key) => key.search(/no(top|bottom|center)?duration/) === 0
```
luwes, line 88: nice 😎 

`.values()` can be left off if you like. `AttributeTokenList`'s iterator returns that as well.

nit (non-blocking) to be even more strict w/ start/end  could use a `/^no(top|bottom|center)?duration$/.test(key)`

cjpillsbury: cool might as well make those changes since they're easy and are good improvements.

gkatsev (approved):

PR #339, opened 2022-08-24
ci: fix upgrading canary deps

luwes:
Fixes the upgrading of the dependencies after a new canary was released of a package.

### todo

- [x] fix build number
potential fix by @gkatsev 
```bash
jq '. - map(select(contains("alpha") or contains("beta"))) | map(split(".")[3] | split("-")[0] | tonumber) | sort | last'
```

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -44,16 +44,30 @@ function release {
   npm --force --allow-same-version version $VERSION -m "chore(release): %s"
   git push --follow-tags
   npx conventional-github-releaser -p angular
+
+  echo "Beginning release $PKG_NAME@$VERSION"
   npm publish --access public
+
+  echo "Release dist for $PKG_NAME@$VERSION"
+  ls dist
   # `npm view name@x.x.x version` will return an empty string if not available
   until [ "$(npm view $PKG_NAME@$VERSION version)" != "" ];
   do
     echo "Waiting for publish to complete"
     sleep 3
   done
+
   # update all workspaces from the workspace root (../..) with the new version
-  # make sure publish.sh is called in topological order, `lerna run` does this
-  npx lerna-update-wizard --non-interactive --lazy --dependency $PKG_NAME@$VERSION ../..
+  # make sure publish.sh is called in topological order, `lerna ls --toposort` does this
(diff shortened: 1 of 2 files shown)


cjpillsbury (commented):
question: For testing this locally without having published artifacts, I was replacing `npm publish` with `npm pack` followed by `npm run prepublishOnly` and a few other hacks. Have you been able to test these changes locally (even if not exhaustively)?

cjpillsbury:
question: While investigating the issue, I noticed we're using both `yarn` and `npm` here. Was there a reason we didn't just use `yarn` equivalents for all of the `npm` commands? In particular, I noticed the release generated `package-lock.json` files, suggesting that our `yarn.lock` dependency versions aren't being respected in this process. Does that sound plausible to you?

luwes:
1. I only focused on testing upgrading the dependencies in each package because that's I think where it got stuck
  didn't try the actual publish flow.

2. iirc npm publish was required for the publish in the CI to work, why we added `--npm-client=npm` in the `publish-release` command
I assumed it didn't create a package-lock.json, I'll test this out. shouldn't happen.

cjpillsbury (approved):
🚢 🇮🇹

scripts/publish.sh
```diff
@@ -44,45 +44,87 @@ function release {
...
   PRE_VERSION=$(npx semver ${LAST_VERSION:-$PKG_VERSION} -i prerelease --preid canary)
   VERSION=$PRE_VERSION-$(git rev-parse --short HEAD)
-  echo "Beginning canary release for $PKG_NAME@$VERSION"
+
+  echo "Beginning release $PKG_NAME@$VERSION"
   npm --no-git-tag-version version $VERSION
   npm publish --tag canary --access public
-  echo "canary release dist used for $PKG_NAME@$VERSION"
+
+  echo "Release dist for $PKG_NAME@$VERSION"
```
gkatsev, line 107: can we make it clear that this is a canary release here?

luwes: it's in the version itself ;) 
Beginning release @mux/playback-core@0.9.1-canary.11-f46ae2d

PR #349, opened 2022-08-26
feat: use CSS vars to toggle off elements

luwes:
This change adds the bulk of the refactor so the controls are toggled with CSS vars instead of the `controlsList` property.

IMO this feels a lot better as an API.

```html
<mux-player style="--center-play-button: none; --top-controls: none"></mux-player>
```

The manual adding of parts in the MediaTheme is replaced by a little script that adds parts automatically based on the tag names of the Media Chrome elements. This to prevent putting this extra work on the theme developer.

We can consider adding that to MediaTheme but I'd lean to not doing that to avoid surprises. 
CSS parts will also not be required once MediaChrome has the correct CSS vars for display.

- [x] Todo is fixing `hide-duration` now with CSS vars.

### DIFF
--- a/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
@@ -2,7 +2,7 @@
 import Link from "next/link";
 import Script from 'next/script';
 import MuxPlayer from "@mux/mux-player-react";
-import { useRef, useState, useEffect, useCallback } from "react";
+import { useRef, useState } from "react";
 import mediaAssetsJSON from "@mux/assets/media-assets.json";
 
 const INITIAL_PRIMARY_COLOR = undefined;
@@ -16,7 +16,6 @@ const INITIAL_NOHOTKEYS = false;
 const INITIAL_DEFAULT_SHOW_REMAINING_TIME = true;
 const INITIAL_PLAYBACK_RATES = [0.25, 0.5, 1, 1.5, 2, 3];
 const INITIAL_ENV_KEY = "5e67cqdt7hgc9vkla7p0qch7q";
-const INITIAL_CONTROLS_CSS_VARS = [];
 const INITIAL_SELECTED_CSS_VARS = {};
 
 const toMetadataFromMediaAsset = (mediaAsset: typeof mediaAssetsJSON[0], mediaAssets: typeof mediaAssetsJSON) => {
@@ -63,12 +62,7 @@ function MuxPlayerPage() {
   const [autoplay, setAutoplay] = useState<"muted" | boolean>(INITIAL_AUTOPLAY);
   const [primaryColor, setPrimaryColor] = useState<string|undefined>(INITIAL_PRIMARY_COLOR);
   const [secondaryColor, setSecondaryColor] = useState<string|undefined>(INITIAL_SECONDARY_COLOR);
-  const [controlsCssVars, setControlsCssVars] = useState(INITIAL_CONTROLS_CSS_VARS);
   const [selectedCssVars, setSelectedCssVars] = useState(INITIAL_SELECTED_CSS_VARS);
-
-  useMutationObserver(mediaElRef.current?.theme.shadowRoot, useCallback(() => {
-    setControlsCssVars(getAvailableControlsCssVars(mediaElRef.current?.theme?.shadowRoot));
-  }, [setControlsCssVars]));
 
   return (
     <div>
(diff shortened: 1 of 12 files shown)


examples/nextjs-with-typescript/pages/MuxPlayer.tsx
```diff
@@ -62,7 +63,12 @@ function MuxPlayerPage() {
   const [autoplay, setAutoplay] = useState<"muted" | boolean>(INITIAL_AUTOPLAY);
   const [primaryColor, setPrimaryColor] = useState<string|undefined>(INITIAL_PRIMARY_COLOR);
   const [secondaryColor, setSecondaryColor] = useState<string|undefined>(INITIAL_SECONDARY_COLOR);
-  const [controlslist, setControlslist] = useState(INITIAL_CONTROLS_LIST);
+  const [controlsCssVars, setControlsCssVars] = useState(INITIAL_CONTROLS_CSS_VARS);
+  const [selectedCssVars, setSelectedCssVars] = useState(INITIAL_SELECTED_CSS_VARS);
+
+  useMutationObserver(mediaElRef.current?.theme.shadowRoot, useCallback(() => {
```
cjpillsbury, line 69: question(blocking): These examples are supposed to be for reference for folks using Mux Player. Do we expect other folks will ever want this list of CSS vars? If yes, we should really expose them in a way that doesn't require drilling into the theme.

examples/nextjs-with-typescript/pages/MuxPlayer.tsx
```diff
@@ -235,4 +241,41 @@ function MuxPlayerPage() {
...
+    cssVars[`--${matches[2]}-${matches[3]}`] = 'none';
+    cssVars[`--${section}-${matches[2]}-${matches[3]}`] = 'none';
+  }
+  return Object.keys(cssVars);
+}
+
+const DEFAULT_OPTIONS = {
+  config: { attributes: true, childList: true, subtree: true },
+};
+function useMutationObserver(targetEl, cb, options = DEFAULT_OPTIONS) {
```
cjpillsbury, line 266: issue(blocker): Creating a mutation observer hook to manage this introduces too much complexity for an example and shouldn't be necessary. If it is, I suspect we need to revisit why and resolve that.

cjpillsbury (changes requested):
Did a quick once over. At least on a quick glance, the code changes make sense. Since this refactor was supposed to improve devex, I have concerns about the examples changes that I believe we should address. Also, let's make sure we don't merge this until we have adequate documentation.

examples/nextjs-with-typescript/pages/MuxPlayer.tsx
luwes: 👍 good point, does look a bit messy. I'll make it a static list.

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -101,6 +78,21 @@ export default class MediaThemeMux extends MediaTheme {
       `,
       this.shadowRoot as Node
     );
+
+    // Possibly add this functionality to the base class MediaTheme.
+    this.shadowRoot?.querySelectorAll('*').forEach((element) => {
```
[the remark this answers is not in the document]
luwes: 1. good point, just to be safe because when changing permutations some elements will be replaced and need the `part` attribute again. the goal is to have the theme developer not worry about this logic that can be automated.

2. currently it's still needed but I believe once we add these `top|center|bottom` parts in Media Chrome it's possible to remove the section parts for each component and just target the part and put the right CSS var in it.

packages/mux-player/README.md
```diff
@@ -117,6 +117,70 @@ Enable the [Google Cast](https://developers.google.com/cast) button in the contr
 <script defer src="https://www.gstatic.com/cv/js/sender/v1/cast_sender.js?loadCastFramework=1"></script>
 ```
 
+### Hiding controls
+
+By default, Mux Player will show all the controls available:
```
gkatsev, line 122: Might worth mentioning that depending on display size and stream type, not everything may be shown.

packages/mux-player/src/media-theme-mux/styles.css
```diff
@@ -114,15 +114,15 @@ media-control-bar :is([role='button'], [role='switch'], button) {
   padding: 4.4% 3.2%;
 }
 
-.mxp-spacer {
+.spacer {
```
gkatsev, line 117: At first, I thought this is too generic, but it should be fine because it's only inside the shadow DOM, right?

gkatsev (approved):

packages/mux-player/src/media-theme-mux/styles.css
luwes: indeed, at least some benefits thank god :P

packages/mux-player/README.md
luwes: good call, will add something in those lines

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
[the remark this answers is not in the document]
cjpillsbury: cool makes sense to me. Something to keep an eye on for improving at some point.

examples/nextjs-with-typescript/pages/MuxPlayer.tsx
```diff
@@ -209,20 +209,52 @@ function MuxPlayerPage() {
...
+          <label htmlFor="controlsvars-control">Hide controls CSS vars </label>
           <select
-            id="controlslist-control"
+            id="controlsvars-control"
             multiple
-            onChange={(event) => setControlslist(
-              Array.from(event.target.selectedOptions)
-                .map(({ value }) => value).join(' ')
+            onChange={(event) => setSelectedCssVars(
+              Object.fromEntries(Array.from(event.target.selectedOptions)
```
cjpillsbury, line 217: idea(non-blocking): cool. Fyi `useState` supports functional updates in case you wanted to simplify this: https://reactjs.org/docs/hooks-reference.html#functional-updates

cjpillsbury (approved):
😎

PR #350, opened 2022-08-26
Update publish script, fix releases

gkatsev:
Probably should squash-merge this PR rather than rebase-merge.


scripts/publish.sh
```diff
@@ -37,20 +58,51 @@ function processCommandLineArgs {
 
 function release {
   BUMP=$(npx -p conventional-changelog-angular -p conventional-recommended-bump -c 'conventional-recommended-bump -p angular')
-  VERSION=$(yarn version --no-git-tag-version --new-version ${1:-$BUMP})
+  # jq to get version from yarn https://jqplay.org/s/LqIS_qy2MBx
+  VERSION=$(yarn version --no-git-tag-version --new-version ${release_type:-$BUMP} --json | jq -r 'select(.data | startswith("New version")).data | split(": ")[1]')
```
gkatsev, line 62: `yarn version` has different output from `npm version`, this parses their output via JSON to get the proper value. Needs jq -r for "raw output" to get the raw value, otherwise the yarn version below fails.

scripts/publish.sh
```diff
@@ -60,13 +112,16 @@ function release {
   for name in ${DEPENDANT_PKGS}; do
     scope+="--scope $name "
   done
-  npx lerna exec $scope -- npm pkg set dependencies.$PKG_NAME=$VERSION
+
+  if "$dry_run"; then
+    echo "Running the following command" npx lerna exec $scope -- npm pkg set dependencies.$PKG_NAME=$VERSION > /dev/null
+  fi
+  npx lerna exec $scope -- npm pkg set dependencies.$PKG_NAME=$VERSION > /dev/null
```
gkatsev, line 119: pipe `npm pkg set` to `/dev/null` so that it won't print out `{}`

gkatsev:
Unfortunately, adding dry_run makes the code a bit more annoying to read.

scripts/publish.sh
```diff
@@ -37,20 +58,51 @@ function processCommandLineArgs {
...
-  git push --follow-tags
-  npx conventional-github-releaser -p angular
 
-  echo "Beginning release $PKG_NAME@$VERSION"
-  yarn publish --access public --non-interactive
+  if "$dry_run"; then
+    echo
+    echo "In non-dry-run, will run the following commands"
+    echo "git add CHANGELOG.md"
+    echo git commit -m "docs(CHANGELOG): $VERSION"
```
luwes, line 70: shouldn't this be be in quotes `git commit ...`?

gkatsev: It doesn't *need* it, but I think worth adding for consistency.

luwes (approved):
LGTM 👍

PR #353, opened 2022-08-29
feat: add --controls-backdrop-color CSS var to allow changing the backdrop color

gkatsev:
~"controls shade color" is sort of a placeholder to kick off a discussion for a better name.~

You can try it out on the MuxPlayer nextjs test page.

### DIFF
--- a/packages/mux-player/README.md
+++ b/packages/mux-player/README.md
@@ -198,6 +198,18 @@ Instead, some components expose **parts** that can be targeted with the [CSS par
 <mux-player playback-id="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"></mux-player>
 ```
 
+### Controls Backdrop Color
+
+We expose a CSS variable to set the controls backdrop color.
+
+```css
+mux-player {
+  --controls-backdrop-color: rgb(0 0 0 / 0%);
+}
+```
+
+Turning this off completely as implications on the accessibility of the controls as they may not meet [the contrast ratio requirements for WCAG 2.1](https://www.w3.org/TR/WCAG/#contrast-minimum) without it.
+
 Supported **parts**:
 `seek-live`, `layer`, `media-layer`, `poster-layer`, `vertical-layer`, `centered-layer`, `gesture-layer`,
 `top`, `center`, `bottom`, `play`, `button`, `seek-backward`, `seek-forward`, `mute`,
--- a/packages/mux-player/README.md
+++ b/packages/mux-player/README.md
@@ -208,7 +208,7 @@ mux-player {
 }
 ```
 
-Turning this off completely as implications on the accessibility of the controls as they may not meet [the contrast ratio requirements for WCAG 2.1](https://www.w3.org/TR/WCAG/#contrast-minimum) without it.
+Turning this off completely has implications on the accessibility of the controls as they may not meet [the contrast ratio requirements for WCAG 2.1](https://www.w3.org/TR/WCAG/#contrast-minimum) without it.
 
 Supported **parts**:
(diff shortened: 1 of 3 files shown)


luwes:
I've brainstormed about this before, what do y'all think of `backdrop`?
see https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop

gkatsev:
backdrop might be good. should it be `--backdrop-color` or `--controls-backdrop-color`? Or maybe `--media-backdrop-color`?

luwes:
`--controls-backdrop-color` seems more accurate to me. 


`--media-backdrop-color` could mean for transparent video the color behind the video element
https://ataylor32.github.io/demo-html5-transparent-video/

gkatsev:
New design doesn't have a full player backdrop, so, going to close this, and we can revisit when we start targeting that design update.

packages/mux-player/README.md
```diff
@@ -198,6 +198,18 @@ Instead, some components expose **parts** that can be targeted with the [CSS par
...
+
+We expose a CSS variable to set the controls backdrop color.
+
+```css
+mux-player {
+  --controls-backdrop-color: rgb(0 0 0 / 0%);
+}
+```
+
+Turning this off completely as implications on the accessibility of the controls as they may not meet [the contrast ratio requirements for WCAG 2.1](https://www.w3.org/TR/WCAG/#contrast-minimum) without it.
```
luwes, line 211:
```suggestion
Turning this off completely has implications on the accessibility of the controls as they may not meet [the contrast ratio requirements for WCAG 2.1](https://www.w3.org/TR/WCAG/#contrast-minimum) without it.
```

luwes:
was wondering if we should check if it's worth to open up adding the background prop.
would be nice for gradients and such.
https://github.com/muxinc/elements/pull/389

the only thing that is not ideal is that it had to be a layer lower than vertical-layer so on gestures-layer...

packages/mux-player/README.md
```diff
@@ -198,6 +198,18 @@ Instead, some components expose **parts** that can be targeted with the [CSS par
...
+
+We expose a CSS variable to set the controls backdrop color.
+
+```css
+mux-player {
+  --controls-backdrop-color: rgb(0 0 0 / 0%);
+}
+```
+
+Turning this off completely has implications on the accessibility of the controls as they may not meet [the contrast ratio requirements for WCAG 2.1](https://www.w3.org/TR/WCAG/#contrast-minimum) without it.
```
cjpillsbury, line 211: suggestion(non-blocking)
> Turning this off completely or adjusting it significantly can effect accessibility considerations and compliance for [the contrast ratio requirements ...

(or similar)
Also, in my meeting w/the a11y audit company, they let me know that https://www.w3.org/TR/WCAG22/ just happened (which they tested against, iirc)

cjpillsbury (approved):
LGTMe. Will explore a followup on the example app in my refactor effort, per conversation.

cjpillsbury:
> the only thing that is not ideal is that it had to be a layer lower than vertical-layer so on gestures-layer...

@luwes Won't this be incompatible with your recent change to rely on `display: none` for disabling gestures?

packages/mux-player/README.md
gkatsev: Looks like 2.2 is still in CR and not a TR just yet, but I do think it'll be officially official 🔜

luwes:
@cjpillsbury no the gestures-layer itself which itself is a slot element now will not be display none. only the elements within

luwes (approved):
👍 we can revisit gradients later if there is interest

PR #355, opened 2022-08-30
fix: polyfills mutating global

### LINKED ISSUE #334: Bug: Document getting set in Node "@mux/mux-player": "0.1.0-beta.26",

`"@mux/mux-player": "0.1.0-beta.26",`

Node

macOS

import '@mux/mux-player'

### Describe the bug

We're seeing that after importing @mux/mux-player document is no longer undefined within Node which in turn breaks several 3rd party packages with one-such being `prismjs` who uses the existence of document as a check for node. I 

## Steps To Reproduce

1. SSR'd site NextJS / Svelte Kit ect
2. Install and import @mux/mux-player 
3. Observe document no longer being undefined on Node.

I attempted to find where in mux player this was getting set, but I was unable to find it.

luwes:
fix #334 

2 special cases:

- castable-video doesn't import the polyfills to not have to make it a full build package
  checks `globalThis.HTMLVideoElement && globalThis.customElements`
- `DocumentFragment` had to be set on globalThis because a 3rd party pkg depends on it, @github/template-parts


packages/mux-audio/package.json
```diff
@@ -47,11 +47,11 @@
     "publish-release": "../../scripts/publish.sh"
   },
   "dependencies": {
-    "@mux/playback-core": "0.9.0"
+    "@mux/playback-core": "0.9.0",
+    "@mux/polyfills": "0.1.0"
```
cjpillsbury, line 51: issue(blocking): Can't be a dependency. polyfills isn't published 💔

luwes: - made a atmux/polyfills -> shared-polyfills so we can still use that atmux scope.
- since it's bundled in no need to create builds/bundles for shared-polyfills

gkatsev (approved):

cjpillsbury (approved):

PR #356, opened 2022-08-30
Feat/title support

cjpillsbury:
Adds title UI support, including css vars/parts, for mux-player and mux-player-react.


packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -85,7 +88,6 @@ export default class MediaThemeMux extends MediaTheme {
           exportparts="layer, media-layer, poster-layer, vertical-layer, centered-layer"
         >
           <slot name="media" slot="media"></slot>
-          <media-poster-image slot="poster" src="${props.poster}"></media-poster-image>
```
luwes, line 88: blocking: shouldn't be removed right?

luwes (changes requested):

luwes:
blocking: default font-size should be 16px from designs

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
cjpillsbury: oh man good catch. Not sure why that happened with the rebase merge.

luwes:
non-blocking: to me here the addition of the pipe char is less good looking I think. mostly because the spacing is asymmetrical 

SCR-20220830-q05

luwes (approved):
LGTM 👍

PR #365, opened 2022-09-06
fix: update to hls.js@1.2.2

gkatsev:


luwes (dismissed):
LGTM 👍

luwes (approved):

packages/mux-player/src/index.ts
```diff
@@ -593,12 +593,12 @@ class MuxPlayerElement extends VideoApiElement {
   /**
    * @deprecated please use ._hls instead
    */
-  get hls() {
+  get hls(): PlaybackEngine | undefined {
```
cjpillsbury, line 596: thought(non-blocking): this should be inferrable if `this._hls` is typed as expected, but should not be a blocker.

cjpillsbury (approved):

packages/mux-player/src/index.ts
gkatsev: That's what I thought as well, but when I went to do this update, tsc was complaining that these weren't typed.

PR #366, opened 2022-09-07
tools: upgrade esbuild, use new --watch=forever

luwes:
instead of open-process hack evanw/esbuild#1885


gkatsev (approved):
Workings locally for me 👍🏻

PR #368, opened 2022-09-07
fix: startTime on iOS when preload=auto

luwes:
Fixes the issue where startTime was not being set on iOS when `preload=auto`, the default in some cases.

After setting currentTime 2 issues happened:
- setting currentTime did not actually work, when hitting play it would start at 0s
- the `seeking` event was fired but not the `seeked` event and subsequently no `timeupdate` events were fired causing the time range and time display to show wrong info

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -164,7 +164,7 @@ export const getStreamTypeConfig = (streamType?: ValueOf<StreamTypes>) => {
   return {};
 };
 
-let muxMediaState: WeakMap<HTMLMediaElement, { error?: MediaError }> = new WeakMap();
+let muxMediaState: WeakMap<HTMLMediaElement, Partial<MuxMediaProps> & { error?: MediaError }> = new WeakMap();
 
 export const getError = (mediaEl: HTMLMediaElement) => {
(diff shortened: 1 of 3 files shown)


packages/playback-core/src/index.ts
```diff
@@ -438,6 +435,27 @@ export const loadMedia = (
...
+function seekInSeekableRange(event: Event) {
+  const mediaEl = event.target as HTMLMediaElement;
+  const startTime = muxMediaState.get(mediaEl)?.startTime;
+  if (!startTime) return;
+
+  if (inSeekableRange(mediaEl.seekable, mediaEl.duration, startTime)) {
+    // Setting preload to `none` from `auto` was required on iOS to fix a bug
+    // that caused no `timeupdate` events to fire after seeking ¯\_(ツ)_/¯
+    const wasAuto = mediaEl.preload === 'auto';
+    if (wasAuto) {
```
cjpillsbury, line 447: question(blocking): Just confirming, this issue does _**not**_ happen when `preload = 'metadata'`, correct?

luwes: yes, double checked that. it's working correctly with `preload=metadata`

PR #371, opened 2022-09-08
fix: template parts issue

luwes:
fixes the issue where the title would not render again after removal.

### DIFF
--- a/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
+++ b/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
@@ -274,7 +274,7 @@ const TitleDisplay = ({ title }: ComponentProps) => html`
 
 // prettier-ignore
 export const AudioVodChrome = (props: ThemeMuxTemplateProps) => html`
-  <media-control-bar>${props.title ? TitleDisplay(props) : ''}</media-control-bar>
+  <media-control-bar>${props.title ? TitleDisplay(props) : html``}</media-control-bar>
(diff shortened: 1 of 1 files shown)


packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -274,7 +274,7 @@ const TitleDisplay = ({ title }: ComponentProps) => html`
 
 // prettier-ignore
 export const AudioVodChrome = (props: ThemeMuxTemplateProps) => html`
-  <media-control-bar>${props.title ? TitleDisplay(props) : ''}</media-control-bar>
+  <media-control-bar>${props.title ? TitleDisplay(props) : html``}</media-control-bar>
```
cjpillsbury, line 277: question(non-blocking): Is this a shortcoming of our current template implementation? feels clunky (already feels clunky to have to put anything there), but I don't want this to block the fix.

cjpillsbury (approved):
Approved with one callout.

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
luwes: yes I hear you, I ran into this w/ the `<track ` element in template.ts. 

I thought I tried something at the time and it wasn't that easy but yes agree it would be nicer if a `false` would or empty string would result in an empty TemplateResult. happy to check it out after GA.

PR #372, opened 2022-09-08
feat: always prefer MSE on android

gkatsev:
Follow-up to #325. There was a second place where we were deciding default MSE preference, which didn't get updated.


luwes (approved):
👍

PR #373, opened 2022-09-08
chore: add workflow to auto-add issues to project

gkatsev:
This PR adds a GitHub Action that adds new issues and PRs to the project using the https://github.com/actions/add-to-project action.

Unfortunately, this requires a Personal Access Token with repo and project access. I've added one to this repo, which will expire in 30 days, but we should consider how we want that generated longer term and as we roll this action out to other repos.


luwes (approved):
looks good!

PR #382, opened 2022-09-12
Poc/memory leaks demo

cjpillsbury:
This PR includes the code changes to (🤞) handle any dangling references in playback-core and updates some of our examples for testing/validation. Some notes:
* There still appears to be another memory leak that doesn't happen in vanilla html but does happen in React (even when directly using our web components in react)
* ~The playback-core changes currently include disabling hls.js workers to make memory profiling easier.~ (removed from the codebase. Still present in some of the one-off example pages)
* A separate "official" PR should be issued, likely without any of the example changes, if everything looks good here.

Steps to test reproducing the memory leak:
1. load one of the updated/created pages on this PR, or any page that has a list of different playback ids (e.g. generated from our media-assets)
  1. These will be on the `elements-demo-nextjs` or `elements-demo-vanilla` pages deployed from vercel if you are not testing locally
3. Use dev tools in the browser of your choice to do memory profiling.
4. Switch playback ids several times.
5. Take a heap snapshot in the memory profiler
6. Confirm/disconfirm that the number of hls.js instances has grown with the number of playback ids loaded.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -4,7 +4,7 @@ import Hls, { HlsConfig } from 'hls.js';
 import { AutoplayTypes, setupAutoplay } from './autoplay';
 import { MediaError } from './errors';
 import { setupTracks, addTextTrack, removeTextTrack } from './tracks';
-import { isKeyOf, inSeekableRange } from './util';
+import { isKeyOf, inSeekableRange, addEventListenerWithTeardown } from './util';
 import type { Autoplay, UpdateAutoplay } from './autoplay';
 
 export type ValueOf<T> = T[keyof T];
@@ -182,9 +182,11 @@ export const teardown = (mediaEl?: HTMLMediaElement | null, hls?: Pick<Hls, 'det
   }
   if (mediaEl?.mux && !mediaEl.mux.deleted) {
     mediaEl.mux.destroy();
-    mediaEl.mux;
+    delete mediaEl.mux;
   }
   if (mediaEl) {
+    mediaEl.removeAttribute('src');
+    mediaEl.load();
     mediaEl.removeEventListener('error', handleNativeError);
     mediaEl.removeEventListener('error', handleInternalError);
     mediaEl.removeEventListener('durationchange', seekInSeekableRange);
@@ -397,7 +399,7 @@ export const loadMedia = (
     switch (mediaEl.preload) {
       case 'none':
         // when preload is none, load the source on first play
-        mediaEl.addEventListener('play', () => hls.loadSource(src), { once: true });
+        addEventListenerWithTeardown(mediaEl, 'play', () => hls.loadSource(src), { once: true });
         break;
 
       case 'metadata':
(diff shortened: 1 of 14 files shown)


packages/playback-core/src/index.ts
```diff
@@ -397,7 +399,16 @@ export const loadMedia = (
     switch (mediaEl.preload) {
       case 'none':
         // when preload is none, load the source on first play
-        mediaEl.addEventListener('play', () => hls.loadSource(src), { once: true });
+        const loadSrcOnPlay = () => hls.loadSource(src);
+        mediaEl.addEventListener('play', loadSrcOnPlay, { once: true });
+
+        mediaEl.addEventListener(
```
cjpillsbury, line 405: given that we rely on this pattern throughout to ensure proper reference cleanup, we may want to just make a function to do this for us. defer to others on priority.

packages/playback-core/src/index.ts
```diff
@@ -182,9 +182,11 @@ export const teardown = (mediaEl?: HTMLMediaElement | null, hls?: Pick<Hls, 'det
   }
   if (mediaEl?.mux && !mediaEl.mux.deleted) {
     mediaEl.mux.destroy();
-    mediaEl.mux;
+    delete mediaEl.mux;
   }
   if (mediaEl) {
+    mediaEl.removeAttribute('src');
```
cjpillsbury, line 188: Per @luwes's proposal to better ensure `emptied` is fired. 🔥

packages/playback-core/src/tracks.ts
```diff
@@ -69,7 +69,10 @@ export function setupTracks(
     }
   });
 
-  hls.on(Hls.Events.DESTROYING, () => {
+  mediaEl.addEventListener('emptied', () => {
+    mediaEl.textTracks.removeEventListener('change', changeHandler);
+  });
+  hls.once(Hls.Events.DESTROYING, () => {
```
cjpillsbury, line 75: this likely is no longer necessary

packages/playback-core/src/index.ts
gkatsev: yeah, having a "onEventPerSource", or something like that, function would be nice.

packages/playback-core/src/index.ts
```diff
@@ -182,9 +182,11 @@ export const teardown = (mediaEl?: HTMLMediaElement | null, hls?: Pick<Hls, 'det
   }
   if (mediaEl?.mux && !mediaEl.mux.deleted) {
     mediaEl.mux.destroy();
-    mediaEl.mux;
+    delete mediaEl.mux;
```
gkatsev, line 185: Should mux-embed do this for us?

cjpillsbury: Looks like it already does and I just have trust issues. 😄

cjpillsbury: Actually on closer inspection this only happens when a new "monitor" is setup. I think we should leave this as is for now and follow up w/exploring "deleting on destroy" in mux-embed (not sure if there are Good Reasons™️ for not doing that).

packages/playback-core/src/tracks.ts
luwes: yup can be removed

PR #383, opened 2022-09-12
fix: audio and nohotkeys props not being removed when set to false

gkatsev:
The properties were being removed, but then the functions weren't being exited early, so, they would get re-added.


luwes (approved):

cjpillsbury (approved):
👍

PR #384, opened 2022-09-13
Add exportparts to mux-player so we can style nested video

mikker:
This is needed to allow adding CSS to a nested web component via the `part()` syntax.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts

Tested and it works.

```html
exportparts="video: video"
```

```css
mux-player::part(video) {
  border: 3px solid red;
}
```


luwes:
closing in favor of https://github.com/muxinc/elements/pull/392

Issue #308, opened 2022-07-21
Bug: Memory leak on react components


lpalombo:
### Checklist

- [x] Verified whether your issue exists in the latest version Mux Element(s) release: `npm install @mux/[element_name]` or `yarn @mux/[element_name]`.
  - (NOTE: If the issue is not occurring in the latest version, we may not resolve it.)
- [x] Confirmed that this issue hasn't [already been filed](https://github.com/muxinc/elements/issues?q=is%3Aopen+is%3Aissue+label%3Abug).

### Which Mux Elements/Packages has a bug?

- mux-video-react
- mux-player-react

### Describe the bug

Hls objects are not released from javascript heap when either of the components are unmounted.

## Steps To Reproduce

1. Mount component with either `<MuxVideo />` or `<MuxPlayer />`
2. Unmount component (in my case navigating to different route via Next.js)
3. Notice that among other things Hls objects remain in javascript heap with references to detached HTMLVideoElements

![image](https://user-images.githubusercontent.com/9061614/180328505-2806ff0e-31f4-4a6b-845c-660d014187bb.png)

I assume this is simply because the `teardown` function is not passed in from `@mux/playback-core` into the components and called as a return function in the first `useEffect` hook. I would run my own cleanup function on my end, but the component doesn't provide an API for cleanup either, and calling `video.mux.destroy()` doesn't appear to cleanup Hls.

cjpillsbury:
@lpalombo this is great feedback (sad that it's happening, but happy you're helping us identify issues). We'll be prioritizing identifying root cause and fixing ASAP!

cjpillsbury:
Hey @lpalombo just an update: It looks like there may be several compounding issues here. There's a PR [here](https://github.com/muxinc/elements/pull/385) to resolve some "dangling references" issues under the hood, but we're still finding some issues that are happening on the React side, even when directly using our web components directly in a React app. @gkatsev found some issues on the React repo that may be related. I've gone ahead and also pushed up a draft PR with some custom test pages to investigate different permutations for hunting down different cases, found [here](https://github.com/muxinc/elements/pull/382), which I'll be updating periodically. Feel free to ping us if you have any questions or wanted to help out.

luwes:
Thanks for the report @lpalombo! This should be fixed in the latest release
https://github.com/muxinc/elements/releases/tag/%40mux%2Fmux-player%401.0.0-beta.0

Let us know if you see different.

luwes:
[...]

PR #385, opened 2022-09-13
fix(playback-core): Make sure event handlers and other references are…

cjpillsbury:
… cleaned up to avoid memory leaks.

This should resolve any memory leaks that are a direct result of dangling references in playback-core.

Relates to #308

### DIFF
--- a/packages/playback-core/src/autoplay.ts
+++ b/packages/playback-core/src/autoplay.ts
@@ -1,4 +1,5 @@
 import Hls from 'hls.js';
+import { addEventListenerWithTeardown } from './util';
 
 type PlaybackEngine = Hls;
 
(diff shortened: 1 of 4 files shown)


packages/playback-core/src/autoplay.ts
```diff
@@ -43,23 +44,28 @@ export const setupAutoplay = (
...
-        hasPlayed = true;
-      },
-      { once: true }
-    );
+
+    if (!hasPlayed) {
+      addEventListenerWithTeardown(
+        mediaEl,
+        'playing',
+        () => () => {
```
gkatsev, line 52: why does this have an extra arrow function wrapper? (Also, the other occurrences in this file)

cjpillsbury: So the reason for this is I didn't want to only do half of the "Move fast, break things" mantra.

luwes (approved):
LGTM :shipit:

gkatsev (approved):

PR #389, opened 2022-09-14
feat: add Apple-ish gradient

luwes:
just opening to show what we could enable users to do with having a CSS `background` prop.


luwes:
close until there is interest in this feature

PR #392, opened 2022-09-15
fix: Add exportparts to mux-player so we can style nested video

luwes:
Original at https://github.com/muxinc/elements/pull/384 by @mikker

> This is needed to allow adding CSS to a nested web component via the `part()` syntax.
> 
> See [developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts)
> 
> Tested and it works.
> 
> ```
> exportparts="video: video"
> ```
> 
> ```css
> mux-player::part(video) {
>   border: 3px solid red;
> }
> ```

### DIFF
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -45,7 +45,6 @@ export const content = (props: MuxTemplateProps) => html`
     playbackrates="${props.playbackRates ?? false}"
     default-show-remaining-time="${props.defaultShowRemainingTime ?? false}"
     title="${props.title ?? false}"
-    exportparts="top, center, bottom, layer, media-layer, poster-layer, vertical-layer, centered-layer, gesture-layer, seek-live, play, button, seek-backward, seek-forward, mute, captions, airplay, pip, fullscreen, cast, playback-rate, volume, range, time, display"
     hotkeys="${props.hotKeys || false}"
     poster="${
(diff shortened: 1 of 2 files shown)


packages/mux-player/src/template.ts
```diff
@@ -105,6 +105,7 @@ export const content = (props: MuxTemplateProps) => html`
           : false
       }"
       cast-stream-type="${isLive(props) ? 'live' : false}"
+      exportparts="video"
```
gkatsev, line 108: is the reason why it's only needed here because it's not in the shadow dom of mux-player?

luwes: yup, CSS parts need to be forwarded between multiple shadom doms

gkatsev (approved):
My question isn't blocking as I confirmed it's working.

packages/mux-player/src/template.ts
luwes: I wish we didn't have to add such big attributes for themes but yea 🤷‍♂️

mikker:
Glorious! Thank you. Love Mux ❤️🧡💛💚💙💜

luwes:
@mikker thank you for the contribution! ❤️

PR #393, opened 2022-09-16
Fix/react event callback types

cjpillsbury:
Also did some additional "under the hood" cleanup of types and vestigial code.

### DIFF
--- a/packages/mux-player-react/src/index.tsx
+++ b/packages/mux-player-react/src/index.tsx
@@ -214,12 +214,6 @@ const MuxPlayer = React.forwardRef<
   MuxPlayerRefAttributes,
   Omit<MuxPlayerProps, 'playerSoftwareVersion' | 'playerSoftwareName'>
 >((props, ref) => {
-  const {
-    /** @TODO Remove these once defaults are added to mux-player (CJP) */
-    forwardSeekOffset = 10,
-    backwardSeekOffset = 10,
-  } = props;
-
   const innerPlayerRef = useRef<MuxPlayerElement>(null);
   const playerRef = useCombinedRefs(innerPlayerRef, ref);
   const [remainingProps] = usePlayer(innerPlayerRef, props);
@@ -230,8 +224,6 @@ const MuxPlayer = React.forwardRef<
       ref={playerRef as typeof innerPlayerRef}
       playerSoftwareName={playerSoftwareName}
       playerSoftwareVersion={playerSoftwareVersion}
-      forwardSeekOffset={forwardSeekOffset}
-      backwardSeekOffset={backwardSeekOffset}
       {...remainingProps}
     />
   );
--- a/packages/mux-player-react/src/index.tsx
+++ b/packages/mux-player-react/src/index.tsx
@@ -13,6 +13,9 @@ import { getPlayerVersion } from './env';
 export { MediaError };
 
 type ValueOf<T> = T[keyof T];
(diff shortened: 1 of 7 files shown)


luwes (approved):
nice!

packages/mux-player-react/src/index.tsx
```diff
@@ -68,42 +73,42 @@ export type MuxPlayerProps = {
...
+  onPlaying?: GenericEventListener<MuxPlayerElementEventMap['playing']>;
+  onTimeUpdate?: GenericEventListener<MuxPlayerElementEventMap['timeupdate']>;
+  onPause?: GenericEventListener<MuxPlayerElementEventMap['pause']>;
+  onSeeking?: GenericEventListener<MuxPlayerElementEventMap['seeking']>;
+  onSeeked?: GenericEventListener<MuxPlayerElementEventMap['seeked']>;
+  onStalled?: GenericEventListener<MuxPlayerElementEventMap['stalled']>;
+  onSuspend?: GenericEventListener<MuxPlayerElementEventMap['suspend']>;
+  onEnded?: GenericEventListener<MuxPlayerElementEventMap['ended']>;
+  onError?: GenericEventListener<MuxPlayerElementEventMap['error']>;
+  // onPlayerReady?: EventListener;
```
spenweb, line 99: @luwes why was it decided to remove the `onPlayerReady` prop and corresponding functionality?

luwes: @spenweb so far there was no need for this callback. The MuxPlayer is an inline JS embed, once the JS constructor executes the API is ready to accept calls. There are all the native video events available like `loadstart`, `loadedmetadata`, etc. which should provide enough indication when the src is set and loaded etc.

Do you have a need for an onPlayerReady callback?

spenweb: Yes, I need to be able to change the volume after the player UI has initialized. I've been experiencing odd behavior where I am able to change the volume but the volume UI slider does not update. This leads me to believe that I need to wait for the player UI to initialize before I can set the volume.

Here is my current work-around hack using a useEffect:
```ts
// Control the Mux player's volume state
useEffect(() => {
  // volume is coming from a prop
  targetVolumeRef.current = volume

  let timeoutId: NodeJS.Timeout | null = null
  let cancel: boolean

  const enforcer = () => {
    if (
      !muxPlayerRef.current ||
      cancel ||
      targetVolumeRef.current === undefined
    ) {
      return
    }
    // `updateVolumeEquilibrium` just checks to see if the target volume has been achieved
    const isEquilibriumAchieved = updateVolumeEquilibrium()

    if (!isEquilibriumAchieved) {
      // If equilibrium has not been achieved, set Mux Player to match volume value
      muxPlayerRef.current.volume = targetVolumeRef.current

      // Recursively call this function until equilibrium is achieved using a timeout
      // Recursion may not be needed but is here out of paranoia (with experience from YouTube player)
      timeoutId = setTimeout(enforcer, 100)
    }
  }

  // TODO: Find reliable way to know when Mux player is ready
  // @see https://github.com/muxinc/elements/issues/676
  setTimeout(enforcer, 500)

  return () => {
    if (timeoutId) {
      clearTimeout(timeoutId)
    }
    cancel = true
  }
}, [volume, src, updateVolumeEquilibrium])
```

Not a big fan of having to rely on timeouts and recursion if not necessary. That's why having a callback when the Mux Player UI is ready would be fantastic!

PR #401, opened 2022-09-19
chore: update svelte-kit example to svelte-kit rc

decepulis:
SvelteKit went through some [major changes](https://github.com/sveltejs/kit/discussions/5748), recently. (Again.)
[Rumor on the street is](https://twitter.com/geoffrich_/status/1568264897905164289?s=21&t=HZWdeIZ1UgzodAn_kVzwMQ) that this is the last batch of said major changes. 
I thought I'd bring the repo up to date as long as I was in the neighborhood. More specifically...

## In SvelteKit...
I started a new SvelteKit project using their latest template and moved our files over to there. What did that practically do?
- updated dependencies
- updated all sorts of config files to the latest conventions
- renamed a few files and variables to fit with the latest routing and naming conventions

Though from a ux experience, nothing _should_ have changed

## Elsewhere...
Since SvelteKit now explicitly requires Vite 3, while our Vue example is on Vite 2, I added a `nohoist` to the root-level `package.json` to isolate the two.


package.json
```diff
@@ -24,6 +24,10 @@
     "scripts/*",
     "shared/*"
   ],
+  "nohoist": [
+    "examples/svelte-kit/vite/**",
+    "examples/vue-with-typescript/vite/**"
+  ],
```
decepulis, line 30: Don't miss this root level change! It seems important. I don't think it broke anything, but since monorepos scare me, maybe y'all should double check 👀

examples/svelte-kit/src/routes/+page.svelte
```diff
@@ -10,13 +10,13 @@
...
 		<li>
-			<a sveltekit:prefetch href="/mux-video">Mux Video Demo</a>
+			<a data-sveltekit-prefetch href="/mux-video">Mux Video Demo</a>
 		</li>
 		<li>
-			<a sveltekit:prefetch href="/mux-audio">Mux Audio Demo</a>
+			<a data-sveltekit-prefetch href="/mux-audio">Mux Audio Demo</a>
 		</li>
 		<li>
-			<a sveltekit:prefetch href="/mux-player">Mux Player Demo</a>
```
dylanjha, line 19: This is the kind of stuff that svelte changed? Yikes

dylanjha (commented):
LGTM since y'all say this works it makes sense to have our example be on the most recent version of svelte

Issue #396, opened 2022-09-16
Bug: @mux/mux-video-react autoPlay cannot be changed


shrugs:
### Which Mux Elements/Packages has a bug?

- `@mux/mux-video-react`

### Which browser(s) are you using?

- Firefox

### Which operating system(s) are you using?

macOS

### How are you using Mux Elements?

npm package `@mux/mux-video-react`

### Describe the bug

When setting the `updateAutoplay` function, one must pass a function -> function, since when `useState`'s setter receives a Function, it evaluates it. To store a function in state one must `setState(() -> () -> { ...code... })`

https://github.com/muxinc/elements/blob/b96b98746bbed68d4f82b0ac943c0ce194c1a72e/packages/mux-video-react/src/index.tsx#L67

As it is, react runs that function once, then sets `updateAutoplay` to the return value of that function, which is `undefined`. As such, when the effect runs `updateAutoplay` after changing the autoplay value in props, an invalid function is invoked.

## Steps To Reproduce

1. Render mux video element with autoplay = false
2. render mux video element with autoplay = true
3. see error

cjpillsbury:
@shrugs good catch! Poor `MuxVideo` (react) hasn't gotten enough attention these days. Adding a ticket, but also feel free to issue a PR if you wanted to contribute!

shrugs:
i'll propose this today!

PR #403, opened 2022-09-20
fix: allow autoplay to change in mux-video-react

shrugs:
see #396 for context


gkatsev (approved):
Thanks for the PR! This does seem like the correct fix as it now matches the [initial value as well](https://github.com/muxinc/elements/blob/82e337374857afd064c88b56d4e9fdb2f0fc4450/packages/mux-video-react/src/index.tsx#L47)

cjpillsbury (approved):
Awesome. Thanks for the assist, @shrugs!

PR #408, opened 2022-09-21
feat: remove deprecated .hls and .video props

gkatsev:
Please use ._hls for hls.js access if you must and use .media instead of .video for consistency.

BREAKING CHANGE: remove deprecated .hls and .video props.


luwes (approved):
👍

PR #410, opened 2022-09-21
fix: hide fullscreen button when fullscreen is unavailable

gkatsev:
Depends on https://github.com/muxinc/media-chrome/pull/311


cjpillsbury (approved):
awesome. Can merge now and do the media chrome version bump as a followup once it's released or hold off.

PR #414, opened 2022-09-22
feat(mux-player-react) lazy

decepulis:


packages/mux-player-react/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "outDir": "dist",
     "incremental": true,
     "target": "ES2019",
-    "module": "es6",
+    "module": "ES2020",
```
decepulis, line 9: This seems like a big change that shouldn't be missed.
ES2020 added dynamic import support, necessary for React.lazy

cjpillsbury: question(blocking): Have you confirmed this is doing what you expect? FYI we use `esbuild` (and not directly using `tsc`) for our build targets, so I'm honestly not sure how that relationship shakes out.

decepulis: I can at least confirm the build fails without it! During the  `build:types` phase:
```
@mux/mux-player-react:build: src/lazy.tsx(4,41): error TS1323: Dynamic imports are only supported when the '--module' flag is set to 'es2020', 'es2022', 'esnext', 'commonjs', 'amd', 'system', 'umd', 'node12', or 'nodenext'.
```

cjpillsbury: ah that tracks since we use `tsc` specifically for type def generation.

cjpillsbury: might want to confirm that the dist looks like you'd expect as well. You can muck around with our esbuild script usage locally to e.g. not minify if that makes things easier.

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const Fallback = (props: Omit<MuxPlayerProps, 'playerSoftwareVersion' | 'playerSoftwareName'>) => {
+  const { style, ...rest } = props;
+  return (
+    <>
+      {/* TODO: can we add mux-player to JSX.IntrinsicElements */}
+      {/* @ts-ignore */}
+      <mux-player
+        data-mux-player-react-placeholder
+        style={{
```
decepulis, line 15: Let's say a user is trying to style the component with styled-components, via className. Would placing styles here override those styles? Are we preventing the user from styling their mux-player? If so, we'll need to figure out how to define these styles in a lower-specificity way :(

packages/mux-player-react/package.json
```diff
@@ -28,13 +35,17 @@
...
     "build:cjs": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/cjs.json --format=cjs --loader:.css=text --outdir=dist --out-extension:.js=.cjs.js --external:react --external:@mux/* --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:esm": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/esm.json --format=esm --loader:.css=text --outdir=dist --out-extension:.js=.mjs --external:react --external:@mux/* --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
+    "build:cjs:lazy": "esbuild src/lazy.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/cjs.lazy.json --format=cjs --loader:.css=text --outdir=dist --out-extension:.js=.cjs.js --external:react --external:@mux/* --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
+    "build:esm:lazy": "esbuild src/lazy.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/esm.lazy.json --format=esm --loader:.css=text --outdir=dist --out-extension:.js=.mjs --external:react --external:@mux/* --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
```
decepulis, line 45: I KNOW these metafiles are wrong. I just don't know enough about metafiles to do it correctly.

gkatsev: Why do you say the metafiles are wrong?

Though, you'd probably want to add the two new metafiles to .npmignore

decepulis: I guess I don't know what metafiles are! And having trouble finding resources on them. I assumed, based on the naming of the other two, that they have some sort of standard naming that I didn't follow here.

But if you don't see any issues, then, um, great 😄

gkatsev: The metafiles are json that contain a bunch of information about the bundle. Things like which dependencies are included and what not. It's a file that esbuild generates. You can upload the file to https://www.bundle-buddy.com/esbuild, for example, to see it visually.

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,91 @@
...
+          backgroundRepeat: 'no-repeat',
+          backgroundSize: 'contain',
+          // user-implemented styles
+          ...style,
+        }}
+        // we pass the rest of the props to mux-player too.
+        // There is a possibility that mux-player imports before Suspense has had the opportunity
+        // to replace this placeholder with MuxPlayerReact.
+        // In that case, we want this placeholder to look as much like the incoming player as possible.
+        {...rest}
```
decepulis, line 33: I'm _sure_ this has side-effects that I haven't thought of yet.

examples/create-react-app-with-typescript/src/pages/MuxPlayerLazy.tsx
```diff
@@ -0,0 +1,16 @@
+import { useRef } from 'react';
+
+import MuxPlayer from '@mux/mux-player-react/lazy';
+
+function MuxPlayerSuspensePage() {
+  const mediaElRef = useRef(null)
+  return <MuxPlayer
+    ref={mediaElRef}
```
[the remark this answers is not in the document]
dylanjha: My S1 here is that it should be null until it's the MuxPlayer

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,121 @@
...
+              attributeName="transform"
+              attributeType="XML"
+              type="rotate"
+              dur="1s"
+              from="0 50 50"
+              to="360 50 50"
+              repeatCount="indefinite"
+            />
+          </path>
+        </svg>
```
luwes, line 81: I think we shouldn't show a loading spinner while loading.
it looks a bit distracting to me and the blurry image already gives an indication that something is loading there.
it it has multiple players on a page it's gonna look too noisy with all the spinners.

examples: 
https://player.vimeo.com/video/638369396
https://fast.wistia.net/embed/iframe/oifkgmxnkb?twitter=true

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (isIntersecting && onIntersection) {
+      onIntersection();
+    }
+  }, [isIntersecting, onIntersection]);
+
+  return (
+    <>
+      {/* TODO: can we add mux-player to JSX.IntrinsicElements */}
+      {/* @ts-ignore */}
+      <mux-player
```
luwes, line 30: shouldn't this import the React MuxPlayer? with the correct prop / event mapping etc

packages/mux-player-react/src/lazy.tsx
gkatsev: Yeah, loading spinner often makes things seem slower. Potentially, worth showing the loading spinner if it's taking more than a couple of seconds to load.

decepulis: Loading spinner pops in after half a second, here. Though yeah I'm considering removing it altogether. Next image works fine without a spinner. I think I'll remove it unless anyone's got strong opinions otherwise.

gkatsev: half a second is too short. Should be like 2 seconds minimum. I'm good with just removing it.

packages/mux-player-react/src/lazy.tsx
decepulis: So this is just the placeholder. And the placeholder should (in theory, [not quite in practice](https://mux.slack.com/archives/C032TQ30ARL/p1664827962292839?thread_ts=1664812026.713469&channel=C032TQ30ARL&message_ts=1664827962.292839)) exist only when code for mux-player and mux-player-react is still loading. And, of course, since that js is still loading, a `mux-player` element is basically just a div. 

The reason we use a `mux-player` html element here is not to have any of player's functionality, but rather, to inherit any styles targeted at `mux-player`.

packages/mux-player-react/src/lazy.tsx
dylanjha: > since the bundle is already imported, the fallback will be acting like a normal mux player until isIntersecting notifies React to replace the fallback with the real MuxPlayerReact

I see. That makes perfect sense. So the clever hack of putting an un-registered custom element of `<mux-player>` as the placeholder during suspense is coming back to haunt us

decepulis: We _could_ fix this with a wrapper.... though then we'd have to add additional/different styling guidelines for /lazy. It won't be just a drop-in replacement.

packages/mux-player-react/src/lazy.tsx
decepulis: it gon

packages/mux-player-react/src/lazy.tsx
luwes: cool makes sense

luwes:
noticed there are some merge conflicts. 
I removed `build:types` from the build command by the way, it's defined as a dependency in turbo.json now.

packages/mux-player-react/src/lazy.tsx
gkatsev: Is it worth noting this down in a comment in source?

examples/create-react-app-with-typescript/src/pages/MuxPlayerLazy.tsx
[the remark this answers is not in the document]
decepulis: Yeah I'll stick with that. That's what lazy components tend to do anyways.

packages/mux-player-react/src/lazy.tsx
decepulis: For sure. Unless we ditch this approach and use a wrapper 🤔
https://github.com/muxinc/elements/pull/414#discussion_r987016810

packages/mux-player-react/src/lazy.tsx
decepulis: Oh btw it's not just the risk of multiple interactions. We also have this multiple-loading flicker.

https://user-images.githubusercontent.com/8933386/194163231-b12f8f83-fb3d-4f75-95e2-e0ef76f1c212.mov

decepulis: Mmkay I went for the wrapper approach. Feels like an API defeat, but I'm just too worried about jank from our original approach.

packages/mux-player-react/src/lazy.tsx
decepulis: Not a problem now that we have a wrapper.

packages/mux-player-react/package.json
```diff
@@ -28,10 +35,14 @@
...
     "dev:cjs": "yarn build:cjs --watch=forever",
     "dev:esm": "yarn build:esm --watch=forever",
+    "dev:cjs:lazy": "echo 'esbuild cjs does not support code-splitting. See https://esbuild.github.io/api/#splitting for details'",
+    "dev:esm:lazy": "yarn build:esm:lazy --watch=forever",
     "dev:types": "yarn build:types -w",
-    "dev": "npm-run-all --parallel dev:types dev:cjs dev:esm",
+    "dev": "npm-run-all --parallel dev:types dev:cjs dev:esm dev:esm:lazy",
     "build:cjs": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/cjs.json --format=cjs --loader:.css=text --outdir=dist --out-extension:.js=.cjs.js --external:react --external:@mux/* --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:esm": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/esm.json --format=esm --loader:.css=text --outdir=dist --out-extension:.js=.mjs --external:react --external:@mux/* --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
+    "build:cjs:lazy": "echo 'esbuild cjs does not support code-splitting. See https://esbuild.github.io/api/#splitting for details'",
```
decepulis, line 44: So this is a thing

packages/mux-player-react/src/index.tsx
```diff
@@ -68,6 +67,7 @@ export type MuxPlayerProps = {
   metadataViewerUserId?: string;
   primaryColor?: string;
   secondaryColor?: string;
+  placeholder?: string;
```
decepulis, line 70: Moving this to its rightful place

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  return (
+    <div
+      ref={intersectionRef}
+      data-mux-player-react-lazy
+      style={{
+        // default mux-player styling
+        aspectRatio: '16/9',
+        display: 'block',
+        width: '100%',
```
decepulis, line 46: One downside to this approach? This wrapper doesn't automagically inherit mux player's styling. If we make a styling change to mux player, we'll have to make it here, too.

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,93 @@
...
+        // some styles just for this placeholder
+        position: 'relative',
+        backgroundColor: 'var(--media-background-color, #000)',
```
decepulis, line 49: Is this the correct way to get the same background color as mux player?

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const MuxPlayerIndex = React.lazy(() => import('./index'));
+
+const Overlay = () => (
+  <div
+    data-mux-player-react-lazy-overlay
+    style={{
+      position: 'absolute',
+      inset: 0,
+      backgroundColor: 'var(--controls-backdrop-color, rgba(0, 0, 0, 0.6))',
```
decepulis, line 16: Is this the correct way to get the same overlay color as mux player?

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,93 @@
...
+import type { MuxPlayerProps, MuxPlayerRefAttributes } from './index';
+
+const MuxPlayerIndex = React.lazy(() => import('./index'));
+
+const Overlay = () => (
+  <div
+    data-mux-player-react-lazy-overlay
+    style={{
+      position: 'absolute',
+      inset: 0,
```
decepulis, line 15: Considering we're using `aspect-ratio`, I'd say we're safe to use `inset`, too? 

https://caniuse.com/?search=inset

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const isBrowser = useIsBrowser();
+  // 2. The player has entered the viewport (if enabled).
+  const [isIntersecting, setIsIntersecting] = useState(false);
+  const shouldLoad = isBrowser && (loading === 'page' || isIntersecting);
+
+  return (
+    <Wrapper
+      style={style}
+      className={className}
+      placeholder={playerProps.placeholder}
```
decepulis, line 79: Quirk:
The wrapper intercepts `style` and `className`. There is no way for a user to pass a className or style to mux-player currently.

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,93 @@
+import React, { Suspense, useEffect, useState } from 'react';
+
+import useIsBrowser from './useIsBrowser';
+import useIsIntersecting from './useIsIntersecting';
+
+import type { MuxPlayerProps, MuxPlayerRefAttributes } from './index';
+
+const MuxPlayerIndex = React.lazy(() => import('./index'));
+
+const Overlay = () => (
```
decepulis, line 10: This overlay darkens the blurhash placeholder in the same way that the eventually-loading player would

examples/nextjs-with-typescript/pages/MuxPlayerLazy.tsx
```diff
@@ -0,0 +1,15 @@
...
+  const mediaElRef = useRef(null)
+  return <MuxPlayer
+    ref={mediaElRef}
+    playbackId="a4nOgmxGWg6gULfcBbAa00gXyfcwPnAFldF8RdsNyk8M"
+    streamType="on-demand"
+    placeholder="data:image/jpeg;base64,/9j/2wBDABQODxIPDRQSEBIXFRQYHjIhHhwcHj0sLiQySUBMS0dARkVQWnNiUFVtVkVGZIhlbXd7gYKBTmCNl4x9lnN+gXz/2wBDARUXFx4aHjshITt8U0ZTfHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHx8fHz/wAARCAASACADASIAAhEBAxEB/8QAGQAAAgMBAAAAAAAAAAAAAAAAAAUBAwQC/8QAHRAAAgIDAAMAAAAAAAAAAAAAAAECAwQSIREiMf/EABUBAQEAAAAAAAAAAAAAAAAAAAME/8QAGREAAwEBAQAAAAAAAAAAAAAAAAEDMRFR/9oADAMBAAIRAxEAPwBS4pEa7Geq2cn7GuLXgvT8ImuaUWYqkui/JpVT4M7bWlwWZNjm+g1wafTbEtiACoBnM/gvyQAOuCz0/9k="
+  />;
+}
+
+export default MuxPlayerSuspensePage;
```
decepulis, line 15: Do we wanna shape this page up and make it public-facing? I've just been using it for testing.

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  return (
+    <>
+      {/* 
+        Why do we have a mux-player element before the mux-player bundle is even loaded?
+        Before the bundle is loaded, this mux-player element just acts like a div. 
+        However, by calling this placeholder "mux-player", 
+        it now gets the same CSS applied to it that the eventual "real" mux-player element will.
+      */}
+      {/* TODO: can we add mux-player to JSX.IntrinsicElements */}
```
decepulis, line 34: Would anybody be interested in trying this out for me? For the life of me I couldn't figure it out.

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,108 @@
...
+interface MuxPlayerLazyProps extends MuxPlayerProps {
+  loading?: 'page' | 'viewport';
+}
+
+const MuxPlayer = React.forwardRef<MuxPlayerRefAttributes, MuxPlayerLazyProps>((props, ref) => {
+  const { loading = 'viewport', ...playerProps } = props;
+
+  // We load mux player once two conditions are met:
+  // 1. We're in a browser (react.lazy doesn't work on the server in react 17)
+  const isBrowser = useIsBrowser();
```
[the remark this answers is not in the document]
decepulis: Yup! If we don't have the typeof window check in a useEffect, we get hydration errors for the server-side and client-side first renders being different.

packages/mux-player-react/package.json
```diff
@@ -28,10 +35,14 @@
...
+    "build:esm:lazy": "esbuild src/lazy.tsx --splitting --target=es2019 --bundle --sourcemap --metafile=./dist/esm.lazy.json --format=esm --loader:.css=text --outdir=dist --out-extension:.js=.mjs --external:react --external:@mux/* --external:prop-types --define:PLAYER_VERSION=\"'$npm_package_version'\"",
```
gkatsev, line 45: this probably needs to get added to the `build` step. Also, do types get generated for it?

decepulis: I don't think so, because... something something turborepo?
https://github.com/muxinc/elements/pull/414#issuecomment-1267538556
@luwes

luwes: `build:types` is called by Turborepo so it's not needed to add to the list of the `build` command but in the command `build:types` there should probably be something added for the lazy package as well

e.g. "build:types": "tsc normalplayer && tsc lazyplayer"

decepulis: Thanks for clarifying! I'm on it.

gkatsev: But does anything build `build:esm:lazy`?

packages/mux-player-react/src/lazy.tsx
dylanjha: @cjpillsbury

dylanjha:
When this merges it must be squash & merged

packages/mux-player-react/package.json
decepulis: @luwes and @gkatsev sat down to do the thing and here's the output of `build:types` with no changes. Am I understanding correctly that tsc already knows about `lazy` and no action is required here?
image

luwes: yup you're right it's good already. sorry about that

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,107 @@
...
+      {/* @ts-ignore */}
+      <mux-player
+        ref={intersectionRef}
+        data-mux-player-react-lazy-placeholder
+        className={className || ''}
+        style={{
+          // default mux-player styling
+          aspectRatio: '16/9',
+          display: 'block',
+          backgroundColor: 'var(--media-control-background, #000)',
```
decepulis, line 44: Is this the correct way to make sure the placeholder background is the same color as the player background?

packages/mux-player-react/src/lazy.tsx
```diff
@@ -0,0 +1,107 @@
...
+        nohotkeys
+        aria-hidden
+        tabIndex="-1"
+      >
+        {/* Overlay */}
+        <div
+          style={{
+            position: 'absolute',
+            inset: 0,
+            backgroundColor: 'var(--controls-backdrop-color, rgba(0, 0, 0, 0.6))',
```
decepulis, line 67: Is this the correct way to get the color of the controls backdrop? The placeholder/poster overlay?

gkatsev: Yes, though, if you change it to `var(--controls-backdrop-color, var(--controls, rgba(0, 0, 0, 0.6)))`, it'll auto turn off if folks use `--controls: none` to turn off controls

decepulis: Ah shoot. I can't take this suggestion. The placeholder _always_ has `--controls: none` so that the player chrome wouldn't load on the placeholder before Suspense has had the chance to replace the placeholder with the real player. 

Since the placeholder assigns the `--controls` attribute, there's no way that I can think of of knowing the user's `--controls` preference within. So we'll just default to always darkening the placeholder.

gkatsev: Ah, cool. Forgot that it's the placeholder and that controls are disabled.

packages/mux-player-react/src/lazy.tsx
luwes: I'm not sure if this is the right CSS var. `--media-control-background` changes the background of each control separately.

better one I think is `var(--media-background-color, #000)` which is the background color of `<media-controller>`

https://github.com/muxinc/media-chrome/blob/5ea0d24bc18de58587b8f90137f72cfde585879f/src/js/media-container.js#L38-L38

decepulis: Thank you!

luwes (approved):
LGTM 👍

dylanjha (approved):

PR #416, opened 2022-09-23
Disable some controls when there isn't a playback-id

gkatsev:
Depends on https://github.com/muxinc/media-chrome/pull/320 for the disabled buttons functionality. This PR adds styles and the attributes as appropriate.

### DIFF
--- a/packages/mux-player/src/media-chrome/time-display.ts
+++ b/packages/mux-player/src/media-chrome/time-display.ts
@@ -20,7 +20,7 @@ const ButtonPressedKeys = ['Enter', ' '];
 
 class MxpTimeDisplay extends globalThis.HTMLElement {
   static get observedAttributes() {
-    return ['hide-duration', 'remaining'];
+    return ['hide-duration', 'remaining', 'disabled', 'aria-disabled'];
   }
   static styles: string = styles;
   static template: HTMLTemplateElement = template;
@@ -98,6 +98,12 @@ class MxpTimeDisplay extends globalThis.HTMLElement {
         this.timeDisplayEl?.setAttribute('remaining', '');
       } else {
         this.timeDisplayEl?.removeAttribute('remaining');
+      }
+    } else if (attrName === 'disabled' || (attrName === 'aria-disabled' && _oldValue !== newValue)) {
+      if (newValue == null) {
+        this.timeDisplayEl?.removeAttribute(attrName);
+      } else {
+        this.timeDisplayEl?.setAttribute(attrName, newValue);
       }
     }
   }
--- a/packages/mux-player/src/media-chrome/time-display.ts
+++ b/packages/mux-player/src/media-chrome/time-display.ts
@@ -89,17 +89,11 @@ class MxpTimeDisplay extends globalThis.HTMLElement {
     }
   }
 
-  attributeChangedCallback(attrName: string, _oldValue: string | null, newValue: string | null) {
+  attributeChangedCallback(attrName: string, oldValue: string | null, newValue: string | null) {
     if (attrName === 'hide-duration') {
       this.#toggleDuration();
     }
-    if (attrName === 'remaining') {
-      if (newValue != null) {
-        this.timeDisplayEl?.setAttribute('remaining', '');
-      } else {
-        this.timeDisplayEl?.removeAttribute('remaining');
-      }
-    } else if (attrName === 'disabled' || (attrName === 'aria-disabled' && _oldValue !== newValue)) {
+    if (['remaining', 'disabled', 'aria-disabled'].includes(attrName) && oldValue !== newValue) {
       if (newValue == null) {
         this.timeDisplayEl?.removeAttribute(attrName);
       } else {
--- a/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
+++ b/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
@@ -15,7 +15,7 @@ const MediaChromeSizes = {
 };
 
 type ThemeMuxTemplateProps = {
-  src?: string | null;
+  hasSrc: boolean;
   streamType: string | null;
   nohotkeys?: boolean;
   hotkeys: string | null;
--- a/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
+++ b/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
@@ -206,14 +206,14 @@ const MediaPlayButton = (props: ComponentProps) => html`
   </media-play-button>`;
 
 // prettier-ignore
-const MediaSeekBackwardButton = ({ backwardSeekOffset, src }: ComponentProps) => html`
-  <media-seek-backward-button seek-offset="${backwardSeekOffset}" disabled="${!src}" aria-disabled="${src ? false : 'true'}">
+const MediaSeekBackwardButton = ({ backwardSeekOffset, hasSrc }: ComponentProps) => html`
+  <media-seek-backward-button seek-offset="${backwardSeekOffset}" disabled="${!hasSrc}" aria-disabled="${hasSrc ? false : 'true'}">
     ${icons.SeekBackward({ value: backwardSeekOffset })}
   </media-seek-backward-button>`;
 
 // prettier-ignore
-const MediaSeekForwardButton = ({ forwardSeekOffset, src }: ComponentProps) => html`
-  <media-seek-forward-button seek-offset="${forwardSeekOffset}" disabled="${!src}" aria-disabled="${src ? false : 'true'}">
+const MediaSeekForwardButton = ({ forwardSeekOffset, hasSrc }: ComponentProps) => html`
+  <media-seek-forward-button seek-offset="${forwardSeekOffset}" disabled="${!hasSrc}" aria-disabled="${hasSrc ? false : 'true'}">
     ${icons.SeekForward({ value: forwardSeekOffset })}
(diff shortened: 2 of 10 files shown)


packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -72,6 +74,8 @@ export default class MediaThemeMux extends MediaTheme {
       title: this.getAttribute('title'),
     };
 
+    props.nohotkeys = props.src == null ? true : props.nohotkeys;
```
cjpillsbury, line 77: 😎

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -85,6 +89,7 @@ export default class MediaThemeMux extends MediaTheme {
           ${cssStr}
         </style>
         <media-controller
+          gestures-disabled="${props.src == null}"
```
cjpillsbury, line 92: 😎 might break down longer term, but works for now!

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
cjpillsbury, line 77: might be worth adding a code comment or two around these, since src <--> disabling may not be obvious to folks.

packages/mux-player/src/template.ts
```diff
@@ -28,6 +28,13 @@ const isLiveOrDVR = (props: MuxTemplateProps) =>
 
 export const content = (props: MuxTemplateProps) => html`
   <${unsafeStatic(castThemeName(props.theme) ?? 'media-theme-mux')}
+    src="${
+      !!props.src
+        ? props.src
+        : props.playbackId
+        ? getSrcFromPlaybackId(props.playbackId, { domain: props.customDomain, token: props.tokens.playback })
+        : false
+    }"
```
luwes, line 37: I really think we shouldn't use `src` here as it's implying the value of src is a dependency for the the theme while it's not really. can this be renamed to something like `nosrc`, `unloaded`.

it's a bit a shame this has to be passed at all, is there any way this can be derived from within?

luwes (dismissed):

packages/mux-player/src/template.ts
gkatsev: Looks like we can do something like `hasSrc = this.querySelector('mux-video').hasAttribute('src')`. Does that make sense?

gkatsev: Unfortunately, it looks like `hasAttribute()` on `mux-video` is too early.

luwes: ok that's a shame. a shorter attribute makes sense then I think.

gkatsev: Let me try and make it a boolean attribute like `has-src` or something

packages/mux-player/src/media-chrome/time-display.ts
```diff
@@ -99,6 +99,12 @@ class MxpTimeDisplay extends globalThis.HTMLElement {
       } else {
         this.timeDisplayEl?.removeAttribute('remaining');
       }
+    } else if (attrName === 'disabled' || (attrName === 'aria-disabled' && _oldValue !== newValue)) {
```
cjpillsbury, line 102: suggestion(non-blocking): aside from the state change logic, this is effectively the same as the `remaining` condition. We should be able to simplify this to something like our `includes` check convention, i.e.:
```js
if (['remaining', 'disabled', 'aria-disabled'].includes(attrName) && oldValue !== newValue) {
// ...
```

packages/mux-player/src/media-theme-mux/styles.css
```diff
@@ -61,6 +61,12 @@
   width: 100%;
 }
 
+[disabled],
+[aria-disabled='true'] {
+  opacity: 60%;
```
cjpillsbury, line 66: 👌

packages/mux-player/src/template.ts
```diff
@@ -26,8 +26,16 @@ const isLive = (props: MuxTemplateProps) => [StreamTypes.LIVE, StreamTypes.LL_LI
 const isLiveOrDVR = (props: MuxTemplateProps) =>
   [StreamTypes.LIVE, StreamTypes.LL_LIVE, StreamTypes.DVR, StreamTypes.LL_DVR].includes(props.streamType as any);
 
-export const content = (props: MuxTemplateProps) => html`
+export const content = (props: MuxTemplateProps) => {
+  const hasSrc = !!props.src
+    ? !!props.src
+    : props.playbackId
+    ? !!getSrcFromPlaybackId(props.playbackId, { domain: props.customDomain, token: props.tokens.playback })
```
cjpillsbury, line 33: question(blocking): I think we don't need `getSrcFromPlaybackId` in this nested ternary? Unless I'm missing something, we should be able to just check:

```js
const hasSrc = !!(props.src || props.playbackId);
```

Related: Should this just live in `getProps()`?

packages/mux-player/test/template.test.js
```diff
@@ -19,9 +19,10 @@ describe('<mux-player> template render', () => {
     );
   });
 
-  it('template live extra-small', function () {
+  it('template live extra-small w/o src, not live', function () {
```
cjpillsbury, line 22: nitpick(blocking) lol maybe make this a little more verbose, since I'm not sure what it means to be "template live... not live" 😅

cjpillsbury (changes requested):
a few small asks. Otherwise this is 😎 (looks good, too).

packages/mux-player/test/template.test.js
gkatsev: oops, the second "live" is "not in live window". I'll update it.

packages/mux-player/src/template.ts
```diff
@@ -125,9 +133,10 @@ export const content = (props: MuxTemplateProps) => html`
     ${
       isLiveOrDVR(props)
         ? html`<button
+            disabled="${!hasSrc}"
```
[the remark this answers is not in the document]
gkatsev: The issue is that there should be a difference between disabled no source and not-clickable because we're in the live window. Since, currently, the "in live window" styles are applied on aria-disabled. So, my current workaround was to leave current styles alone and target disabled+aria-disabled as the no-src option, to minimize changes in here and the CSS. However, maybe we want a separate attribute that indicates that this is in the live window separate from disabled/aria-disabled?

packages/mux-player/src/template.ts
gkatsev: yeah, maybe we `hasSrc` should move into `getProps()` at this point.

packages/mux-player/src/template.ts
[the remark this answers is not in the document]
cjpillsbury: we can probably hold off on that since we're anticipating an "official" implementation soon-ish.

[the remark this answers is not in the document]
gkatsev: so, keep as is and fix it later?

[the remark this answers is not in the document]
cjpillsbury: yeah keeping it as is is fine for now

packages/mux-player/src/styles.css
```diff
@@ -22,21 +21,24 @@ a {
...
   color: #fb3c4d;
 }
 
+[part~='seek-live'][disabled] {
+  opacity: 60%;
+  cursor: not-allowed;
+}
+[part~='seek-live'][in-live-window] {
+  opacity: 100%;
+  cursor: unset;
```
gkatsev, line 39: if we're in-live-window, the control is also disabled, which gets a 60% opacity, but we want it visible to users here and not dimmed.
Same with the cursor, we want it to not be a pointer but not be not-allowed.

cjpillsbury: good description.

cjpillsbury (approved):
😍

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -193,21 +199,21 @@ const ChromeRenderer = (props: ThemeMuxTemplateProps) => {
 };
 
 // prettier-ignore
-const MediaPlayButton = () => html`
-  <media-play-button>
+const MediaPlayButton = (props: ComponentProps) => html`
+  <media-play-button disabled="${!props.hasSrc}" aria-disabled="${props.hasSrc ? false : 'true'}">
```
luwes, line 203: nit (non-blocking): 
```suggestion
  <media-play-button disabled="${!props.hasSrc}" aria-disabled="${props.hasSrc && 'true'}">
```

packages/mux-player/src/media-theme-mux/styles.css
```diff
@@ -61,6 +61,12 @@
...
+  cursor: not-allowed;
```
luwes, line 67: (blocking): is this really needed? I feel it a bit much. like the fact that the cursor would not become a hand would be sufficient IMO

luwes (changes requested):

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
gkatsev: that does simplify it a bit 👍🏻

luwes (approved):

luwes (approved):

PR #424, opened 2022-09-26
cleanup playback-core

luwes:
index.ts of playback-core was getting a bit disorganized, moving the type related things to a types file and moving the utils that are not needed externally to the util file.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -3,15 +3,23 @@ import Hls from 'hls.js';
 import { setupAutoplay } from './autoplay';
 import { MediaError } from './errors';
 import { setupTracks, addTextTrack, removeTextTrack } from './tracks';
-import { inSeekableRange, addEventListenerWithTeardown, toPlaybackIdParts, getType, getStreamTypeConfig } from './util';
+import { inSeekableRange, addEventListenerWithTeardown, toPlaybackIdParts, getType } from './util';
 import type { Autoplay, UpdateAutoplay } from './autoplay';
-import { PlaybackTypes, ExtensionMimeTypeMap, MuxMediaProps, MuxMediaPropsInternal } from './types';
+import {
+  ValueOf,
+  StreamTypes,
+  PlaybackTypes,
+  ExtensionMimeTypeMap,
+  MuxMediaProps,
+  MuxMediaPropsInternal,
+} from './types';
 
 export { mux, Hls, MediaError, Autoplay, UpdateAutoplay, setupAutoplay, addTextTrack, removeTextTrack };
 export * from './types';
 
 const userAgentStr = globalThis?.navigator?.userAgent ?? '';
 const isAndroid = userAgentStr.toLowerCase().indexOf('android') !== -1;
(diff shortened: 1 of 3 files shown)


packages/playback-core/src/index.ts
```diff
@@ -23,173 +21,28 @@ export const generatePlayerInitTime = () => {
...
-  }
-
-  const { src } = props;
-
-  if (!src) return '';
-
-  return inferMimeTypeFromURL(src);
-};
-
-export const getStreamTypeConfig = (streamType?: ValueOf<StreamTypes>) => {
```
gkatsev, line 164: non-blocking: I kind of think this should stay in this file because this method defines the default configs. Though, I won't insist on this if you'd prefer otherwise.

gkatsev (commented):
The methods that moved to util.ts are no longer exported from the module, is that an issue?

packages/playback-core/src/index.ts
luwes: good call, will move

gkatsev (approved):

packages/playback-core/src/index.ts
```diff
@@ -1,19 +1,24 @@
...
 import { setupTracks, addTextTrack, removeTextTrack } from './tracks';
-import { isKeyOf, inSeekableRange, addEventListenerWithTeardown } from './util';
-import type { Autoplay, UpdateAutoplay } from './autoplay';
-
-export type ValueOf<T> = T[keyof T];
-export type Metadata = Partial<Options['data']>;
-export type PlaybackEngine = Hls;
-export { mux, Hls, MediaError, Autoplay, UpdateAutoplay, setupAutoplay, addTextTrack, removeTextTrack };
+import { inSeekableRange, addEventListenerWithTeardown, toPlaybackIdParts, getType } from './util';
+import {
```
cjpillsbury, line 7: nitpick(blocking): for types, we should try to use `import type` wherever possible if that's all we're importing.

packages/playback-core/src/index.ts
```diff
@@ -1,19 +1,24 @@
...
+  ValueOf,
+  StreamTypes,
+  PlaybackTypes,
+  ExtensionMimeTypeMap,
+  MuxMediaProps,
+  MuxMediaPropsInternal,
+} from './types';
+
+export { mux, Hls, MediaError, setupAutoplay, addTextTrack, removeTextTrack };
+export * from './types';
```
cjpillsbury, line 17: question(blocking): If there's syntax for `export type` here, update it. If it takes any meaningful amount of time to figure that out, leave it as is.

packages/playback-core/src/util.ts
```diff
@@ -1,4 +1,4 @@
-type KeyTypes = string | number | symbol;
+import { isKeyOf, ExtensionMimeTypeMap, MimeTypeShorthandMap, MuxMediaProps } from './types';
```
cjpillsbury, line 1: nitpick(blocking): `import type`

cjpillsbury (changes requested):
a few minor nitpicks, otherwise LGTM.

PR #425, opened 2022-09-28
fix: subsequent preload changes

luwes:
updates the logic for preload to be updated after initial src load. e.g. from preload=none to preload=auto
this is required for starting PiP when `preload` was set to none.

I did make a change to playback-core so it returns the setAutoplay and setPreload functions in an object along with the "engine" (hls.js) because this makes it a lot easier in the React versions, don't have to mess with those hooks too much. 

in the WC versions it's also cleaner and less error prone.

less issues with importing types etc.

### DIFF
--- a/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
@@ -71,6 +71,7 @@ const DEFAULT_INITIAL_STATE: Partial<MuxPlayerProps> = Object.freeze({
   muted: undefined,
   debug: undefined,
   autoPlay: undefined,
+  preload: undefined,
   startTime: undefined,
   currentTime: undefined,
   paused: undefined,
@@ -320,6 +321,7 @@ function MuxPlayerPage({ location }: Props) {
         volume={state.volume}
         paused={state.paused}
         autoPlay={state.autoPlay}
+        preload={state.preload}
         streamType={state.streamType}
         audio={state.audio}
         primaryColor={state.primaryColor}
@@ -442,6 +444,12 @@ function MuxPlayerPage({ location }: Props) {
           name="autoPlay"
           onChange={genericOnChange}
           values={[true, false, 'any', 'muted']}
+        />
+        <EnumRenderer
+          value={state.preload}
+          name="preload"
+          onChange={genericOnChange}
+          values={['none', 'metadata', 'auto']}
         />
         <BooleanRenderer
           value={state.muted}
--- a/packages/mux-audio-react/src/index.tsx
+++ b/packages/mux-audio-react/src/index.tsx
@@ -4,16 +4,13 @@ import PropTypes from 'prop-types';
 import {
   allMediaTypes,
   initialize,
-  type Autoplay,
-  type UpdateAutoplay,
-  setupAutoplay,
   MuxMediaProps,
   StreamTypes,
   PlaybackTypes,
   toMuxVideoURL,
-  PlaybackEngine,
   generatePlayerInitTime,
 } from '@mux/playback-core';
+import type { PlaybackCore } from '@mux/playback-core';
 import { getPlayerVersion } from './env';
 
 export type Props = Omit<
@@ -38,23 +35,18 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
     src: outerSrc,
     children,
     autoPlay,
+    preload,
     ...restProps
   } = props;
 
   const [playerInitTime] = useState(generatePlayerInitTime());
-
   const [src, setSrc] = useState<MuxMediaProps['src']>(toMuxVideoURL(playbackId) ?? outerSrc);
-
-  const playbackEngineRef = useRef<PlaybackEngine | undefined>(undefined);
+  const playbackCoreRef = useRef<PlaybackCore | undefined>(undefined);
   const innerMediaElRef = useRef<HTMLAudioElement>(null);
-
   const mediaElRef = useCombinedRefs(innerMediaElRef, ref);
-  const updateAutoplayRef = useRef<UpdateAutoplay | undefined>(undefined);
-  const [updateAutoplay, setUpdateAutoplay] = useState(() => (x: any) => {});
 
   useEffect(() => {
-    const src = toMuxVideoURL(playbackId) ?? outerSrc;
-    setSrc(src);
+    setSrc(toMuxVideoURL(playbackId) ?? outerSrc);
   }, [outerSrc, playbackId]);
 
   useEffect(() => {
@@ -66,17 +58,16 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
       playerSoftwareVersion,
       autoplay: autoPlay,
     };
-    const nextPlaybackEngineRef = initialize(propsWithState, mediaElRef.current, playbackEngineRef.current);
-    playbackEngineRef.current = nextPlaybackEngineRef;
-    setUpdateAutoplay(() => {
-      if (!mediaElRef.current) return;
-      setupAutoplay(mediaElRef.current, autoPlay, playbackEngineRef.current);
-    });
+    playbackCoreRef.current = initialize(propsWithState, mediaElRef.current, playbackCoreRef.current);
   }, [src]);
 
   useEffect(() => {
-    updateAutoplay(autoPlay);
+    playbackCoreRef.current?.setAutoplay(autoPlay);
   }, [autoPlay]);
+
+  useEffect(() => {
+    playbackCoreRef.current?.setPreload(preload);
+  }, [preload]);
 
   return (
     /** @TODO Fix types relationships (CJP) */
--- a/packages/mux-audio/package.json
+++ b/packages/mux-audio/package.json
@@ -53,7 +53,7 @@
   "devDependencies": {
     "@mux/test-esm-exports": "0.1.0",
     "@open-wc/testing": "^3.0.3",
-    "@web/dev-server-esbuild": "^0.2.16",
+    "@web/dev-server-esbuild": "^0.3.2",
     "@web/dev-server-import-maps": "^0.0.6",
     "@web/test-runner": "^0.13.26",
     "downlevel-dts": "^0.9.0",
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -109,14 +109,16 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
   get preload() {
     const val = this.getAttribute('preload') as HTMLMediaElement['preload'];
     if (val === '') return 'auto';
-    if (val && ['none', 'metadata', 'auto'].includes(val)) return val;
+    if (['none', 'metadata', 'auto'].includes(val)) return val;
     return super.preload;
   }
 
   set preload(val) {
-    if (val === this.preload) return;
+    // don't cause an infinite loop
+    // check the attribute because an empty string maps to the `auto` prop
+    if (val == this.getAttribute('preload')) return;
 
-    if (val && ['', 'none', 'metadata', 'auto'].includes(val)) {
+    if (['', 'none', 'metadata', 'auto'].includes(val)) {
       this.setAttribute('preload', val);
     } else {
       this.removeAttribute('preload');
@@ -287,7 +289,7 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
         if (newValue === oldValue) {
           break;
         }
-        this.__core?.setPreload(this.preload);
+        this.__core?.setPreload(newValue as HTMLMediaElement['preload']);
         break;
       case Attributes.PLAYBACK_ID:
         /** @TODO Improv+Discuss - how should playback-id update wrt src attr changes (and vice versa) (CJP) */
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -1,7 +1,7 @@
 import { globalThis } from 'shared-polyfills';
 import {
   initialize,
-  setupAutoplay,
+  generatePlayerInitTime,
   MuxMediaProps,
   StreamTypes,
   PlaybackTypes,
@@ -9,11 +9,9 @@ import {
   toMuxVideoURL,
   teardown,
   Metadata,
-  mux,
-  generatePlayerInitTime,
   MediaError,
 } from '@mux/playback-core';
-import type { PlaybackEngine, UpdateAutoplay, ExtensionMimeTypeMap } from '@mux/playback-core';
+import type { PlaybackCore, PlaybackEngine, Autoplay, ExtensionMimeTypeMap } from '@mux/playback-core';
 import { getPlayerVersion } from './env';
 // this must be imported after playback-core for the polyfill to be included
 import CustomAudioElement, { AudioEvents } from './CustomAudioElement';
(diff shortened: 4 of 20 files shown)


packages/mux-audio-react/src/index.tsx
```diff
@@ -66,18 +58,17 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
       playerSoftwareVersion,
       autoplay: autoPlay,
     };
-    const nextPlaybackEngineRef = initialize(propsWithState, mediaElRef.current, playbackEngineRef.current);
-    playbackEngineRef.current = nextPlaybackEngineRef;
-    setUpdateAutoplay(() => {
-      if (!mediaElRef.current) return;
-      setupAutoplay(mediaElRef.current, autoPlay, playbackEngineRef.current);
-    });
```
luwes, line 74: to have 2 of these was a too noisy, returning a plain object from initialize makes things a lot easier IMO

packages/mux-video/src/index.ts
```diff
@@ -180,6 +177,23 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
     }
   }
 
+  get preload() {
+    const val = this.getAttribute('preload') as HTMLMediaElement['preload'];
+    if (val === '') return 'auto';
```
gkatsev, line 182: this may not be correct. For example, in Chrome, `document.createElement('video').preload` will return `metadata`

packages/mux-audio/src/index.ts
```diff
@@ -236,23 +250,19 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
...
+    teardown(this.nativeEl, this.__core);
+    this.__core = undefined;
   }
 
   attributeChangedCallback(attrName: string, oldValue: string | null, newValue: string | null) {
     super.attributeChangedCallback(attrName, oldValue, newValue);
 
     switch (attrName) {
-      case 'src':
+      case 'src': {
```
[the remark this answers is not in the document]
luwes: yes, eslint complains about that if using constants in a switch / case

packages/mux-video/src/index.ts
luwes: yes, but if you set the attribute preload to an empty string that corresponds with the preload prop being `auto`

basically attrs `preload=""` and `preload="auto"` are the same 

removing will use the default for the browser

packages/playback-core/src/preload.ts
```diff
@@ -0,0 +1,77 @@
...
+      hasLoadedSource = true;
+      hls.loadSource(src);
+    }
+  };
+
+  addEventListenerWithTeardown(
+    mediaEl,
+    'play',
+    () => {
+      switch (preload) {
```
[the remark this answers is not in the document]
luwes: oops, yes I meant to update the `preload` var in `updateHlsPreload` as well. great catch!
I had it before like that but then changed it to a const in that function

packages/mux-video/src/index.ts
luwes: you mean `super.preload` to get the native video element prop? 
it's a shame the attribute callback is not fired without configuring these getters and setters. that would be ideal

gkatsev: er, yes, however it is to get the native element property.

luwes: I think it's correct though, I don't just return only `super.preload` in the getter because I find since the preload property is overridden on mux-video it's best to derive it from the attribute if possible. only for the case when no preload is set we check the super.preload to get the browser default.

gkatsev: Sounds good!

cjpillsbury:
> requires & based off #424

suggestion(non-blocking): looks like you moved this all into one PR. Might want to update the description for clarity.

examples/nextjs-with-typescript/pages/MuxPlayer.tsx
```diff
@@ -71,6 +71,7 @@ const DEFAULT_INITIAL_STATE: Partial<MuxPlayerProps> = Object.freeze({
   muted: undefined,
   debug: undefined,
   autoPlay: undefined,
+  preload: undefined,
```
cjpillsbury, line 74: 😎

packages/mux-audio/src/index.ts
```diff
@@ -109,6 +106,25 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
...
+    const val = this.getAttribute('preload') as HTMLMediaElement['preload'];
+    if (val === '') return 'auto';
+    if (['none', 'metadata', 'auto'].includes(val)) return val;
+    return super.preload;
+  }
+
+  set preload(val) {
+    // don't cause an infinite loop
+    // check the attribute because an empty string maps to the `auto` prop
+    if (val == this.getAttribute('preload')) return;
```
cjpillsbury, line 119: 👌

packages/mux-audio/src/index.ts
```diff
@@ -304,7 +328,7 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
 
 type MuxAudioElementType = typeof MuxAudioElement;
 declare global {
-  var MuxAudioElement: MuxAudioElementType;
+  var MuxAudioElement: MuxAudioElementType; // eslint-disable-line
```
cjpillsbury, line 331: question(non-blocking): What were the linting errors here?

packages/mux-elements-codemod/package.json
```diff
@@ -67,7 +66,6 @@
     "rules": {
       "no-shadow": "error",
       "no-extra-boolean-cast": 0,
-      "@mux/i18n/no-substitutions": "error",
```
cjpillsbury, line 70: question(non-blocking): Why were these changes introduced in this PR?

packages/mux-audio/package.json
```diff
@@ -53,7 +53,7 @@
   "devDependencies": {
     "@mux/test-esm-exports": "0.1.0",
     "@open-wc/testing": "^3.0.3",
-    "@web/dev-server-esbuild": "^0.2.16",
+    "@web/dev-server-esbuild": "^0.3.2",
```
cjpillsbury, line 56: question(non-blocking): Why did we upgrade the version in this PR? Are there changes we should be aware of?

packages/mux-video-react/src/index.tsx
```diff
@@ -63,18 +58,17 @@ const MuxVideo = React.forwardRef<HTMLVideoElement | undefined, Partial<Props>>(
       playerSoftwareVersion,
       autoplay: autoPlay,
     };
-    const nextPlaybackEngineRef = initialize(propsWithState, mediaElRef.current, playbackEngineRef.current);
-    playbackEngineRef.current = nextPlaybackEngineRef;
-    setUpdateAutoplay(() => () => {
-      if (!mediaElRef.current) return;
-      setupAutoplay(mediaElRef.current, autoPlay, playbackEngineRef.current);
-    });
+    playbackCoreRef.current = initialize(propsWithState, mediaElRef.current, playbackCoreRef.current);
```
cjpillsbury, line 61: 😎

packages/mux-video/src/index.ts
```diff
@@ -66,14 +64,12 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
-  // Keeping this named "__hls" since it's exposed for unadvertised "advanced usage" via getter that assumes specifically hls.js (CJP)
-  protected __hls?: PlaybackEngine;
+  protected __core?: PlaybackCore;
   protected __playerInitTime: number;
   protected __metadata: Readonly<Metadata> = {};
   protected __playerSoftwareVersion?: string;
   protected __playerSoftwareName?: string;
-  protected __updateAutoplay?: UpdateAutoplay;
-  protected __errorTranslator?: Function;
+  protected __errorTranslator?: (errorEvent: any) => any;
```
cjpillsbury, line 72: question(non-blocking): Why was this change part of the preload PR?

cjpillsbury, line 72: nitpick(non-blocking): If we're updating the type to be more precise, can this be stricter than `any` for the param & the yield? Feels like it should.

packages/mux-video/src/index.ts
```diff
@@ -415,7 +435,7 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
 
 type MuxVideoElementType = typeof MuxVideoElement;
 declare global {
-  var MuxVideoElement: MuxVideoElementType;
+  var MuxVideoElement: MuxVideoElementType; // eslint-disable-line
```
cjpillsbury, line 438: question(non-blocking): Same as above.

packages/playback-core/src/index.ts
```diff
@@ -23,174 +30,41 @@ export const generatePlayerInitTime = () => {
...
+  props: Partial<MuxMediaPropsInternal>,
+  mediaEl?: HTMLMediaElement | null,
+  core?: PlaybackCore
+) => {
+  // Automatically tear down previously initialized mux data & hls instance if it exists.
+  teardown(mediaEl, core);
 
-  return {};
-};
+  muxMediaState.set(mediaEl as HTMLMediaElement, {});
```
cjpillsbury, line 51: idea(non-blocking): Since we've moved to a factory model, we can likely refactor the module-level state to be defined here in the closure, correct?

packages/playback-core/src/index.ts
```diff
@@ -23,174 +30,41 @@ export const generatePlayerInitTime = () => {
...
-  return muxMediaState.get(mediaEl)?.error;
+  return {
+    engine: nextHlsInstance,
+    setAutoplay,
+    setPreload,
+  };
 };
 
-export const teardown = (mediaEl?: HTMLMediaElement | null, hls?: Pick<Hls, 'detachMedia' | 'destroy'>) => {
+export const teardown = (mediaEl?: HTMLMediaElement | null, core?: PlaybackCore) => {
```
cjpillsbury, line 66: idea(non-blocking): Per our out of band conversation, now that we're using a factory model, `teardown` can likely belong to the `PlaybackCore` object.

packages/playback-core/test/player.test.js
```diff
@@ -0,0 +1,169 @@
+import { assert, aTimeout } from '@open-wc/testing';
+import { initialize, teardown, MediaError, getError } from '../src/index.ts';
+
+describe('playback core', () => {
```
cjpillsbury, line 4: thought(non-blocking): ❤️

packages/playback-core/test/player.test.js
```diff
@@ -0,0 +1,169 @@
...
+    const video = document.createElement('video');
+    initialize(
+      {
+        src: 'https://mux.com/', // not working src url
+      },
+      video
+    );
+    // error event must be attached after initialize(), otherwise the
+    // native error will not be stopped propagating in playback core.
+    await new Promise((resolve) => video.addEventListener('error', resolve));
```
cjpillsbury, line 40: suggestion(non-blocking): This comment isn't quite clear (at least for me). I think you mean something like:

> Wait for the error event event to be dispatched before checking the error state

packages/playback-core/test/player.test.js
```diff
@@ -0,0 +1,169 @@
...
+    const core = initialize(
+      {
+        src: 'https://stream.mux.com/23s11nz72DsoN657h4314PjKKjsF2JG33eBQQt6B95I.m3u8',
+        autoplay: false,
+      },
+      video
+    );
+
+    core.setAutoplay('any');
+    await new Promise((resolve) => video.addEventListener('playing', resolve));
```
cjpillsbury, line 65: nitpick(non-blocking): I believe the assertion below is redundant, since `playing` implies `!paused` (see, e.g.: https://html.spec.whatwg.org/multipage/media.html#event-media-playing). If you're keeping it for parity/clarity with the related tests, that's 👌

packages/playback-core/test/player.test.js
```diff
@@ -0,0 +1,169 @@
...
+    assert.equal(video.error.message, '');
+
+    assert.equal(getError(video).code, MediaError.MEDIA_ERR_SRC_NOT_SUPPORTED);
+    assert.equal(
+      getError(video).message,
+      'An unsupported error occurred. The server or network failed, or your browser does not support this format.'
+    );
+  });
+
+  it('setAutoplay("any")', async function () {
```
cjpillsbury, line 52: suggestion(non-blocking): It's valid to have nested `describe()`s. Consider grouping related tests into their own `describe()`. e.g.:

```js
describe("autoplay behaviors, () => {
  // ...
```

packages/playback-core/test/player.test.js
```diff
@@ -0,0 +1,169 @@
...
+    initialize(
+      {
+        src: 'https://stream.mux.com/23s11nz72DsoN657h4314PjKKjsF2JG33eBQQt6B95I.m3u8',
+        preload: 'none',
+      },
+      video
+    );
+
+    assert.equal(video.preload, 'none', 'preload is none');
+    await aTimeout(3000);
```
cjpillsbury, line 102: nitpick(non-blocking): I think we're testing too much here. It looks like you're trying to test:

1. that `preload="none"` will not load any data
2. that `preload="none"` will **_eventually_** load data by explicitly invoking `play()`.

The first test should probably have some event listeners that will cause the test to fail if they are invoked (e.g. `loadstart` and/or `loadedmetadata`) and passes if, after some amount of time (such as your example 3 seconds), its `readyState=0` (aka "HAVE_NOTHING").

The second test should have the same initialization and pass if the relevant event listeners (e.g. `loadeddata` and/or `canplay` are invoked after invoking `play()` (can use the `done()` function for this case

packages/mux-audio/src/index.ts
luwes: the use of `var`. eslint doesn't know this is typescript, I know it's kind of silly 

they even ignored this valid issue
https://github.com/typescript-eslint/typescript-eslint/issues/3617

packages/playback-core/test/player.test.js
```diff
@@ -0,0 +1,169 @@
...
+    initialize(
+      {
+        src: 'https://stream.mux.com/23s11nz72DsoN657h4314PjKKjsF2JG33eBQQt6B95I.m3u8',
+        preload: 'auto',
+      },
+      video
+    );
+
+    assert.equal(video.preload, 'auto', 'preload is auto');
+    await new Promise((resolve) => video.addEventListener('progress', resolve));
```
cjpillsbury, line 128: nitpick(non-blocking): Similar asks for this test.

cjpillsbury (approved):
A bunch of non-blocking asks (the most involved being around the added playback-core tests, which I'm pumped you added). 🚀

packages/mux-elements-codemod/package.json
luwes: it's a rule that is not needed in that pkg, I feel like if it's not done when you notice small things like that it will never get done so did it "out of scope"

packages/mux-audio/package.json
luwes: yes, this was required because the older plugin did not support the inline `type` keyword which we mentioned yesterday. e.g. `import { type ValueOf } from './types'`

packages/playback-core/src/index.ts
luwes: yes, would love to get this done as a fast follow if possible

packages/playback-core/src/index.ts
luwes: I like that, any chance we might rename it to `destroy()`?
hls.js and mux-embed both have the `destroy` naming

packages/playback-core/test/player.test.js
luwes: this actually has to do with the order of error event listeners being added to the media element because the playback core stops propagation of the native error event. I'll check if we can make this better.

packages/playback-core/test/player.test.js
luwes: good feedback will update in a follow up pr

PR #427, opened 2022-09-28
feat: simple focus trap/passthrough for dialog

gkatsev:


gkatsev:
This will be useful for when we want to do the focus trapping in fullscreen or inside the dialog

PR #430, opened 2022-09-30
fix: playback-id or src nil value

luwes:
allow removing a src with a nil value

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -786,6 +786,9 @@ class MuxPlayerElement extends VideoApiElement {
    * Set the beacon collection domain. Used by Mux Data.
    */
   set beaconCollectionDomain(val: string | undefined) {
+    // don't cause an infinite loop
+    if (val === this.beaconCollectionDomain) return;
+
     if (val) {
       this.setAttribute(MuxVideoAttributes.BEACON_COLLECTION_DOMAIN, val);
     } else {
@@ -840,6 +843,9 @@ class MuxPlayerElement extends VideoApiElement {
    * Set Mux asset custom domain.
    */
   set customDomain(val: string | undefined) {
+    // dont' cause an infinite loop
+    if (val === this.customDomain) return;
+
     if (val) {
       this.setAttribute(MuxVideoAttributes.CUSTOM_DOMAIN, val);
     } else {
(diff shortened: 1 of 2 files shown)


packages/mux-player/src/index.ts
```diff
@@ -786,9 +786,6 @@ class MuxPlayerElement extends VideoApiElement {
    * Set the beacon collection domain. Used by Mux Data.
    */
   set beaconCollectionDomain(val: string | undefined) {
-    // don't cause an infinite loop
-    if (val === this.beaconCollectionDomain) return;
```
luwes, line 790: mux-player doesn't have this issue, the infinite loop is related to CustomVideoElement.js

packages/mux-player/src/index.ts
```diff
@@ -835,9 +840,6 @@ class MuxPlayerElement extends VideoApiElement {
    * Set Mux asset custom domain.
    */
   set customDomain(val: string | undefined) {
-    // dont' cause an infinite loop
-    if (val === this.customDomain) return;
```
luwes, line 839: again no need here 
> mux-player doesn't have this issue, the infinite loop is related to CustomVideoElement.js

packages/mux-player/src/index.ts
```diff
@@ -786,9 +786,6 @@ class MuxPlayerElement extends VideoApiElement {
    * Set the beacon collection domain. Used by Mux Data.
    */
   set beaconCollectionDomain(val: string | undefined) {
-    // don't cause an infinite loop
```
cjpillsbury, line 789: question(blocking): While this may not cause an infinite loop, doesn't this mean we're triggering re-renders unnecessarily (aka when no change occurred) now?

packages/mux-player/src/index.ts
```diff
@@ -808,20 +805,28 @@ class MuxPlayerElement extends VideoApiElement {
   /**
    * Set Mux asset playback id.
    */
-  set playbackId(val: string | undefined) {
-    this.setAttribute(MuxVideoAttributes.PLAYBACK_ID, `${val}`);
+  set playbackId(val) {
```
cjpillsbury, line 808: 👌

cjpillsbury (changes requested):
since out of scope, have a blocking question for the removed identity checks. Otherwise, 👌

packages/mux-player/src/index.ts
luwes: yes that's a valid point, just a small thought. 

should we gate that in the property or in the attributeChangedCallback? 
makes more sense in the callback I think because attributes can always still be set and callback runs each time on a set

cjpillsbury: agree that feels like the "right" place to do the check, but I think that should be a separate PR since it effects all attributes and may introduce a regression, since there are potential "externally set" state vs. "internally set" state considerations there.

cjpillsbury (approved):
awesome.

PR #431, opened 2022-09-30
tools: improve turborepo filter ability

luwes:
this enables filtering by package for linting or testing or building.

before we couldn't just do `yarn test --filter @mux/mux-video` because the `--filter` arg was an addition to `--filter @mux/*` and it would run testing on all element packages.

required was fixing the dependencies in turbo.json so they can properly be filtered etc.
this change removes the lines 
- `"pretest": "yarn build:packages --filter @mux/test-esm-exports",`
- `"predev": "yarn build:packages",`

but Turbo knows it needs to build `@mux/test-esm-exports` for example before running test on a dependent.

Related https://github.com/vercel/turborepo/issues/460


cjpillsbury (approved):
should speed up our github actions as well! 😎

gkatsev (approved):

PR #435, opened 2022-10-03
fix: preload property in mux-player

luwes:
fixes an issue in mux-player where `.preload` was an empty string as default which maps to the `auto` value.

### DIFF
--- a/packages/mux-player/test/player.test.js
+++ b/packages/mux-player/test/player.test.js
@@ -201,9 +201,10 @@ describe('<mux-player>', () => {
     const player = await fixture(`<mux-player
     ></mux-player>`);
     const muxVideo = player.media;
+    const defaultPreload = document.createElement('video').preload;
 
-    assert.equal(player.preload, 'metadata', "Chrome's player default preload is metadata");
-    assert.equal(muxVideo.preload, 'metadata', "Chrome's muxVideo default preload is metadata");
+    assert.equal(player.preload, defaultPreload, `player default preload is ${defaultPreload}`);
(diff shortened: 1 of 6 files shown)


packages/mux-player/test/player.test.js
```diff
@@ -197,6 +197,24 @@ describe('<mux-player>', () => {
     assert.equal(muxVideo.preload, 'auto', `has preload enabled`);
   });
 
+  it('preload behaves like expected', async function () {
+    const player = await fixture(`<mux-player
+    ></mux-player>`);
+    const muxVideo = player.media;
+
+    assert.equal(player.preload, 'metadata', "Chrome's player default preload is metadata");
```
gkatsev, line 205: wouldn't this fail if we run this in not chrome?

luwes: yes for sure but so far we don't test other browsers

gkatsev: I guess, should we? And if so, maybe better to make this prepped for that?

Maybe testing it's equal to `document.createElement('video').preload`?

cjpillsbury: question(non-blocking): Yeah I'd rather avoid writing tests for a specific browser if possible. For all browsers, it should never `="none"`, correct? Should we test that condition for now?

luwes: ah yes good idea. it could be `none` theoretically, like mobile safari on 3G maybe does that

gkatsev: Mobile safari does preload none, but I think they keep the value as `auto`.

gkatsev (approved):

PR #436, opened 2022-10-03
fix: turbo build clean bug

luwes:
there was a bug in building where the "build" task was dependent on "clean" and "build:types" and "dev" was dependent on the "clean" task and the "^build" task. it comes down to some `dist` folders were being cleaned while another script was trying to `build` or `build:types` to the same folder.

I believe we can finally get rid of the auto clean because the build is now done in order from "dependency" to "dependents".


cjpillsbury (approved):
lgtm

PR #437, opened 2022-10-03
fix: disable buttons behind dialog

gkatsev:
Alternative to https://github.com/muxinc/elements/pull/427

### DIFF
--- a/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
+++ b/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
@@ -16,6 +16,7 @@ const MediaChromeSizes = {
 
 type ThemeMuxTemplateProps = {
   hasSrc: boolean;
+  isDialogOpen: boolean;
   streamType: string | null;
   nohotkeys?: boolean;
   hotkeys: string | null;
@@ -37,6 +38,8 @@ export default class MediaThemeMux extends MediaTheme {
   static get observedAttributes() {
     return [
       'audio',
+      'has-open-dialog',
+      'has-src',
       'nohotkeys',
       'hotkeys',
       'stream-type',
@@ -59,6 +62,7 @@ export default class MediaThemeMux extends MediaTheme {
   render() {
     const props: ThemeMuxTemplateProps = {
       hasSrc: this.hasAttribute('has-src'),
+      isDialogOpen: this.hasAttribute('has-open-dialog'),
       audio: this.hasAttribute('audio'),
       nohotkeys: this.hasAttribute('nohotkeys'),
       hotkeys: this.getAttribute('hotkeys'),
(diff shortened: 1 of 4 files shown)


packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -75,7 +79,7 @@ export default class MediaThemeMux extends MediaTheme {
     };
 
     // if we don't have a src, we want to disable all hotkeys
-    props.nohotkeys = props.hasSrc ? props.nohotkeys : true;
+    props.nohotkeys = !props.hasSrc || props.isDialogOpen ? true : props.nohotkeys;
```
cjpillsbury, line 82: nitpick(non-blocking): `props.nohotkeys || !props.hasSrc || props.isDialogOpen`

packages/mux-player/src/template.ts
```diff
@@ -29,6 +29,7 @@ const isLiveOrDVR = (props: MuxTemplateProps) =>
 export const content = (props: MuxTemplateProps) => html`
   <${unsafeStatic(castThemeName(props.theme) ?? 'media-theme-mux')}
     has-src="${props.hasSrc}"
+    has-open-dialog="${props.isDialogOpen}"
```
cjpillsbury, line 32: thought(non-blocking): We'll want to be thinking about how to handle these scenarios in a theme-agnostic way going forward.

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
gkatsev: In the case where a source isn't set, or the dialog is open, we want to override the user-provided value of `nohotkeys`.

cjpillsbury: only if it's `false`, no?

cjpillsbury (approved):
approved with a couple of non-blocking comments/callouts.

packages/mux-player/src/template.ts
gkatsev: Yeah, as I mentioned, https://github.com/muxinc/elements/pull/427 is theoretically a more correct fix and would be theme agnostic. However, that change is a bit more complicated. There's potentially a performance issue but there's also an issue around binding the key against the document and trying to find all focusable elements and mux-player, which might not be possible depending on how mux-player is included on the page.

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
gkatsev: Ah, right, if nohotkeys is true, it probably doesn't matter what the others are set to. Was overthinking things.
Yeah, your suggestion is better.

PR #439, opened 2022-10-04
Docs/mux player refs cleanup

cjpillsbury:
Updating docs to be:

- exhaustive
    - This includes adding commented out rows for undocumented/unofficial/internal properties/methods/etc. This will give us a cleaner process for any API changes to our components so we don't accidentally forget to add documentation.
- accurate
    - This includes adding additional details and columns
- maintainable
    - This includes deferring detailed descriptions to other sources, such as MDN.
    - This includes not prioritizing the order for "importance" and instead relying on where attributes, properties, etc. are defined in the codebase.

Given the formatting rules for markdown and the changes in order, reviewing is most likely easiest by previewing the markdown itself or using the "Rich Diff" view.

- [x] mux-player
- [x] mux-player-react


packages/mux-player/src/index.ts
```diff
@@ -761,6 +761,17 @@ class MuxPlayerElement extends VideoApiElement {
...
 
+  /**
+   * Set the default hidden captions flag.
+   */
+  set defaultHiddenCaptions(val: boolean | undefined) {
+    if (!val) {
+      this.removeAttribute(PlayerAttributes.DEFAULT_HIDDEN_CAPTIONS);
+    } else {
+      this.setAttribute(PlayerAttributes.DEFAULT_HIDDEN_CAPTIONS, '');
+    }
```
luwes, line 772: 👍

packages/mux-player/REFERENCE.md
```diff
@@ -1,110 +1,144 @@
...
-| `nohotkeys`                   | `boolean`                                                           | Toggles keyboard shortcut (hot keys) support when focus in inside the player                                                                                                                                                                                                                                                                                                                     | `false`       |
-| `hotkeys`                     | `string`                                                            | Offers a way to disable certain keyboard shortcuts                                                                                                                                                                                                                                                                                                                                               | N/A           |
```
luwes, line 36: should we start using https://prettier.io/docs/en/ignore.html#range-ignore for markdown tables?
for Media Chrome too. there's no way to see the real changes anymore w/ Prettier on

cjpillsbury: Maybe. Let's discuss this in the next eng sync. Ideally I'd like these to be generated from in-code docs/types if possible or some other way of treating this as "data" instead of our current strategy (too prone to human error, adds extra work that robots could do, etc.).

cjpillsbury: Fwiw, for this particular case, the order changed and some of the table column counts changed anyway, so the diff would still not be especially handy. Not sure if you caught it, but the description for the PR suggests some easier ways to review.

packages/mux-player/src/video-api.ts
```diff
@@ -370,12 +372,14 @@ class VideoApiElement extends globalThis.HTMLElement implements VideoApiElement
...
 
   set playsInline(val) {
-    if (val) {
-      this.setAttribute(AllowedVideoAttributes.PLAYSINLINE, '');
-    } else {
-      // Remove boolean attribute if false, 0, '', null, undefined.
-      this.removeAttribute(AllowedVideoAttributes.PLAYSINLINE);
-    }
+    logger.error('playsInline is set to true by default and is not currently supported as a setter.');
+    return;
```
luwes, line 376: what's the use of the return?

luwes (approved):
lgtm

PR #440, opened 2022-10-04
fix: potential duplicate rendering of template()

luwes:
there was a report of a duplicate render which could happen potentially if the template strings cache would lose an entry while the mux-player element and its rendered shadowRoot stayed intact. the next render on the shadowRoot would render an additional result of the `template()` function. 

fix is to clear the shadow dom if this happens.


packages/mux-player/src/index.ts
```diff
@@ -251,18 +251,22 @@ class MuxPlayerElement extends VideoApiElement {
...
-        [StreamTypes.LIVE, StreamTypes.LL_LIVE, StreamTypes.DVR, StreamTypes.LL_DVR].includes(this.streamType as any)
-      ) {
-        // playback core should handle the seek to live on first play
-        if (this.hasPlayed) {
-          seekToLive(this);
+    this.addEventListener(
+      'mediaplayrequest',
+      (event) => {
+        if (
+          (event.composedPath()[0] as Element)?.localName === 'media-play-button' &&
```
gkatsev, line 258: TIL `composedPath`

gkatsev (approved):

axpro:
I'm using the react player in a next.js app and when building for production the player would duplicate when changing the size of the player container with a state update. This seems to fix that.
Is there an estimate of time when this will be released? 

Thank you

luwes:
@axpro thanks for reporting, we are starting on fixing this bug. any chance you have a reproduction of this in Codesandbox or similar? I can't reproduce it so far.

luwes:
until we know for sure what this causes we can't merge this in because it would be kind of a bandaid solution.

normally the player should re-use the 1st core dom structure that is updated repeatedly but for some reason this condition is true making it append the core structure again
https://github.com/muxinc/elements/blob/main/packages/mux-player/src/html.ts#L134-L134

if the element is cleared on a subsequent render the whole internal DOM structure is renewed which is less efficient and the player state will be reset. ideally the core structure should never be cleared and re-rendered.

axpro:
In your example codesandbox click play, unmute and then trigger render. The video will start again.

I'm not sure it matters but I'm on Mac Ventura and I'm using Chrome 106.0.5249.119.

On Safari 16.1 (18614.2.9.1.10) that particular playbackId gives me "Media failed to decode". This id works (a4nOgmxGWg6gULfcBbAa00gXyfcwPnAFldF8RdsNyk8M) and the same thing happens, it starts playing from the beginning.

luwes:
@axpro ah I see it's not persisting the player state. that's actually another thing altogether. 
this PR would provide a solution for when the player was rendered twice at the same time one below the other.

axpro:
It's very hard for me to pinpoint exactly what causes the duplicate player bug and I can't replicate it consistently. What I did notice is that it only happens in the production builds and that it happens more often when I interact with the player (play/pause + mute/unmute) before changing the state. The behavior seems to indicate that some sort of race condition is in play somewhere.

luwes:
closing until this comes up again

PR #441, opened 2022-10-04
fix: use CSS.registerProperty on vars to declare them as colors

gkatsev:
Declaring the properties as colors means that the browsers will properly reject invalid values.

### DIFF
--- a/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
+++ b/packages/mux-player/src/media-theme-mux/media-theme-mux.ts
@@ -55,6 +55,37 @@ export default class MediaThemeMux extends MediaTheme {
     ];
   }
 
+  constructor() {
+    super();
+
+    this.#setupCSSProperties();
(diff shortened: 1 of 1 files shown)


packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -55,6 +55,37 @@ export default class MediaThemeMux extends MediaTheme {
     ];
   }
 
+  constructor() {
+    super();
+
+    this.#setupCSSProperties();
```
gkatsev, line 61: I put this into media-theme-mux because it felt the right place for them. Since, these are theme properties, the theme should declare them.
Had to use the JS API because using the `@property` declaration was stuck inside the shadow DOM and thus not available to users outside of mux-player. The CSS API is global and thus will be present.

luwes (approved):
cool lgtm!

PR #442, opened 2022-10-04
chore: graduate mux-player and mux-player-react

gkatsev:
To release mux-player and mux-player-react as no longer beta, we need to graduate them as per https://github.com/lerna/lerna/blob/main/commands/version/README.md#--conventional-graduate.

Since we run things in CI, we'll need a follow-up PR to remove this flag afterwards.


cjpillsbury (approved):
👌

PR #443, opened 2022-10-04
chore: package versioning doesn't need --conventional-graduate anymore 

gkatsev:
Followup from https://github.com/muxinc/elements/pull/442


cjpillsbury (approved):
👌

PR #447, opened 2022-10-06
feat: add ability to unset poster

luwes:
- this change adds a feature so users can remove the derived poster from the playback-id by setting the poster property or attribute to an empty string. e.g. `player.poster = ''`
- upgrade Media Chrome to v0.13.0


packages/mux-player/src/index.ts
```diff
@@ -593,6 +604,78 @@ class MuxPlayerElement extends VideoApiElement {
...
+  /**
+   * Set the string that reflects the src HTML attribute, which contains the URL of a media resource to use.
+   */
+  set src(val) {
+    if (val) {
+      this.setAttribute(VideoAttributes.SRC, val);
+    } else {
+      this.removeAttribute(VideoAttributes.SRC);
+    }
+  }
```
luwes, line 647: sorry I moved these blocks, they are moved more to the top because of importance. didn't change logic

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -138,8 +138,8 @@ export default class MediaThemeMux extends MediaTheme {
           <slot name="media" slot="media"></slot>
           <media-poster-image
```
gkatsev, line 139: did we want to expose this with a `poster` part?

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
```diff
@@ -138,8 +138,8 @@ export default class MediaThemeMux extends MediaTheme {
...
             slot="poster"
-            src="${props.poster}"
-            placeholder-src="${props.placeholder}"
+            src="${props.poster ?? false}"
+            placeholder-src="${props.placeholder ?? false}"
```
gkatsev, line 142: these are necessary because `poster` and `placeholder` props may not always be booleans and that might cause issues when toggling the values?

packages/mux-player/test/player.test.js
```diff
@@ -216,37 +188,75 @@ describe('<mux-player>', () => {
...
     );
     assert.equal(
-      muxVideo.getAttribute('poster'),
+      mediaPosterImage.getAttribute('src'),
       'https://image.mux.com/xLGf7y8cRquv7QXoDB02zEe6centwKfVmUOiPSY02JhCE/thumbnail.jpg?time=1',
       `has poster attr added`
     );
+  });
+
+  it('poster can be unset with an empty string', async function () {
```
gkatsev, line 222: I thought it gets unset with `null`?

packages/mux-player/test/player.test.js
```diff
@@ -216,37 +188,75 @@ describe('<mux-player>', () => {
...
     assert.equal(
-      muxVideo.poster,
+      player.poster,
+      'https://image.mux.com/DS00Spx1CV902MCtPj5WknGlR102V5HFkDe/thumbnail.jpg',
+      'uses the derived poster if no poster attribute is present'
+    );
+
+    player.poster = '';
+    assert.equal(player.poster, '');
+    assert(!mediaPosterImage.hasAttribute('src'), 'media-poster-image does have a poster attribute');
```
gkatsev, line 245: Should the comment be `does *not* have a poster`?

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
luwes: I'm fine with that, will add it

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
luwes: it's our template logic that works like this. when setting an explicit `false` it will remove the attribute.
otherwise it will still set an attribute value with null or undefined

packages/mux-player/test/player.test.js
luwes: yes I adapted this style from @cjpillsbury, the message is shown when the test fails so it makes sense to say what happened but yes it might be all mixed up now because I didn't used to write it that way with other tests 🤷

packages/mux-player/test/player.test.js
luwes: we can't use `null` because it means absence of poster attribute which means the poster will be derived from the playback-id

gkatsev: aah, right

gkatsev:
When testing this, I'm seeing that `media-poster-image`'s `img` element has as `null` src:
```html
<img aria-hidden="true" id="image" src="null">
```
Though, media-poster-image itself doesn't have a src
```
<media-poster-image slot="poster" part="poster"></media-poster-image>
```
image

and thus shows a broken image icon in the top left corner.
image

not exactly sure why it's coming up this way, seems like things are configured as expected 🤔

gkatsev (approved):
Looks like my last comment about the broken icon and null src is a media-chrome issue. This PR LGTM and can probably go in separately from a media-chrome fix.

gkatsev:
Fix is here https://github.com/muxinc/media-chrome/pull/328

packages/mux-player/src/media-theme-mux/media-theme-mux.ts
cjpillsbury: upvote! 😎

packages/mux-player/src/template.ts
```diff
@@ -28,39 +23,29 @@ const isLiveOrDVR = (props: MuxTemplateProps) =>
 
 export const content = (props: MuxTemplateProps) => html`
   <${unsafeStatic(castThemeName(props.theme) ?? 'media-theme-mux')}
-    has-src="${props.hasSrc}"
-    has-open-dialog="${props.isDialogOpen}"
+    has-src="${props.hasSrc ?? false}"
```
cjpillsbury, line 26: idea(non-blocking): 😎 eventually we may want to figure out a way to hide these requirements "under the hood" in our `html()` impl. Would also reduce redundancy/inconsistency/chance of error.

packages/mux-player/src/video-api.ts
```diff
@@ -22,8 +22,6 @@ const AllowedVideoAttributes = {
   LOOP: 'loop',
   MUTED: 'muted',
   PLAYSINLINE: 'playsinline',
-  SRC: 'src',
```
cjpillsbury, line 25: idea(non-blocking): I'm wondering if we should just export this to avoid the proliferation of map objects and related.

packages/mux-player/test/player.test.js
cjpillsbury: yeah the "does" may be the confusing part. For a light lift on rephrasing standardization, we could phrase these as "should"/"shouldn't" statements directly related to the assertion. e.g. for this case:

> media-poster-image shouldn't have a `src` attribute set.

cjpillsbury (changes requested):
Overall LGTM. Just make sure you:
docs(non-blocking): Update REFERENCE.md to describe other props in default.
Also:
thought(non-blocking): Noticed we moved poster's default/derived logic to the getter but not `src`. Your new approach feels like a better pattern overall if we can conform to it.

luwes:
@gkatsev verified your media chrome PR fixed that issue. upgraded media chrome in this PR

decepulis:
I'm so glad this is a thing. I've been using `poster="www.google.com"` to simulate this 😂

PR #454, opened 2022-10-07
chore: add scrollable options in nextjs/MuxPlayer

gkatsev:
Sets the height of the options section to be 100vh minus the player height and everything else.

Player height defaults to 450 and will grab the first height that's set when loaded. Will update this if the player size is changed via the "width cutoffs for reponsinve player chrome/ui" option.

### DIFF
--- a/examples/nextjs-with-typescript/components/renderers.tsx
+++ b/examples/nextjs-with-typescript/components/renderers.tsx
@@ -6,11 +6,11 @@ export const toWordsFromCamel = (string: string) => {
   return `${first}${rest.replace(/[A-Z]/g, (match) => ` ${match}`)}`
 };
 
-export const BooleanRenderer = ({ 
-  name, 
-  value, 
-  label, 
-  onChange 
+export const BooleanRenderer = ({
+  name,
+  value,
+  label,
+  onChange
 }: { name: string, value: boolean | undefined, label?: string, onChange: (obj: any) => void }) => {
   const labelStr = label ?? toWordsFromCamel(name);
   return (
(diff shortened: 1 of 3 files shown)


examples/nextjs-with-typescript/styles.css
```diff
@@ -105,6 +105,11 @@ audio {
   margin-bottom: 1rem;
 }
 
+.options {
+  overflow: scroll;
```
luwes, line 109:
```suggestion
  overflow: auto;
```

otherwise it shows even empty scroll bars 😢

gkatsev: I guess, though, it's basically always going to scroll. It's scrolling for me and I have a 4K monitor and I run my windows full height.

examples/nextjs-with-typescript/pages/MuxPlayer.tsx
```diff
@@ -268,6 +259,24 @@ const getControlCustomizationCSSVars = (state) => {
...
+
+const getUrl = ({ req, resolvedUrl }) => {
+  const { headers } = req;
+  const refererUrl = headers.referer && new URL(headers.referer);
+  const baseUrlHost = headers.host.toLowerCase();
+  const refererHost = refererUrl?.host?.toLowerCase();
+
+
+  if (refererHost === baseUrlHost && headers["sec-fetch-site"] === "same-origin") {
+    return new URL(refererUrl?.origin ? refererUrl.origin + resolvedUrl : './');
```
gkatsev, line 272: the main change in this method is making sure that `resolvedUrl` is always part of the url, because otherwise, there are cases where `/MuxPlayer` doesn't make it in the url

examples/nextjs-with-typescript/components/renderers.tsx
```diff
@@ -169,19 +169,19 @@ export const EnumMultiSelectRenderer = ({
             id={`${name}-control`}
             multiple
             size={values.length}
+            defaultValue={value ?? []}
```
gkatsev, line 172: `defaultValue` should be an array of the values we want selected.

luwes (approved):

PR #455, opened 2022-10-07
fix: also add media-object-fit/position to video

luwes:
this adds the CSS var to the underlying video element in addition to the media poster image for easy DX.
no need to add a separate CSS part selector.

tagging this commit as a fix intentionally.

### DIFF
--- a/packages/mux-video/src/CustomVideoElement.js
+++ b/packages/mux-video/src/CustomVideoElement.js
@@ -60,6 +60,8 @@ template.innerHTML = `
     max-height: 100%;
     min-width: 100%;
     min-height: 100%;
+    object-fit: var(--media-object-fit);
+    object-position: var(--media-object-position);
   }
 </style>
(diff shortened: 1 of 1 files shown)


gkatsev (approved):

packages/mux-player/src/styles.css
```diff
@@ -8,6 +8,11 @@
   display: none;
 }
 
+mux-video::part(video) {
```
gkatsev, line 11: question(non-blocking): should this go into `mux-video`?

luwes: right maybe.

and I also just saw `object-fit: fill` is the default CSS value but media-poster-image was already set to cover so best to have those match right?

luwes: scratch that `object-fit: contain;` is the default for video in browsers

SCR-20221007-jmo

PR #465, opened 2022-10-14
fix: lazy placeholder lower specificity styles

decepulis:
Currently, placeholder styles are set via `style={}` attribute, making them inline styles. Therefore, if you use a class (e.g., a tailwind class of `aspect-1`), it will affect the loading player, but not the placeholder. (On account of the tailwind class having lower specificity than inline styles)

My awkward solution? An inline `<style>` block with really unique selector names (`data-mux-player-react-placeholder` and `data-mux-player-react-placeholder-overlay`) to prevent global conflicts.

Not ideal, but imo better than the alternative.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,16 @@
 
 Elements are written to make it easy to work in a variety of web frameworks. If we don't (yet) have a framework-specific "flavor" of an element, we always provide a pure [Web Component](https://developer.mozilla.org/en-US/docs/Web/Web_Components) version that work with HTML and can be used by most modern web frameworks and web application environments.
 
-| Name                                          | Description                                 | Framework     | Official Release Status |
-| --------------------------------------------- | ------------------------------------------- | ------------- | ----------------------- |
-| [mux-video](packages/mux-video)               | A Mux-flavored HTML5 video element          | Web Component | Public Beta             |
-| [mux-audio](packages/mux-audio)               | A Mux-flavored HTML5 audio element          | Web Component | Public Beta             |
-| [mux-video-react](packages/mux-video-react)   | A Mux-flavored React video component        | React         | Public Beta             |
-| [mux-player](packages/mux-player)             | A Mux-flavored HTML5 media player component | Web Component | Public Beta             |
-| [mux-player-react](packages/mux-player-react) | A Mux-flavored React media player component | React         | Public Beta             |
+| Name                                              | Description                                   | Framework     | Official Release Status |
+| ------------------------------------------------- | --------------------------------------------- | ------------- | ----------------------- |
+| [mux-video](packages/mux-video)                   | A Mux-flavored HTML5 video element            | Web Component | Public Beta             |
+| [mux-video-react](packages/mux-video-react)       | A Mux-flavored React video component          | React         | Public Beta             |
+| [mux-audio](packages/mux-audio)                   | A Mux-flavored HTML5 audio element            | Web Component | Public Beta             |
+| [mux-audio-react](packages/mux-audio-react)       | A Mux-flavored HTML5 audio component          | React         | Public Beta             |
+| [mux-player](packages/mux-player)                 | A Mux-flavored HTML5 media player component   | Web Component | 1.x.x 🎉                |
+| [mux-player-react](packages/mux-player-react)     | A Mux-flavored React media player component   | React         | 1.x.x 🎉                |
+| [mux-uploader](packages/mux-uploader)             | A Mux-flavored HTML5 media uploader component | Web Component | Public Beta             |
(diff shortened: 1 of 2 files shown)


packages/mux-player-react/src/lazy.tsx
```diff
@@ -52,42 +44,44 @@ const Fallback = (props: FallbackProps) => {
...
+        // we need to make sure that the placeholder isn't interactive and its player chrome in doesn't get rendered
+        nohotkeys
+        aria-hidden
+        tabIndex={-1}
+      >
+        <div data-mux-player-react-lazy-placeholder-overlay />
+      </mux-player>
+      <style>{
+        /* css */ `
+        [data-mux-player-react-lazy-placeholder]:not([bump-specificity]) {
```
decepulis, line 64: `mux-player:not([audio])` was taking precedence over this rule in that brief moment of "the bundle is loaded but suspense hasn't unloaded the placeholder". `:not([bump-specificty])` is my way of matching that specificity so that these styles hold on

packages/mux-player-react/src/lazy.tsx
```diff
@@ -52,42 +44,44 @@ const Fallback = (props: FallbackProps) => {
...
+          display: block;
+          background-color: var(--media-background-color, #000);
+          width: 100%;
+          position: relative;
+          ${placeholder ? `background-image: url(${placeholder});` : ''}
+          background-repeat: no-repeat;
+          background-size: var(--media-object-fit, contain);
+          background-position: var(--media-object-position, 50% 50%);
+          --controls: none;
+          --controls-backdrop-color: rgba(0, 0, 0, 0.6);
```
decepulis, line 75: By manually specifying `--controls-backdrop-color` here, we opt out of the "if controls is none, make backdrop color none as well" behavior. We need to do this because, in that brief moment of "the bundle is loaded but suspense hasn't swapped it out yet", our `--controls-backdrop-color` was getting set to `none`, causing a bright flash.

packages/mux-player-react/src/lazy.tsx
gkatsev: what about `mux-player[data-mux-player-react-lazy-placeholder]`? Should be same specificity without using unused attributes.

packages/mux-player-react/src/lazy.tsx
luwes: does this still work if a user changes the `--controls-backdrop-color` CSS var from outside?
other color than `rgba(0, 0, 0, 0.6)`

decepulis: A user setting `--controls-backdrop-color` should overwrite this declaration, affecting both mux-player and the overlay within, so long as their declaration has higher specificity!

packages/mux-player-react/src/lazy.tsx
luwes: does @gkatsev solution work @decepulis? that seems a bit more elegant code wise

packages/mux-player-react/src/lazy.tsx
luwes: cool thanks

packages/mux-player-react/src/lazy.tsx
decepulis: Great timing. I _just_ finished testing the solution. It works and it's pushed!

luwes (approved):
👍

packages/mux-player-react/src/lazy.tsx
```diff
@@ -52,42 +44,44 @@ const Fallback = (props: FallbackProps) => {
...
+        }
+        mux-player [data-mux-player-react-lazy-placeholder-overlay] {
+          position: absolute;
+          inset: 0;
```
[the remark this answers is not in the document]
decepulis: Inset is shorthand for top/bottom/left/right. Since it's supported slightly better than aspect-ratio, I figured it's safe to use here.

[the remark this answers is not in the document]
gkatsev: shorthand to make the overlay the same size as its container.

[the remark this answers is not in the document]
decepulis: Discovering it changed my _life_. I used to hate typing those four lines 😭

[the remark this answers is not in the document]
gkatsev: there's some nice shorthands nowadays.
Like block and inline props like margin-block and margin-inline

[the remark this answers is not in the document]
luwes: wow, didn't know inset was for replacing those lines. cool!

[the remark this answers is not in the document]
gkatsev: also, just confirmed that `inset-inline` and `inset-block` also exist 🎉

PR #468, opened 2022-10-17
feat(playback-core): Add CMCD support (mutative).

cjpillsbury:
⚠️ POC - Will likely be refactored before final merge. Pushing for demo/sharing purposes.


packages/playback-core/src/index.ts
```diff
@@ -30,6 +30,16 @@ export const generatePlayerInitTime = () => {
   return mux.utils.now();
 };
 
+/** @TODO REPLACE WITH mux.publicUtils.generateUUID() (CJP) */
+export const generateUUID = function () {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
```
gkatsev, line 35: should we use `crypto.randomUUID()` when available?

packages/playback-core/src/index.ts
```diff
@@ -47,6 +57,12 @@ export const initialize = (
 ) => {
   // Automatically tear down previously initialized mux data & hls instance if it exists.
   teardown(mediaEl, core);
+  // NOTE: metadata should never be nullish/nil. Adding here for type safety due to current type defs.
+  const { metadata = {}, playbackId, src } = props;
+  const { view_session_id = generateUUID(), video_id = playbackId || src } = metadata;
```
luwes, line 62: a session shouldn't change just because the user changes video id right?
calling `generateUUID()` at the top of the file seems more accurate to me

gkatsev: The spec is a bit ambiguous, but it does seem like doing it per load is more helpful.
Spec says:

> A playback session typically ties together segments belonging to a single media asset.

packages/playback-core/src/index.ts
cjpillsbury: nah we should use `mux-embed`'s `generateUUID` for consistency imo (per callout). That said, we may want `mux-embed` to use `crypto.randomUUID()` under the hood when available if it isn't already.

packages/playback-core/src/index.ts
luwes: oh ok, my mind went right away to browser session. if it's clearly in the spec all good

packages/playback-core/src/index.ts
gkatsev: Missed that TODO, but yeah, mux-embed should use that API if available.

packages/playback-core/src/index.ts
```diff
@@ -30,12 +30,34 @@ export const generatePlayerInitTime = () => {
...
 
+const toPlaybackIdFromParameterized = (playbackIdWithParams: string | undefined) => {
+  if (!playbackIdWithParams) return undefined;
+  const [playbackId] = playbackIdWithParams.split('?');
+  // `|| undefined` is here to handle potential invalid cases
+  return playbackId || undefined;
+};
+
+const toPlaybackIdFromSrc = (src: string | undefined) => {
+  if (!src || !src.startsWith('https://stream.')) return undefined;
```
gkatsev, line 49: what if a custom domain is used?

gkatsev: Should this use `isMuxVideoSrc`?

cjpillsbury: Re: custom domains - I'm pretty confident those *should* be fine based on how they work, but I should confirm that there won't be exceptions. (tl;dr - custom video domains are still "ours", aka not a proxy, just a CNAME mapping)
Re: `isMuxVideoSrc` - This function isn't called directly, but instead used by `toVideoId`, which does check `isMuxVideoSrc` (which includes custom domains, fwiw)

luwes: I was a little confused too. custom domains always still have the `https://stream.` prefix then, correct?

gkatsev: In a conversation elsewhere, it sounds like that's a yes.

cjpillsbury: yessir(s)!

packages/mux-player/src/index.ts
```diff
@@ -80,7 +80,7 @@ const PlayerAttributes = {
 };
 
 function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
-  return {
+  const props = {
```
cjpillsbury, line 83: NOTE: Intentionally committing this, since I frequently have to refactor to add e.g. console.logs or similar in this function.

gkatsev: I've done that a bunch as well, so, 👍🏻

packages/playback-core/package.json
```diff
@@ -49,7 +49,7 @@
   },
   "dependencies": {
     "hls.js": "1.2.3",
-    "mux-embed": "^4.7.0"
+    "mux-embed": "^4.14.0"
```
cjpillsbury, line 52: NOTE: A version bump was necessary to get a version of mux-embed with an exposed `generateUUID` in its public `utils`. Upgrading to latest version, bc why not.

packages/playback-core/src/index.ts
```diff
@@ -126,11 +157,19 @@ export const setupHls = (
       liveDurationInfinity: true,
     };
     const streamTypeConfig = getStreamTypeConfig(streamType);
+    const cmcd = experimentalCmcd
+      ? {
+          useHeaders: true,
+          sessionId: metadata.view_session_id,
+          contentId: metadata.video_id,
+        }
+      : undefined;
```
cjpillsbury, line 166: NOTE: `hls.js` will get into a weird state if there is an empty object for `cmcd` in its options. We may want to open an issue for this on the hls.js repo, though it is a low priority.

shared/test-esm-exports/package.json
```diff
@@ -9,9 +9,9 @@
   "scripts": {
     "build": "esbuild ./hls.mjs ./mux-embed.mjs --bundle --outdir=dist --out-extension:.js=.js --format=esm"
   },
-  "dependencies": {
-    "hls.js": "1.2.3",
-    "mux-embed": "^4.7.0"
+  "peerDependencies": {
```
cjpillsbury, line 12: NOTE: This change was to make version maintenance easier for `test-esm-exports`. Since this package is, by design, intended to rely on the versions of hls.js & mux-embed used in the "real" packages, declaring these versions as `peerDependencies` of `"*"` should more accurately describe their usage here.

packages/playback-core/src/index.ts
```diff
@@ -47,6 +69,13 @@ export const initialize = (
 ) => {
   // Automatically tear down previously initialized mux data & hls instance if it exists.
   teardown(mediaEl, core);
+  // NOTE: metadata should never be nullish/nil. Adding here for type safety due to current type defs.
+  const { metadata = {} } = props;
+  const { view_session_id = generateUUID() } = metadata;
```
cjpillsbury, line 74: NOTE: Currently, I'm not keying the updates to these values based on `experimentalCmcd`. On the plus side, this avoids unnecessary forking paths and provides more predictable info in Mux Data for Mux Player, regardless of `CMCD` usage. On the minus side, this does mean we're "doing slightly more work" and "sending slightly more data" than is strictly required.

PR #472, opened 2022-10-17
fix: should have focus outline on dialog's close button

gkatsev:
Firefox doesn't show a focus outline on the dialog due to no `:host-context` support. However, it doesn't seem to be necessary. Seems `:focus-visible` is working well enough.
As far as I can tell, this is the only focus outline issue that we currently have in the player, currently.


cjpillsbury (approved):
🏄‍♂️

PR #474, opened 2022-10-18
feat: add storyboard url getter, don't set poster and storyboard urls if the provided tokens aren't valid

gkatsev:


cjpillsbury (approved):
Looks awesome.

packages/mux-player/src/index.ts
```diff
@@ -676,6 +678,22 @@ class MuxPlayerElement extends VideoApiElement {
     }
   }
 
+  get storyboard() {
+    if (
+      this.playbackId &&
+      !this.audio &&
+      this.streamType &&
```
luwes, line 685: this should also include an OR with the case that `this.streamType` is not set.

aka this getter should return an URL if `this.streamType` is not set.

luwes (changes requested):
return storyboard url if this.streamType is null

packages/mux-player/src/index.ts
gkatsev: good call 👍🏻 I'll add it in tomorrow.

gkatsev: Updated. First, added some tests in https://github.com/muxinc/elements/pull/474/commits/6bd47c8b344f75b569babeb18f9d29e643b9853e to show that it fails and that the fix in https://github.com/muxinc/elements/pull/474/commits/bb0842dc5b0e6fd367b66878a803a84491461d97 would fix the tests

packages/mux-player/src/index.ts
```diff
@@ -678,12 +678,19 @@ class MuxPlayerElement extends VideoApiElement {
...
+   * we aren't an audio player and the stream-type isn't live.
+   */
   get storyboard() {
     if (
       this.playbackId &&
       !this.audio &&
-      this.streamType &&
-      ![StreamTypes.LIVE, StreamTypes.LL_LIVE, StreamTypes.DVR, StreamTypes.LL_DVR].includes(this.streamType as any)
+      (!this.streamType ||
+        (this.streamType &&
```
luwes, line 690: nit (non-blocking): `!this.streamType ||` the right hand of this condition already means `this.streamType` is not null so no need for `(this.streamType && )`

luwes (approved):
👍

PR #475, opened 2022-10-19
feat: conditionally use title for title metadata

gkatsev:
If video title metadata isn't explicitly provided, use the video title attribute instead.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -210,13 +210,10 @@ export const setupMux = (
       debug,
     } = props;
 
-    const metadata = Object.keys(props.metadata).reduce((mdata: Record<string, string>, prop) => {
-      const val = props.metadata[prop];
-      if (val !== '') {
-        mdata[prop] = val;
-      }
-      return mdata;
-    }, {});
+    const metadata = {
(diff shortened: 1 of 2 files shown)


dylanjha (approved):

cjpillsbury (approved):
👍

packages/playback-core/src/index.ts
```diff
@@ -207,10 +207,17 @@ export const setupMux = (
       playerSoftwareName: player_software_name,
       playerSoftwareVersion: player_software_version,
       beaconCollectionDomain,
-      metadata,
       debug,
     } = props;
 
+    const metadata = Object.keys(props.metadata).reduce((mdata, prop) => {
```
cjpillsbury: nitpick(blocking) Just to keep it narrow (not sure if there are valid cases to send empty strings?), we should probably do something more like:

```js
const metadata = { 
  ...props.metadata, 
  video_title: props.metadata.video_title || undefined,
};
```

cjpillsbury (changes requested):
minor nitpick

cjpillsbury (approved):
👌

cjpillsbury (approved):

PR #480, opened 2022-10-24
fix: don't register prop --controls-backdrop-color

gkatsev:
Chromium browsers are the only browsers that currently support @property/CSS.registerProperty. We were registering --controls-backdrop-color as a <color> so that invalid colors would be ignored and the default value would be used. We also wanted the backdrop color to get unset if controls were hidden with --controls:none but if a backdrop color were provided, to use that color. However, this caused Chrome to mark our var() expression as invalid because `none` isn't a color value and not show the provided --controls-backdrop-color. Given that @property is a Chrome-only feature right now and that it's probably fine if invalid value for --controls-backdrop-color turn the backdrop color off, we shouldn't register --controls-backdrop-color as a color.


gkatsev:
The other `--primary-color` and `--secondary-color` properties are still registered and there aren't any issues there afaik because we aren't doing Weird Stuff with them. There's greater utility to leave them registered than unregistering them.

cjpillsbury (approved):
👎 but ok I get it. (approved, but with disappointment)

luwes (approved):

PR #481, opened 2022-10-25
Feat/disable cookies

cjpillsbury:
Adds `disableCookies` and corresponding attr support to Open (Media) Elements and underlying playback-core. This disables cookies usage by mux-embed (See: [Mux Docs](https://docs.mux.com/guides/data/monitor-html5-video-element#disable-cookies))

To test, use the NextJS app's MuxPlayer page
* When `disableCookies` is `false` or unset, open dev tools and confirm that the `muxData` cookie is present.
* When setting `disableCookies` via the UI, choose a media source (e.g. from our pre-defined list in the UI) to load media and start mux data monitoring. Open dev tools and confirm that the `muxData` cookie has not been created.


luwes (approved):
👍 tested

PR #486, opened 2022-10-28
feat: first iteration on player design update

luwes:
- default vertical design is not included yet

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -817,7 +817,7 @@ class MuxPlayerElement extends VideoApiElement {
    * Get the secondary color used by the player.
    */
   get secondaryColor() {
-    return this.getAttribute(PlayerAttributes.SECONDARY_COLOR) ?? undefined;
+    return this.getAttribute(PlayerAttributes.SECONDARY_COLOR) ?? 'rgb(0 0 0 / .65)';
   }
 
   /**
--- a/packages/mux-player/src/media-theme-mux/styles.css
+++ b/packages/mux-player/src/media-theme-mux/styles.css
@@ -154,7 +154,11 @@ media-control-bar :is(media-text-display, mxp-time-display):first-child {
   background-color: var(--media-control-background, rgba(20, 20, 30, 0.7));
 }
 
-media-control-bar[slot='top-chrome']::before {
+media-control-bar[slot='top-chrome'] {
+  min-height: 42px;
+}
+
+:host([title]) media-control-bar[slot='top-chrome']::before {
   content: '';
   position: absolute;
   width: 100%;
--- a/packages/mux-player/src/media-theme-mux/styles.css
+++ b/packages/mux-player/src/media-theme-mux/styles.css
@@ -156,6 +156,7 @@ media-control-bar :is(media-text-display, mxp-time-display):first-child {
 
 media-control-bar[slot='top-chrome'] {
   min-height: 42px;
+  pointer-events: none;
 }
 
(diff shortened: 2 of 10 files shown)


packages/mux-player/src/media-theme-mux/styles.css
```diff
@@ -98,35 +98,34 @@ media-controller {
...
-  bottom: -2px;
-  background: linear-gradient(
-    180deg,
-    transparent,
-    transparent 3px,
-    var(--media-control-background) 3px,
-    var(--media-control-background) 8px,
-    transparent 8px,
-    transparent
-  );
```
luwes, line 118: this was still a left over of when we made the time range bigger for getting more hover area.
it's now built-in Media Chrome so we can remove it here.

packages/mux-player/src/template.ts
```diff
@@ -106,6 +106,7 @@ export const content = (props: MuxTemplateProps) => html`
               }
             }}"
           >
+            <svg viewBox="0 0 8 8"><circle cx="4" cy="4" r="4"></circle></svg>
```
luwes, line 109: switched to an SVG for the dot, it's more consistent size cross-browser

packages/mux-player/src/index.ts
```diff
@@ -817,7 +817,7 @@ class MuxPlayerElement extends VideoApiElement {
    * Get the secondary color used by the player.
    */
   get secondaryColor() {
-    return this.getAttribute(PlayerAttributes.SECONDARY_COLOR) ?? undefined;
+    return this.getAttribute(PlayerAttributes.SECONDARY_COLOR) ?? 'rgb(0 0 0 / .65)';
```
luwes, line 820: users can still set this to an empty string to remove the secondary color

packages/mux-player/src/media-theme-mux/styles.css
```diff
@@ -143,12 +142,26 @@ media-control-bar :is([role='button'], [role='switch'], button) {
...
+  position: relative;
+}
+
 media-controller::part(vertical-layer) {
   transition: background-color 1s;
 }
 
 media-controller:is([media-paused], :not([user-inactive]))::part(vertical-layer) {
-  background-color: var(--controls-backdrop-color, var(--controls, rgba(0, 0, 0, 0.6)));
+  background-color: var(--controls-backdrop-color, var(--controls, transparent));
```
luwes, line 164: the drop background color can also be set back to before if the user wants to

cjpillsbury (approved):
Good call.

packages/mux-player/src/media-theme-mux/styles.css
luwes: this is put back in to be able to have a better experience on mobile having a higher time range bar

cjpillsbury:
Doing some testing on the hosted app. Will add screenshots of callouts as I identify them.

cjpillsbury:
Screen Shot 2022-10-31 at 1 29 03 PM
Live audio only could use a bit of work. 😅

cjpillsbury:
Screen Shot 2022-10-31 at 1 33 22 PM

Live/DVR + Small shouldn't have the media-control-bar play button

cjpillsbury (changes requested):
snapshots for reference/change requests. I'll also do an audit of the code changes.

cjpillsbury:
Screen Shot 2022-11-01 at 6 49 29 AM

Live + Extra Small has a media-time-range

cjpillsbury (approved):
😻

PR #487, opened 2022-11-02
fix: export media loading indicator

richard-grosjean:
We want the ability to hide the media loading indicator using CSS. This pull request adds a part to the media loading indicator and exports said part so that it is available from outside of the shadow tree.


luwes:
@richard-grosjean thanks for this PR! this looks good, we'll discuss with the team and come back soon.
we might want to make this a CSS var only

luwes:
@richard-grosjean we're going to make this a CSS var only initially, it will cover the show / hide case.
I'll leave this PR open until the CSS var is merged in.

thanks for the contribution!

luwes (changes requested):
CSS var only

PR #489, opened 2022-11-02
fix: adds part to mxp-dialog and exposes it

richard-grosjean:
We want a way to hide the dialog box with CSS. This pull requests adds a part to <mxp-dialog> and exposes it.

Long story short, we are running into cases where the error dialog is causing more harm than good. Our case involves using mux videos as backgrounds for hero elements, so we would rather have the videos fail silently rather than having an annoying error message be displayed.

### DIFF
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -45,7 +45,7 @@ export const content = (props: MuxTemplateProps) => html`
     hotkeys="${props.hotKeys ?? false}"
     poster="${props.poster === '' ? false : props.poster ?? false}"
     placeholder="${props.placeholder ?? false}"
-    exportparts="top, center, bottom, layer, media-layer, poster-layer, vertical-layer, centered-layer, gesture-layer, poster, seek-live, play, button, seek-backward, seek-forward, mute, captions, airplay, pip, fullscreen, cast, playback-rate, volume, range, time, display"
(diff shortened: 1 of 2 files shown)


luwes:
same here.
@richard-grosjean thanks for this PR! this looks good, we'll discuss with the team and come back soon.
we might want to make this a CSS var only

gkatsev:
Maybe it's better to use `<mux-video>` for the background video usecases? Since it doesn't have any UI with it, it's a bit more light weight.

gkatsev:
Which isn't to say, we don't want to allow this in some way.

richard-grosjean:
> Maybe it's better to use `<mux-video>` for the background video usecases? Since it doesn't have any UI with it, it's a bit more light weight.

This sounds like an excellent suggestion! We automatically jumped to the react package as we use Nextjs, but we might end up using both mux-video and mux-player-react in the future depending on the use case.

dylanjha:
@richard-grosjean unrelated to this specific change I'm also curious what errors you were running into, if those errors are showing up correctly in your Mux Data dashboard.

Obviously if there's anything we can do to actually prevent your specific errors from happening in the first place, we'd love to do that!

gkatsev:
> This sounds like an excellent suggestion! We automatically jumped to the react package as we use Nextjs, but we might end up using both mux-video and mux-player-react in the future depending on the use case.

Don't forget that we do have `@mux/mux-video-react` component for you to use.

richard-grosjean:
> > This sounds like an excellent suggestion! We automatically jumped to the react package as we use Nextjs, but we might end up using both mux-video and mux-player-react in the future depending on the use case.
> 
> Don't forget that we do have `@mux/mux-video-react` component for you to use.

We have switched to this now.

dylanjha:
Closing this, sounds like `<mux-video>` (instead of `<mux-player>`) is a better solution for what you're trying to do

PR #492, opened 2022-11-05
fix: add async effect for loading media

luwes:
- this adds an async effect for init playback core and loading the media for the mux-video and mux-audio custom elements.
- made the `mediaEl` required in playback-core making some code and types easier.

### DIFF
--- a/packages/mux-video/src/index.ts
+++ b/packages/mux-video/src/index.ts
@@ -392,11 +392,14 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
     }
   }
 
-  async load() {
-    if (this.#loadRequested || !this.src) return;
+  async #requestLoad() {
+    if (this.#loadRequested) return;
     await (this.#loadRequested = Promise.resolve());
     this.#loadRequested = null;
+    this.load();
+  }
 
+  load() {
     this.#core = initialize(this as Partial<MuxMediaProps>, this.nativeEl, this.#core);
   }
 
@@ -430,13 +433,13 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
         const hadSrc = !!oldValue;
         const hasSrc = !!newValue;
         if (!hadSrc && hasSrc) {
-          this.load();
+          this.#requestLoad();
         } else if (hadSrc && !hasSrc) {
           this.unload();
           /** @TODO Test this thoroughly (async?) and confirm unload() necessary (CJP) */
         } else if (hadSrc && hasSrc) {
           this.unload();
-          this.load();
+          this.#requestLoad();
         }
         break;
       }
--- a/packages/mux-video/src/index.ts
+++ b/packages/mux-video/src/index.ts
@@ -7,7 +7,6 @@ import {
   PlaybackTypes,
   ValueOf,
   toMuxVideoURL,
-  teardown,
   Metadata,
   MediaError,
   getError,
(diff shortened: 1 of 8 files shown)


packages/mux-video/src/index.ts
```diff
@@ -377,27 +378,37 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
     }
   }
 
-  load() {
-    this.__core = initialize(this as Partial<MuxMediaProps>, this.nativeEl, this.__core);
+  async load() {
+    // Reset Hls.js and Mux data synchronously so all needed props are updated but
+    // wait 1 tick w/ loading the source to prevent multiple requests and cancels.
+    console.time('init core');
+    this.#core = initialize(this as Partial<MuxMediaProps>, this.nativeEl, this.#core);
```
gkatsev, line 399: Is it an issue that we're potentially tearing down hls.js and mux-embed a bunch here?
If someone is fiddling with a bunch of props on the player, there might be a lot of churn. It's probably fine, since once it's settled and playback begins, we shouldn't be doing it anymore.

packages/mux-video/src/index.ts
```diff
@@ -415,22 +426,28 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
+        this.load();
         break;
       case 'src': {
-        const hadSrc = !!oldValue;
-        const hasSrc = !!newValue;
-        if (!hadSrc && hasSrc) {
-          this.load();
-        } else if (hadSrc && !hasSrc) {
-          this.unload();
-          /** @TODO Test this thoroughly (async?) and confirm unload() necessary (CJP) */
```
gkatsev, line 431: because of this new async source loading behavior, we don't need this piece and can rely on unloading if `!newValue` and loading on a `newValue`?

packages/mux-video/src/index.ts
```diff
@@ -440,31 +457,18 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
         break;
       case Attributes.PLAYBACK_ID:
         /** @TODO Improv+Discuss - how should playback-id update wrt src attr changes (and vice versa) (CJP) */
         this.src = toMuxVideoURL(newValue ?? undefined, { domain: this.customDomain }) as string;
         break;
-      case Attributes.DEBUG: {
-        const debug = this.debug;
-        if (!!this.mux) {
-          /** @TODO Link to docs for a more detailed discussion (CJP) */
-          console.info(
```
gkatsev, line 459: since isn't an issue anymore because setting this tears down hlsjs and mux-embed, right?

gkatsev (commented):
Seems like a reasonable approach.

packages/mux-video/src/index.ts
luwes: indeed a `debug` change will reset playback core

packages/mux-video/src/index.ts
luwes: I tested the performance hit and it's not that bad I think, it's under 1ms if there is no src.
I think it should be fine

packages/mux-video/src/index.ts
luwes: I might put this back, I over simplifed it

luwes: right now it is fine because unload only has the core teardown which is done again in initialize() but there could be a case when the unload method holds more local logic so safer to put it back

cjpillsbury (approved):
LGTM!

PR #498, opened 2022-11-16
feat: implement Media Chrome HTML based theme 

luwes:
this adds a working version of the Mux player theme as a Media Chrome theme.

MC templating concepts added:

- `{{ mustache }}` like vars
- inner `<template>` elements called directives that can be used for things like conditionals, loops, partials, etc
- `{{ myVar | filter }}` pipe filters on the vars that act as functions on the value stored in the variable
- `{{ myVar ?? 'fallback' }}` coalesce operator which can provide a default value
- `{{>MyPlayButton param="test" another="lala"}}` param passing to partials

todo: 
- [x] take in account escaping of characters in variables and how this should act in partials maybe need a new syntax for partials like `{{>MyPartialInHTML}}`
- [x] delete the old code of the theme
- [x] remove the `@github/template-parts` dependency. 
  it's probably best we maintain our own version and we require the InnerTemplatePart's so 
- [x] move `MediaThemeTemplate` to MediaChrome and formalize there more
- [x] upgrade Media Chrome to stable version


packages/mux-player/src/media-theme-mux.html
```diff
@@ -0,0 +1,641 @@
...
+    <slot name="media" slot="media"></slot>
+    <media-poster-image
+      slot="poster"
+      part="poster"
+      src="{{poster}}"
+      placeholder-src="{{placeholder}}"
+    ></media-poster-image>
+    <media-loading-indicator slot="centered-chrome" no-auto-hide></media-loading-indicator>
+
+    <template partial="TitleDisplay">
```
cjpillsbury, line 291: question(non-blocking): Any reason we couldn't declare the partials above or below all the other elements? If not, that might be a good convention for this use case ("reusable templated component renderings")

packages/mux-player/src/media-theme-mux.html
```diff
@@ -0,0 +1,641 @@
...
+          <text class="value" transform="translate(10 21)" style="font-size: 8px; font-family: 'ArialMT', 'Arial'">
+            {{forwardSeekOffset}}
+          </text>
+        </svg>
+      </media-seek-forward-button>
+    </template>
+
+    <template partial="MuteButton">
+      <media-mute-button part="bottom mute button" disabled="{{disabled}}" aria-disabled="{{disabled}}">
+        <svg aria-hidden="true" viewBox="0 0 24 24" slot="high">
```
cjpillsbury, line 348: suggestion(non-blocking): It would be unfortunate if we lost the ability for designers to create svgs that they could just use with the src. This would be a good thing to follow up on so we don't have to again inline everything. The same applies to the styles.css change imo.

packages/mux-player/test/web-test-runner.config.mjs
```diff
@@ -19,7 +19,7 @@ export default {
     esbuildPlugin({
       ts: true,
       json: true,
-      loaders: { '.css': 'text', '.svg': 'text' },
+      loaders: { '.css': 'text', '.svg': 'text', '.html': 'text' },
```
cjpillsbury, line 22: suggestion(non-blocking): Since this isn't "standard html," (aka it is intended to be used as a template) and per some of the other build-level callouts (css & svg inclusions), we should consider making this extension something like `.html.templ` (or whatever feels right that isn't simply `.html`) so we can distinctly target it for build tools going forward.

scripts/esbuilder/esbuilder.js
```diff
@@ -55,6 +55,7 @@ const options = {
   metafile: true,
   plugins: [i18nPlugin],
   loader: {
+    '.html': 'text',
```
cjpillsbury, line 58: same as above

yarn.lock
```diff
@@ -10638,10 +10633,10 @@ mdn-data@2.0.4:
   resolved "https://registry.yarnpkg.com/mdn-data/-/mdn-data-2.0.4.tgz#699b3c38ac6f1d728091a64650b65d388502fd5b"
   integrity sha512-iV3XNKw06j5Q7mi6h+9vbx23Tv7JkjEVgKHW4pimwyDGWm0OIQntJJ+u1C6mg6mK1EaTv42XQ7w76yuzH7M2cA==
 
-media-chrome@0.16.0:
-  version "0.16.0"
-  resolved "https://registry.yarnpkg.com/media-chrome/-/media-chrome-0.16.0.tgz#500a8c80bc2f4da2128ff203ccb612a1c53ee517"
-  integrity sha512-QCa4f2DHcr8UuTenXGtPSE8313uz5wRiZVvc5iAepmgupF/b4QDiIJyXByt8WBgJWps+Up/qWC4hvNMPsbva2w==
+media-chrome@0.16.1-canary.10-e20953a:
```
cjpillsbury, line 10636: issue(blocking): we should hold off on merging this until we get an official release out with your changes

packages/mux-player/src/template.ts
```diff
@@ -1,18 +1,27 @@
...
+import muxTheme from './media-theme-mux.html';
+import './dialog';
+import { getSrcFromPlaybackId } from './helpers';
+import { html } from './html';
 import { i18n, stylePropsToString } from './utils';
 
 import type { MuxTemplateProps } from './types';
 import { StreamTypes } from '@mux/playback-core';
 
+const muxTemplate = document.createElement('template');
```
cjpillsbury, line 16: question(blocking): Is this needed? From what I can tell, we're

1. pulling in an html template file as a string
2. using JS to create a template element to then set the innerHTML to that string
3. using a clone of that template instance's content as the thing to render into the template.

Couldn't we skip all of that and simply do something like:

```js
const staticTemplateStrPart = `
<style>
  ${cssStr}
</style>
${muxTheme}`;

export const template = (props: MuxTemplateProps) => html`
${staticTemplateStrPart}
${content(props)}
`;
```

? If not, can you briefly explain why not?

packages/mux-player/src/media-theme-mux.html
luwes: this felt most natural to me, above the layout templates where we use them

packages/mux-player/src/html.ts
```diff
@@ -1,5 +1,6 @@
-import { TemplateInstance, NodeTemplatePart, createProcessor, AttributeTemplatePart } from '@github/template-parts';
-import type { TemplatePart, TemplateTypeInit } from '@github/template-parts';
+import { document } from 'shared-polyfills';
+// @ts-ignore
+import { TemplateInstance, ChildNodePart, AttrPart, Part } from 'media-chrome/dist/media-theme-element.js';
```
cjpillsbury, line 3: 😎

packages/mux-player/src/html.ts
```diff
@@ -108,13 +109,13 @@ export function processPart(part: TemplatePart, value: unknown): void {
...
-const renderedTemplates = new WeakMap<Node | NodeTemplatePart, HTMLTemplateElement>();
-const renderedTemplateInstances = new WeakMap<Node | NodeTemplatePart, TemplateInstance>();
+const renderedTemplates = new WeakMap<Node | ChildNodePart, HTMLTemplateElement>();
+const renderedTemplateInstances = new WeakMap<Node | ChildNodePart, TemplateInstance>();
 export class TemplateResult {
   constructor(
     public readonly strings: TemplateStringsArray,
     public readonly values: unknown[],
-    public readonly processor: TemplateTypeInit
+    public readonly processor: any
```
cjpillsbury, line 118: question(non-blocking): Why did this have to be converted to `any`? Is it possible to avoid this?

packages/mux-player/test/web-test-runner.config.mjs
luwes: I don't think that's needed here, it's valid HTML. it being the html extension right away gets the right syntax in the editor. if we had many html files we import maybe but this is really a rare case

packages/mux-player/src/html.ts
```diff
@@ -149,67 +150,22 @@ export class TemplateResult {
...
-        staticValues.forEach((staticValue: TemplateResult) => {
-          join(staticValue.strings, staticValue.values);
-        });
-
-        staticStrings[staticStrings.length - 1] = staticStrings[staticStrings.length - 1] + strs[i + 1];
-        hasStatics = true;
-      } else {
-        dynamicValues.push(dynamicValue);
-        staticStrings.push(strs[i + 1]);
+const defaultProcessor = {
```
cjpillsbury, line 153: question(blocking): am I missing something or did we loose our `stringsCache` and any equivalent? Could this contribute to issues we've been discussing re:rendered template GC & duplication?

packages/mux-player/test/web-test-runner.config.mjs
cjpillsbury: right, this was related to being able to externalize e.g. the svg files and the css file

cjpillsbury (changes requested):
a few callouts, but this is looking 🔥

packages/mux-player/src/media-theme-mux.html
cjpillsbury: cool that makes sense to me, even if not my S1. Might be good to get @dylanjha and @gkatsev (and others') S1s here as well, even if I don't think we should block merging this regardless of folks' initial/gut thoughts.

packages/mux-player/src/html.ts
luwes: this logic was to add unsafe statics like a dynamic tag name but this isn't needed anymore because custom themes will be identified by an id attribute. not custom element tags anymore.

I don't think it will have any effect on the issue but lets see. it can only be positive.

packages/mux-player/src/template.ts
luwes: the html string is automatically escaped so it would just be text. we need to take advantage of the processing of this part https://github.com/muxinc/elements/blob/f24efa9501252d1d7385d359bc06475ecdd7f236/packages/mux-player/src/html.ts#L56-L62

packages/mux-player/src/html.ts
luwes: yes, there was always an annoying issue that I didn't want to spend time on. I'll check when moving this to MC

packages/mux-player/src/html.ts
cjpillsbury: I'm not sure if I agree with "it can only be positive," if one of our concerns is GC of the `WeakMap` references (I thought you said the `stringsCache` was the thing guaranteeing a maintained reference), but we can move forward with this and find out.

packages/mux-player/src/template.ts
cjpillsbury: ah we should try to resolve that (in a subsequent effort), since, in theory, the `template="templateId"` should do the processing, no?

cjpillsbury (approved):
Pre-approving with a few concerns we should monitor and pending the media chrome release + update of dependencies here.

Issue #500, opened 2022-11-17
Bug: not all metadata values are recorded


martinbean:
### Checklist

- [X] Verified whether your issue exists in the latest version Mux Element(s) release: `npm install @mux/[element_name]` or `yarn @mux/[element_name]`.
  - (NOTE: If the issue is not occurring in the latest version, we may not resolve it.)
- [X] Confirmed that this issue hasn't [already been filed](https://github.com/muxinc/elements/issues?q=is%3Aopen+is%3Aissue+label%3Abug).

### Which Mux Elements/Packages has a bug?

- common

### Which browser(s) are you using?

- Chrome
- Safari

### Which operating system(s) are you using?

macOS

### How are you using Mux Elements?

npm package

### Describe the bug

Not all metadata values are recorded with views. Particularly:

* `custom-1` and `custom-2`
* `sub-property-id`

## Steps To Reproduce

1. Add a `<mux-player>` element to your page.
2. Try setting the aforementioned properties, both with and without a `metadata-` prefix.

Link to code example:

https://codesandbox.io/s/musing-roentgen-885hzg?file=/index.html

## The current behavior

Values for custom dimensions and sub property ID are not recorded with views.

## The expected behavior

Values for custom dimensions and sub property ID to be recorded with views.

dylanjha:
Hi @martinbean thanks for opening this. I can see why this is confusing.

All metadata fields can be set by the `metadata=` property on the element instance

```
const muxPlayer = document.querySelector('mux-player')
muxPlayer.metadata = {
  video_id: 'xxx',
  video_title: 'xxx',
  viewer_user_id: 'xxx',
  sub_property_id: 'xxx'
}
```

For the web component, there are 3 special Mux Data metadata attributes that can be used for convenience.

- `metadata-video-id`
- `metadata-video-title`
- `metadata-viewer-user-id`

These are the only shorthand attributes, but I'll see if we can expand the attributes to work with all Mux Data metadata fields, I can see how you would expect anything with the `metadata-` prefix to work.

dylanjha:
@martinbean this is fixed on the [latest canary 1.4.0-canary.0-8ee139d](https://www.npmjs.com/package/@mux/mux-player/v/1.4.0-canary.0-8ee139d)

Please try it out! You should be able to provide any Mux Data metadata fields with the `metadata-` attribute prefix now.

dylanjha:
@martinbean this is resolved as of 1.4!

PR #501, opened 2022-11-17
feat: allow <mux-video> and <mux-player> to get any metadata-* attrs set

dylanjha:
Related to #500 

Keeping this as a draft rn to answer this question:

1. Is there any value to keeping the well-defined list of 3 "special" attributes? `Attributes.METADATA_VIDEO_ID`, etc? Autocomplete tooling, typescript flagging in editors? 
2. For `mux-player`, does it makes sense to use the `connectedCallback` in order to access the shadow DOMs `muxVideo.metadata=` prop?

# todo

- [x] resolve outstanding questions
- [x] tests

# mux-video

This solves it for `<mux-video>` so someone could do:

```html
<mux-video
  <!-- other attrs -->
  metadata-sub-property-id="2434"
></mux-video>
```

sub-id_2022-11-17_13-26-21
sub-id_2022-11-17_13-26-09

- [x] Also we'd need to do this for `<mux-audio>` (same code as this).

# mux-player

```html
<mux-player
  <!-- other attrs -->
  metadata-sub-property-id="2434"
></mux-player>
```

Makes a small internal change to how metadata is passed from `<mux-player>` -> `<mux-video>`. Now we *only* use the `muxVideo.metadata=` prop for assigning metadata. The attrs syntax is removed from the template


packages/mux-video/src/index.ts
```diff
@@ -377,11 +377,30 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
     const viewer_user_id = this.getAttribute(Attributes.METADATA_VIEWER_USER_ID);
+
+    const otherMetadataAttrs: { [key: string]: string } = {};
+    this.getAttributeNames()
+      .filter((attrName) => {
+        return (
+          attrName.startsWith('metadata-') &&
+          ![Attributes.METADATA_VIDEO_ID, Attributes.METADATA_VIDEO_TITLE, Attributes.METADATA_VIEWER_USER_ID].includes(
+            attrName
+          )
```
dylanjha, line 388: TS error:

```
src/index.ts:387:13 - error TS2345: Argument of type 'string' is not assignable to parameter of type '"metadata-video-id" | "metadata-video-title" | "metadata-viewer-user-id"'.

387             attrName
                ~~~~~~~~
```

packages/mux-player/src/index.ts
```diff
@@ -237,8 +237,25 @@ class MuxPlayerElement extends VideoApiElement {
...
+        }
+      });
+    return metadata;
+  }
+
   connectedCallback() {
     this.#renderChrome();
+    const muxVideo = this.shadowRoot?.querySelector('mux-video') as MuxVideoElement;
+    if (muxVideo) {
+      muxVideo.metadata = metadataFromAttrs();
```
cjpillsbury, line 257: issue(blocking): `muxVideo.metadata = this.metadataFromAttrs`

packages/mux-player/src/index.ts
```diff
@@ -1084,7 +1101,7 @@ class MuxPlayerElement extends VideoApiElement {
       logger.error('underlying media element missing when trying to set metadata. metadata will not be set.');
       return;
     }
-    this.media.metadata = val;
+    this.media.metadata = { ...metadataFromAttrs, ...val };
```
cjpillsbury, line 1104: issue(blocking): `this.media.metadata = { ...this.metadataFromAttrs, ...val }`

packages/mux-video/src/index.ts
```diff
@@ -362,11 +362,34 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
     const video_id = this.getAttribute(Attributes.METADATA_VIDEO_ID);
     const video_title = this.getAttribute(Attributes.METADATA_VIDEO_TITLE);
     const viewer_user_id = this.getAttribute(Attributes.METADATA_VIEWER_USER_ID);
+
+    const otherMetadataAttrs: { [key: string]: string } = {};
```
cjpillsbury, line 366: suggestion(non-blocking): Per your callout in your PR description, we shouldn't need to treat the "well-defined metadata attributes" differently here. I'd remove them so we can simplify this code.

cjpillsbury (changes requested):
a couple of quick fixes but otherwise 👌

dylanjha:
Update:

- going to try to add some tests
- examples have been updated so we can test the functionality on the deploy preview -- will revert those example updates before merging
- will want to squash & merge this one with a conventional commit message like:

> feat: allow arbitrary Mux Data `metadata-*` attributes syntax (previously only supported this syntax for `metadata-video-title`, `metadata-video-id` and `metadata-viewer-user-id`

packages/mux-audio/src/index.ts
```diff
@@ -254,7 +248,22 @@ class MuxAudioElement extends CustomAudioElement<HTMLAudioElement> implements Pa
   }
 
   get metadata() {
-    return this.#metadata;
+    const inferredMetadataAttrs: { [key: string]: string } = {};
+    this.getAttributeNames()
+      .filter((attrName) => {
+        return attrName.startsWith('metadata-') && !([Attributes.METADATA_URL] as string[]).includes(attrName);
+      })
+      .map((attrName) => {
```
cjpillsbury, line 256: nitpick(blocking): either use `forEach` if you're producing a side effect (like your implementation in `mux-player`) or refactor to use `reduce()`, e.g.:

```ts
const inferredMetadataAttrs: { [key: string]: string } = this.getAttributeNames()
      .filter((attrName) => {
        return attrName.startsWith('metadata-') && !([Attributes.METADATA_URL] as string[]).includes(attrName);
      })
      .reduce((curAttrs, attrName) => {
        const value = this.getAttribute(attrName);
        if (value != null) {
          curAttrs[attrName.replace(/^metadata-/, '').replace(/-/g, '_') as string] = value;
        }
        return currAttrs;
      }, {});
```

packages/mux-audio/test/player.test.js
```diff
@@ -91,4 +91,15 @@ describe('<mux-audio>', () => {
     await aTimeout(3000);
     assert.equal(player.buffered.length, 0, 'no buffer loaded');
   });
+
+  it('maps arbitrary metadata-* attrs to the metadata prop and populates video_id if not provided', async function () {
```
cjpillsbury, line 95: 😎

packages/mux-video/src/index.ts
```diff
@@ -361,14 +355,21 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
   get metadata() {
-    const video_id = this.getAttribute(Attributes.METADATA_VIDEO_ID);
-    const video_title = this.getAttribute(Attributes.METADATA_VIDEO_TITLE);
-    const viewer_user_id = this.getAttribute(Attributes.METADATA_VIEWER_USER_ID);
+    const inferredMetadataAttrs: { [key: string]: string } = {};
+    this.getAttributeNames()
+      .filter((attrName) => {
+        return attrName.startsWith('metadata-') && !([Attributes.METADATA_URL] as string[]).includes(attrName);
+      })
+      .map((attrName) => {
```
cjpillsbury, line 363: nitpick(blocking): same feedback as `mux-audio`

cjpillsbury (changes requested):
A couple of small asks. We also can end up in a weird state if folks set `metadata-*` attributes and the `metadata` object after initial load that wouldn't be the case with the current impl for our "well-defined" metadata attrs that we may want to solve.

packages/mux-audio/src/index.ts
dylanjha: good call!

cjpillsbury (approved):
🎉

dylanjha:
@cjpillsbury if I understand correctly there may be some edge-case breaking change if a user tries to set the same metadata fields in 2 different ways:

```html
<mux-player video-title="Title AB"></mux-player>
```

AND

```js
document.querySelector('mux-player').metadata = { video_title: 'XY' }
```

The potential change in behavior would be which one gets precedence?

If that's the case, I think we can live with that. Folks really shouldn't be setting the same metadata fields with different values in 2 different ways.

dylanjha:
Verified the behavior on esm-vanilla deploy preview, reverted the changes to `examples/`, will squash merge after the CI builds

cjpillsbury:
> @cjpillsbury if I understand correctly there may be some edge-case breaking change if a user tries to set the same metadata fields in 2 different ways:
> 
> ```
> <mux-player video-title="Title AB"></mux-player>
> ```
> 
> AND
> 
> ```js
> document.querySelector('mux-player').metadata = { video_title: 'Title XY' }
> ```
> 
> The potential change in behavior would be which one gets precedence?
> 
> If that's the case, I think we can live with that. Folks really shouldn't be setting the same metadata fields with different values in 2 different ways.

Kind of... before, neither ended up getting precedence because of the way the metadata attrs" flowed to" the underlying `mux-video` instance + the fact that the the getter for `metadata` always merged the set prop plus the attrs. Since we now never pass the `mux-player` attributes down via attributes, we loose this "automatic merging" behavior.

PR #515, opened 2022-12-08
Add Nuxt.js 3 example

luwes:
There's one issue with Vite not handling our Mux elements minified bundles correctly.
It throws an error 

```
Unexpected error when starting the router: SyntaxError: Identifier 'h' has already been declared
```

It's because our bundles already have a symbol called `h` and Vite tries to inject this import at the top of our bundles

```js
import { h } from '/_nuxt/node_modules/.vite/deps/vue.js?v=8f0c34a1';

var Be=(t,a,e)=>{if(!a.has(t))throw TypeError("Cannot "+e)};var l=(t,a,e)=>(Be(t,a,"read from private field"),e?e.call(t):a.get(t)),c=(t,a,e)=>{if(a.has(t))throw TypeError("Cannot add the same private member more than once");a instanceof WeakSet?a.add(t):a.set(t,e)},k=(t,a,e,i)=>(Be(t,a,"write to private field"),i?i.call(t,e):a.set(t,e),e);var p=(t,a,e)=>(Be(t,a,"access private method"),e);var z=class{addEventListener(){}removeEventListener(){}dispatchEvent(a){return!0}};if(typeof DocumentFragment=="undefined"){class t extends z{}globalThis.DocumentFragment=t}var ae=class extends z{},_e=class extends z{},ra={get(t){},define(t,a,e){},upgrade(t){},whenDefined(t){return Promise.resolve(ae)}},ie,Ne=class{constructor(a,e={}){c(this,ie,void 0);k(this,ie,e==null?void 0:e.detail)}get detail(){return l(this,ie)}initCustomEvent(){}};ie=new WeakMap;function oa(t,a){return new ae}var ft={document:{createElement:oa},DocumentFragment,customElements:ra,CustomEvent:Ne,EventTarget:z,HTMLElement:ae,HTMLVideoElement:_e},yt=typeof window=="undefined"||typeof globalThis.customElements=="undefined",h=yt?ft:globalThis,E=yt?ft.document:globalThis.document;
```

See the snippet `,h=yt?ft:globalThis,`.

It's a bug on their side.
I temporarily fixed it by not minifying our bundles for the Nuxt.js example

Related https://github.com/vitejs/vite/issues/10822


luwes:
closing this for now, I don't think it has that much benefits to add another build example.
we can reference this PR if anyone has issues with Nuxt.js

Issue #517, opened 2022-12-10
Bug: Mux player duplicate render / frozen TemplateObject array not unique per call site


luwes:
Related tagged template spec change
https://github.com/tc39/ecma262/issues/840

### Which Mux Elements/Packages has a bug?

- mux-player

### Which browser(s) are you using?

Chrome 108.0.5359.98
didn't affect Firefox and Safari in my tests

### Which operating system(s) are you using?

MacOS

### How are you using Mux Elements?

npm 

---

I finally reproduced this bug locally, one of the hardest bugs I've had the pleasure to diagnose 😅

It turns out it is unrelated to any JS framework, Nuxt nope, Mithril nope. It's reproducible with vanilla JS..
I think it's a Chrome only bug, at least I can only repro it there.

The bug is that the result of our `html` tagged template does not always return the same frozen TemplateObject array invalidating the WeakMap cache and causing a new DOM fragment to be appended in the shadow root of the player.

https://github.com/muxinc/elements/blob/308e9a7c879594edbe2d4f4ffe462c7efb53299f/packages/mux-player/src/template.ts#L12-L17

https://github.com/muxinc/elements/blob/77d0386606d5ecccb7c322ce487c9287d16374fd/packages/mux-player/src/html.ts#L134-L144

For some reason the frozen TemplateObject array is not unique per call site under a big amount of memory pressure in Chrome.

I could finally reproduce it by adding the app.js file of the ijf.org Judo website which is a huge JS file and adding logic to change the playback id on the fly with Mithril.js. I reproduced it twice by luck, after that nada couldn't reproduce it for another hour or something tried everything. Then thought maybe it's related to some memory increase thing and randomly added this rainbow spiral (https://luwes.github.io/sinuous-rainbow-spiral/) to the Mux player vanilla demo (https://elements-demo-vanilla-mux.vercel.app/mux-player.html) First setting the amount to 2000 for increased memory usage and voila could reproduce it like 50% of the time 🥹

luwes:
I confirmed that the TemplateObject array actually changes by using a regular Map and see if there is a "duplicate" key entry, same array values and this was the case.

![SCR-20221210-bv7](https://user-images.githubusercontent.com/360826/206862360-fedf87ae-148a-4a02-9422-77fedd9be793.png)

luwes:
Update: seems garbage collection is not the culprit, since the old array is still in memory. 
It just doesn't return the same frozen array per call site under some conditions.

[...]

PR #518, opened 2022-12-12
fix: duplicate render by using value cache key

luwes:
not a reference cache key related #517

### DIFF
--- a/packages/mux-player/src/html.ts
+++ b/packages/mux-player/src/html.ts
@@ -108,6 +108,8 @@ export function processPart(part: Part, value: unknown): void {
     processPropertyIdentity(part, value);
 }
 
+// The Map's data will not be garbage collected however the number of created
+// templates will not exceed the html`` defined templates. This is fine.
 const templates = new Map<string, HTMLTemplateElement>();
 const renderedTemplates = new WeakMap<Node | ChildNodePart, HTMLTemplateElement>();
 const renderedTemplateInstances = new WeakMap<Node | ChildNodePart, TemplateInstance>();
--- a/packages/mux-player/src/html.ts
+++ b/packages/mux-player/src/html.ts
@@ -121,7 +121,10 @@ export class TemplateResult {
     public readonly values: unknown[],
     public readonly processor: any
   ) {
-    // Use a control character to join the expression boundaries.
+    // Use a control character to join the expression boundaries. It should be
+    // a character that is not used in the static strings so the key is unique
+    // if the expressions are in a different place even tough the static strings
+    // are identical.
     this.stringsKey = this.strings.join('\x01');
   }
 
--- a/packages/mux-player/src/html.ts
+++ b/packages/mux-player/src/html.ts
@@ -108,24 +108,29 @@ export function processPart(part: Part, value: unknown): void {
     processPropertyIdentity(part, value);
 }
 
-const templates = new WeakMap<TemplateStringsArray, HTMLTemplateElement>();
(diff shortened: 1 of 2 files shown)


packages/mux-player/src/html.ts
```diff
@@ -108,24 +108,29 @@ export function processPart(part: Part, value: unknown): void {
...
+  public readonly stringsKey: string;
+
   constructor(
     public readonly strings: TemplateStringsArray,
     public readonly values: unknown[],
     public readonly processor: any
-  ) {}
+  ) {
+    // Use a control character to join the expression boundaries.
+    this.stringsKey = this.strings.join('\x01');
```
cjpillsbury, line 123: question(blocking): since `templates` is a module-scoped variable, what happens if two template strings in different parts of the codebase are identical? e.g. if this exists in two places:
```js
html`<style>${myStyles}</style>`
```

packages/mux-player/src/html.ts
```diff
@@ -108,24 +108,29 @@ export function processPart(part: Part, value: unknown): void {
     processPropertyIdentity(part, value);
 }
 
-const templates = new WeakMap<TemplateStringsArray, HTMLTemplateElement>();
+const templates = new Map<string, HTMLTemplateElement>();
```
cjpillsbury, line 111: question(blocking): Does this effectively mean no `HTMLTemplateElement` will ever be gc'ed? I'm not seeing any mechanism to do a `templates.delete(stringKey)`

cjpillsbury (changes requested):
a couple of questions before merging

packages/mux-player/src/html.ts
luwes: it will use the same HTMLTemplateElement but most of the times it will be cloned when this condition returns false
https://github.com/muxinc/elements/blob/031a89a0d8a46d715b9ad778a732b720c7694ab5/packages/mux-player/src/html.ts#L140-L140

if it's true it's likely a condition like 

```js
return bool ? html`<style>${myStyles}</style>` : html`<style>${myPaint}</style>`
```

where only the expression value will update.

packages/mux-player/src/html.ts
```diff
@@ -108,24 +108,29 @@ export function processPart(part: Part, value: unknown): void {
...
-  ) {}
+  ) {
+    // Use a control character to join the expression boundaries.
+    this.stringsKey = this.strings.join('\x01');
+  }
 
   get template(): HTMLTemplateElement {
-    if (templates.has(this.strings)) {
-      return templates.get(this.strings) as HTMLTemplateElement;
+    if (templates.has(this.stringsKey)) {
```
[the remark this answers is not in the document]
luwes: yes, the frozen array in Chrome is not returning the same one for the same call site. I suspect it's a Chrome bug but not 100% sure.

packages/mux-player/src/html.ts
luwes: yes, that's right. I don't really see an other way to do it at the moment. 
it should not add more and more templates though, it only goes until each tagged template is evaluated.

quickly tested, it's about 5 entries changing all permutations and such

cjpillsbury: Cool. I think this would be resolved by my proposal of the `<mux-player/>` instance "owning" the reference to the template instead of the cache `Map` object, but that would be a bigger refactor of the underlying architecture. I think something in that direction makes more sense longer term, but let's treat that as a separate ticket.

PR #520, opened 2022-12-13
chore: add duplicate render repro to vanilla demo (DO NOT MERGE)

luwes:


luwes:
both @cjpillsbury and I could reproduce it on commit https://github.com/muxinc/elements/pull/520/commits/364e3ea83b84a65092f18c44a1add101b48fae20

but not on the latest Mux player version before the tagged template cache key fix 
https://github.com/muxinc/elements/pull/520/commits/d6817d46f342b322ce42b5e2faf7cc36fe94836b

conclusion is that potentially it's a bug in the template parts implementation https://github.com/github/template-parts or that code being used in there has an influence on the frozen array of the tagged template (like the generators) 
the latest Mux player uses the template parts implementation of https://github.com/dy/template-parts
and it's also using the new Media Chrome theme system which could make a difference too.

closing for now because the cache key by array string value will fix this bug for sure

PR #527, opened 2022-12-19
fix: memory leak of detached media elements in CastableVideo

luwes:
related to #308

this first commit fixes a memory leak caused by the Set in CastableVideo but there appears another issue for MuxVideoReact. still digging in that one


gkatsev (approved):
Seems reasonable to merge as is and other changes can come in a separate PR. Up to you, though.

PR #536, opened 2023-01-03
test(mux-uploader): add basic init test

davekiss:
This introduces basic tests for mux-uploader initialization.


packages/mux-uploader/package.json
```diff
@@ -46,7 +46,8 @@
     "publish-release": "../../scripts/publish.sh"
   },
   "dependencies": {
-    "@mux/upchunk": "^3.0.0"
+    "@mux/upchunk": "^3.0.0",
+    "sinon": "^15.0.1"
```
luwes, line 50: Sinon must be a dev dependency otherwise it will be installed by the consumers of @mux/mux-uploader which is not needed to run this package.

packages/mux-uploader/test/web-test-runner.config.mjs
```diff
@@ -2,6 +2,22 @@ import { esbuildPlugin } from '@web/dev-server-esbuild';
 import { importMapsPlugin } from '@web/dev-server-import-maps';
 
 export default {
+  testFramework: {
+    config: {
+      ui: 'bdd',
+      timeout: '5000',
+    },
+  },
+  middleware: [
```
cjpillsbury, line 11: thought(non-blocking): Fine for now, though this may be hard to follow/maintain longer term.

packages/mux-uploader/test/web-test-runner.config.mjs
```diff
@@ -2,6 +2,22 @@ import { esbuildPlugin } from '@web/dev-server-esbuild';
...
+    function respondWithUploadSuccess(context, next) {
+      if (context.url === '/mock-upload-endpoint') {
```
cjpillsbury, line 13: suggestion(non-blocking): Add a code comment here referencing/discussing where/how it's used in the tests.

packages/mux-uploader/test/uploader.test.js
```diff
@@ -1,5 +1,69 @@
...
+
+    const { detail } = await oneEvent(uploader, 'uploaderror');
+
+    assert.equal(detail.message, 'No url or endpoint specified -- cannot handleUpload', 'error message matches');
+    assert.exists(uploader.getAttribute('upload-error'), 'upload error is true');
+  });
+
+  it('completes a mock upload', async function () {
+    const uploader = await fixture(`<mux-uploader
+      endpoint="/mock-upload-endpoint"
```
cjpillsbury, line 47: suggestion(non-blocking): Add a code comment here referencing the dev server middleware config dependency.

cjpillsbury (approved):
A couple of non-blocking suggestions. Otherwise, 🎉

cjpillsbury:
@davekiss just a heads up, make sure you squash+merge this and update the commit comment to conform to https://www.conventionalcommits.org/en/v1.0.0/#summary (we rely on this to let the robots automatically figure out versioning, release notes, etc.)

luwes (approved):
👍

packages/mux-uploader/test/web-test-runner.config.mjs
davekiss: I explored using a Sinon fake server and it just didn't seem possible since Upchunk's XHR was modifying and caching a version of `XMLHttpRequest` before Sinon could shim it

gkatsev: it's possible to do so by overriding the global XHR before anything else is loaded.

cjpillsbury: All good/definitely not a blocker. We may be able to apply similar acrobatics tricks that I ended up having to apply for `xhr-mock` in Upchunk (See, e.g.: https://github.com/muxinc/upchunk/blob/master/test/upchunk.spec.ts#L8-L19). I should be able to help investigate that to see if it's possible (either with sinon or simply again using `xhr-mock` for testing here, or if we want to just continue relying on the dev server middleware). Even if we stick with the middleware solution, we may want to break that stuff out into its own module? Either way, I don't think we should worry about it too much just yet.

gkatsev: yeah, I don't think it's a blocker for this PR, but I do think longer term it would be good to set it up. Should be similar to who you have there with xhr-mock.

gkatsev: I do think the sinon direction would be best because it makes the test self-contained as the "test server" is defined inside the test.

PR #538, opened 2023-01-04
Update node version

gkatsev:


packages/playback-core/src/types.ts
```diff
@@ -4,7 +4,7 @@ import Hls, { HlsConfig } from 'hls.js';
 type KeyTypes = string | number | symbol;
 
 // Type Guard to determine if a given key is actually a key of some object of type T
-export const isKeyOf = <T = any>(k: KeyTypes, o: T): k is keyof T => {
+export const isKeyOf = <T extends object = any>(k: KeyTypes, o: T): k is keyof T => {
```
cjpillsbury, line 7: nitpick(blocking): Minimally, let's make this `extends {}` for consistency/generic usage. I'd suggest/recommend just making it identical with the other package impls (aka adding the `Maybe` and the `isNil` check.

cjpillsbury (changes requested):
Minor nitpick outside of the examples fixes.

packages/playback-core/src/types.ts
gkatsev: updated via [fb3340f](https://github.com/muxinc/elements/pull/538/commits/fb3340f6b2d0844b1f563560f61f96e19107f71c)

luwes (approved):
👍

cjpillsbury (approved):
nice!

PR #539, opened 2023-01-06
chore(mux-uploader): improve element modularity

davekiss:
Setting the foundation for future work to be done on the mux-uploader element.

Note, this does introduce a minor breaking change by removing the `type` attribute from the root element. I'm in favor of that; `type` is too vague to contain the value of the progress indicator within the root element.


packages/mux-uploader/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const PROGRESS_INDICATOR_TYPE_BAR = 'bar';
```
cjpillsbury, line 1: suggestion(non-blocking): for "enum-like" and other related-constant lookups, we'll typically do something like:
```ts
export type ProgressIndicatorTypes = {
  BAR: 'bar';
  RADIAL: 'radial';
};

export const ProgressIndicatorTypes: ProgressIndicatorTypes = {
  BAR: 'bar',
  RADIAL: 'radial',
};

```

cjpillsbury (approved):
🎉 lovin' where this is going.

luwes (approved):
LGTM!
the naming `mux-uploader-progress-indicator` is a bit long but I assume it's only used internally, and there is still a tight coupling with the parent so should be fine

luwes (approved):
👍

packages/mux-uploader/src/progress.ts
```diff
@@ -0,0 +1,167 @@
+import { globalThis, document } from 'shared-polyfills';
+
+const template = document.createElement('template');
+const TYPES = {
```
cjpillsbury, line 4: nitpick(blocking): Use the `ProgressTypes` lookup object defined in `constants` and update references in this module accordingly.

packages/mux-uploader/src/progress.ts
```diff
@@ -0,0 +1,167 @@
+import { globalThis, document } from 'shared-polyfills';
```
cjpillsbury, line 1: nitpick(blocking): We should probably keep the module names consistent (aka name it `mux-uploader-progress`.

packages/mux-uploader/src/progress.ts
```diff
@@ -0,0 +1,167 @@
...
+    <circle
+      r="52"
+      cx="60"
+      cy="60"
+    />
+  <svg>
+</div>
+`;
+
+class Progress extends globalThis.HTMLElement {
```
cjpillsbury, line 82: nitpick(blocking): rename to `MuxUploaderProgressElement` for parity with our tacit convention.

packages/mux-uploader/src/progress.ts
```diff
@@ -0,0 +1,167 @@
...
+    this.svgCircle = this.shadowRoot?.querySelector('circle');
+    this.progressBar = this.shadowRoot?.getElementById('progress-bar');
+    this.progressBar?.setAttribute('aria-description', ariaDescription);
+  }
+
+  connectedCallback() {
+    this.setDefaultType();
+    const parent = document.querySelector('mux-uploader');
+
+    parent?.addEventListener('progress', (e) => {
```
cjpillsbury, line 102: suggestion(non-blocking): We as a followup, we should probably make the uploader "push" progress updates to this component, rather than the current impl that "reaches out" for progress updates. This avoids unnecessary assumptions in the DOM structure, avoids potential memory leaks/"dangling references", and allows the component to be more "unit testable".

packages/mux-uploader/src/progress.ts
```diff
@@ -0,0 +1,167 @@
...
+      }
+    }
+  }
+}
+
+if (!globalThis.customElements.get('mux-uploader-progress')) {
+  globalThis.customElements.define('mux-uploader-progress', Progress);
+}
+
+export default Progress;
```
cjpillsbury, line 167: nitpick(blocking): rename to `MuxUploaderProgressElement` for parity with our tacit convention.

cjpillsbury (changes requested):
a few nits, otherwise 🚢 🇮🇹

cjpillsbury (approved):
🏄‍♂️

PR #542, opened 2023-01-12
chore(mux-uploader): move file-select to own component

davekiss:
- breaks out file-selector to its own component
- add tests
- breaking: renames the slot from `upload-button` to `file-select`


packages/mux-uploader/src/mux-uploader-file-select.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  handleFilePickerButtonClick() {
+    // TO-DO: Allow user to reattempt uploading the same file after an error.
+    // Note: Apparently Chrome and Firefox do not allow changing an indexed property on FileList...(TD).
+    // Source: https://stackoverflow.com/a/46689013
+    const attr = this.getAttribute('mux-uploader');
+    const controller = attr ? document.getElementById(attr) : (<ShadowRoot>this.getRootNode()).host;
+
+    controller?.shadowRoot?.querySelector<HTMLInputElement>('#hidden-file-input')?.click();
```
cjpillsbury, line 92: question(blocking): Shouldn't the `#hidden-file-input` be owned by this component as well? Seems like an appropriate separation of concerns/single responsibility here. You'd then just have it dispatch the same event as `MuxUploaderDropElement` and those "look identical" from the perspective of the uploader element.

packages/mux-uploader/src/mux-uploader-file-select.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+</style>
+
+<slot>
+<button type="button">Upload video</button>
+</slot>
+`;
+
+class MuxUploaderFileSelectElement extends globalThis.HTMLElement {
+  protected _filePickerButton: HTMLElement | null | undefined;
```
cjpillsbury, line 47: nitpick(blocking): We've been moving to using native javascript `#private` variables. Change for consistency.

packages/mux-uploader/src/mux-uploader-file-select.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      this.filePickerButton = slot.assignedElements({ flatten: true })[0] as HTMLButtonElement;
+    });
+  }
+
+  protected get filePickerButton() {
+    return this._filePickerButton;
+  }
+
+  protected set filePickerButton(value: HTMLElement | null | undefined) {
```
cjpillsbury, line 73: praise: smart to use the setter here.

packages/mux-uploader/src/mux-uploader-file-select.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (value === this._filePickerButton) return;
+    if (this._filePickerButton) {
+      this._filePickerButton.removeEventListener('click', this.handleFilePickerButtonClick);
+    }
+
+    this._filePickerButton = value;
+    if (this._filePickerButton) {
+      this._filePickerButton.addEventListener('click', this.handleFilePickerButtonClick);
```
cjpillsbury, line 81: thought(non-blocking): Could we make this `useCapture` to easily handle even more cases? I could see some negatives there on assumptions with that approach, so definitely more "food for thought" right now.

cjpillsbury (changes requested):
A couple of callouts, but overall I'm loving the direction this is going.

packages/mux-uploader/src/mux-uploader-file-select.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+
+  handleFilePickerButtonClick() {
+    // TO-DO: Allow user to reattempt uploading the same file after an error.
+    // Note: Apparently Chrome and Firefox do not allow changing an indexed property on FileList...(TD).
+    // Source: https://stackoverflow.com/a/46689013
+    const attr = this.getAttribute('mux-uploader');
+    const controller = attr ? document.getElementById(attr) : (<ShadowRoot>this.getRootNode()).host;
```
cjpillsbury, line 90: suggestion(non-blocking): I believe you can use `(this.getRootNode() as ShadowRoot).host` here.

packages/mux-uploader/src/mux-uploader-file-select.ts
davekiss: yes, I think so too, but i need to get the event bus in place first for that to work which is in another branch at the moment. can we address this in another PR?

cjpillsbury: Deal.

cjpillsbury (approved):
🚀

PR #545, opened 2023-01-13
Feat/cuepoints

cjpillsbury:
Basic implementation of cuepoints support.

# TODO:
- [x] Add custom event for `cuepointchange` to `mux-video`
- [x] Expose interface on `mux-player` (`addCuePoints()`, `cuePoints`, `activeCuePoint`)
- [x] Add docs
- [x] Add tests for components
- [x] Add at least one example to at least one examples project

# Example Usage
```js
// NOTE: value can be whatever you want (including nothing) as long as it is serializable to JSON
const cuePoints = [
      { time: 0, value: { label: 'CTA 1', showDuration: 10 } },
      { time: 15, value: { label: 'CTA 2', showDuration: 5 } },
      { time: 21, value: { label: 'CTA 3', showDuration: 2 } },
    ];
// Or muxVideoEl. NOTE: media src/playback-id should have already been loaded.
muxPlayerEl.addCuePoints(cuePoints);
console.log('cuePoints', ...muxPlayerEl.cuePoints); // should log cuePoints
muxPlayerEl.addEventListener('cuepointchange', (event) => {
  console.log('activeCuePoint (event)', event.detail); // should log current activeCuePoint;
  console.log('activeCuePoint (element)', muxPlayerEl.activeCuePoint); // should log current activeCuePoint;
});
muxPlayerEl.play();
```
For another example, see the deploy of the new simple CuePoints example added to our vanilla example app: https://elements-demo-vanilla-git-fork-cjpillsbury-feat-cuepoints-mux.vercel.app/mux-player-cuepoints.html


packages/mux-player/src/index.ts
```diff
@@ -1145,6 +1145,19 @@ class MuxPlayerElement extends VideoApiElement {
     this.media.metadata = { ...this.metadataFromAttrs, ...val };
   }
 
+  async addCuePoints<T = any>(cuePoints: { timestamp: number; value: T }[]) {
+    if (!this.media) return Promise.reject(new Error('Cannot add cuePoints until mux-player has initialized'));
```
gkatsev, line 1149: comment(non-blocking): would be nice if we had a ready event or something so that this could be handled automatically for users 🙂

packages/mux-video/src/index.ts
```diff
@@ -380,6 +402,16 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
     }
   }
 
+  // NOTE: This needs to be public so it is accessible to playback-core
```
gkatsev, line 405: Is there a need for this method? Wouldn't playback-core triggering a cuepointchange on the media-el be sufficient?

packages/playback-core/src/tracks.ts
```diff
@@ -137,3 +139,106 @@ export function removeTextTrack(mediaEl: HTMLMediaElement, track: TextTrack) {
   );
   trackElement?.remove();
 }
+
+const DEFAULT_CUEPOINTS_TRACK_LABEL = 'cuepoints';
```
gkatsev, line 143: thought: should we namespace this like mux-cuepoints? Will help avoid (unlikely) naming collisions.

packages/mux-player/src/index.ts
```diff
@@ -1145,6 +1145,19 @@ class MuxPlayerElement extends VideoApiElement {
     this.media.metadata = { ...this.metadataFromAttrs, ...val };
   }
 
+  async addCuePoints<T = any>(cuePoints: { timestamp: number; value: T }[]) {
```
gkatsev, line 1148: thought: should passing in `null` remove all cuepoints? Alternatively, have a separate `removeCuePoints` method that you can pass in the cuepoints to remove. So, like `removeCuePoints(cuePoints())` would remove all.

gkatsev (commented):
Seems reasonable. Would we want to document on how to get the specific cuepoints track if anyone wants to listen to the `enter` or `exit` events of a particular cue? Alternatively, `cuePoints` could include a reference to the actual VTTCue object?

dylanjha:
Developer-facing API looks good 👍🏼

examples/vanilla-ts-esm/public/mux-player-cuepoints.html
```diff
@@ -0,0 +1,86 @@
...
+    ></mux-player>
+    <div>
+      <b>Active CuePoint:</b> <pre id="cuepoint-renderer">Unset</pre>
+    </div>
+    <script>
+      const playerEl = document.getElementById('cued-player');
+
+      function addCuePointsToPlayer() {
+        const cuePoints = [
+          { timestamp: 1, value: 'Simple Value' },
```
luwes, line 60: nit (non-blocking): why not the shorter wording `time`, saves some characters

packages/mux-video/src/index.ts
```diff
@@ -493,4 +507,10 @@ if (!globalThis.customElements.get('mux-video')) {
 
 export { PlaybackEngine, PlaybackEngine as Hls, ExtensionMimeTypeMap as MimeTypes, MediaError, VideoEvents };
 
+/** @TODO REMOVE ME: Just testin' (CJP) */
+// const muxVideoEl = new MuxVideoElement();
+// muxVideoEl.addCuePoints<number|boolean>([{ timestamp: 0, value: 20 }, { timestamp: 0, value: true }]);
+// const cuePoints: ({ timestamp: number, value: number | boolean })[] = [{ timestamp: 0, value: 20 }, { timestamp: 0, value: true }];
+// muxVideoEl.addCuePoints(cuePoints);
```
luwes, line 514: don't forget to remove

luwes (changes requested):
looks great overall. one blocking ask, I think since that method is async already might as well prevent that error from showing.

packages/mux-player/src/index.ts
cjpillsbury: I'm not sure this should be blocking. This is really an edge case that we're accounting for. A few callouts here:

> a simple promise that resolves once this.media is available

1. I'm not sure this is "a simple promise", unless I'm missing something. What are we monitoring for when the promise should resolve? `#init()` doesn't have a direct event to listen for. We could add a poll, or we listen for an indirect event, or a few other paths, but I'm not seeing a 1:1 for this case. Moreover, in the current basic implementation, `addCuePoints()` is tied to a `src`/`playback-id` (See docs, examples, and tests), which brings up:
2. Even if we refactored this to solve this case, I'm not sure the proposal would be sufficient and may confuse the user, since adding CuePoints before a src has been loaded will get "blown away" once the src has loaded (or more specifically, once `playback-core` has been `initialize()`ed. While there is a path forward for solving these cases as well, that's feeling like scope creep for this first pass addition of CuePoint support based on prior conversations.

examples/vanilla-ts-esm/public/mux-player-cuepoints.html
cjpillsbury: @muxinc/open-elements Thoughts here? Eventually, we'll plausibly (somehow) support wall clock times (for live) and potentially ranges. If folks feel strongly one way or another, now would be the time to change this key name.

packages/playback-core/src/tracks.ts
luwes: I would leave this as is, `cuepoints`. seems very unlikely to collide

packages/mux-player/src/index.ts
luwes: true, a promise does add more complex logic. 
(something like this https://github.com/luwes/super-media-element#usage)

calling `this.#init()` in `addCuePoints()` would be even simpler to make sure media is defined.

2. true

cjpillsbury: awesome I like the `this.#init()` idea. I'll be sure to add it.

examples/vanilla-ts-esm/public/mux-player-cuepoints.html
cjpillsbury: refactored to `time` per out of band convo. 🎉

luwes (approved):

PR #546, opened 2023-01-16
docs(mux-player): Document media-object-size and media-object-position css variables

AdamJaggard:
Mux Player supports `--media-object-size` and `--media-object-position` css variables but doesn't have them documented. This just adds them to the `REFERENCE.md` file.

See: #455 for the original feature PR.


cjpillsbury (approved):
👌

PR #547, opened 2023-01-16
feat(mux-player): Seek-to-live button to use Media Chrome

AdamJaggard:
Replaces the custom seek-to-live button with the new one from Media Chrome.

Currently using the latest canary because there's a playback issue linking the head of main with the currently deployed media-chrome. @cjpillsbury can you let me know when that fix is deployed and I will update the dependency.

image

Untitled

### DIFF
--- a/packages/mux-player/package.json
+++ b/packages/mux-player/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mux/mux-player",
-  "version": "1.5.1",
+  "version": "1.6.0",
   "description": "An open source Mux player web component that Just Works™",
   "main": "./dist/index.cjs.js",
   "module": "./dist/index.mjs",
@@ -50,9 +50,9 @@
     "publish-release": "../../scripts/publish.sh"
   },
   "dependencies": {
-    "@mux/mux-video": "0.12.0",
-    "@mux/playback-core": "0.15.0",
-    "media-chrome": "0.17.1"
+    "@mux/mux-video": "0.12.1",
+    "@mux/playback-core": "0.15.1",
+    "media-chrome": "0.18.0"
   },
   "devDependencies": {
     "@mux/esbuilder": "0.1.0",
--- a/packages/mux-player/package.json
+++ b/packages/mux-player/package.json
@@ -52,7 +52,7 @@
   "dependencies": {
     "@mux/mux-video": "0.12.0",
     "@mux/playback-core": "0.15.0",
-    "media-chrome": "0.17.1-canary.10-464ac4a"
+    "media-chrome": "0.17.1"
   },
   "devDependencies": {
     "@mux/esbuilder": "0.1.0",
--- a/packages/mux-player/package.json
+++ b/packages/mux-player/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mux/mux-player",
-  "version": "1.5.1",
+  "version": "1.6.0",
   "description": "An open source Mux player web component that Just Works™",
   "main": "./dist/index.cjs.js",
   "module": "./dist/index.mjs",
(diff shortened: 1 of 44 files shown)


packages/mux-player/src/template.ts
```diff
@@ -108,23 +108,20 @@ export const content = (props: MuxTemplateProps) => html`
...
     ${isLiveOrDVR(props)
-      ? html`<button
+      ? html`<media-live-button
           slot="seek-live"
           part="${isLive(props) ? 'top' : 'bottom'} seek-live button"
           disabled="${props.inLiveWindow || !props.hasSrc || props.isDialogOpen}"
           aria-disabled="${(props.inLiveWindow || !props.hasSrc || props.isDialogOpen) && 'true'}"
-          in-live-window="${props.inLiveWindow}"
+          media-time-is-live="${props.inLiveWindow}"
           onclick="${function (this: HTMLButtonElement, evt: Event) {
```
gkatsev, line 117: I don't think we need this onclick anymore, as it does it natively now: https://github.com/muxinc/media-chrome/blob/8c1e730b4baaa316f285d2be719451a1dea55c00/src/js/media-live-button.js#L59-L72

packages/mux-player/src/template.ts
```diff
@@ -108,23 +108,20 @@ export const content = (props: MuxTemplateProps) => html`
         : html``}
     </mux-video>
     ${isLiveOrDVR(props)
-      ? html`<button
```
gkatsev, line 111: I think this is fine for a first pass, but I think ultimately, we'd want it to be in template itself replacing the slot https://github.com/muxinc/elements/blob/main/packages/mux-player/src/media-theme-mux.html

AdamJaggard: Just had a chat about this with @cjpillsbury, I'm going to move it to the template before merging 🙂

packages/mux-player/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@mux/mux-video": "0.12.0",
     "@mux/playback-core": "0.15.0",
-    "media-chrome": "0.16.2"
+    "media-chrome": "0.17.1-canary.10-464ac4a"
```
gkatsev, line 55: Marking this as a blocker, so we don't forget.

gkatsev (changes requested):

luwes:
SCR-20230120-gl7

vertical alignment is slightly off. 
the LIVE text is 14px font-size, the title 16px.

I tried to quickly fix it via web inspector but was not as easy as it seems. 
if you change the LIVE text to 16px for a moment you can see the correct alignment but then that label is too big.

I think the solution is to move the LIVE text 1px down but preferably not with position relative top 1px or something hacky like that. maybe line-height or one of the flex properties, vertical-align, ....

packages/mux-player/src/media-theme-mux.html
```diff
@@ -501,7 +501,7 @@
       </template>
       <media-control-bar>
         {{>PlayButton}}
-        <slot name="seek-live"></slot>
+        <media-live-button disabled="{{disabled}}" aria-disabled="{{disabled}}"></media-live-button>
```
luwes, line 504: (blocking): use a template partial for this button with a `part` attribute

luwes (changes requested):
overall it looks great.
I'd prefer if we get the vertical 1px and required `part` attribute fixed before merging

packages/mux-player/src/media-theme-mux.html
```diff
@@ -451,6 +452,15 @@
       </media-cast-button>
     </template>
 
+    <template partial="LiveButton">
+      <!-- NOTE: Keeping "seek-live" part for backwards compatibility for now -->
+      <media-live-button
+        part="{{section ?? 'top'}} live seek-live button"
```
luwes, line 458: 👍

packages/mux-player/src/media-theme-mux.html
```diff
@@ -261,7 +261,8 @@
       line-height: 24px;
     }
 
-    .title-display {
+    :is(.title-display, media-live-button) {
+      color: inherit;
       font-size: 16px;
```
luwes, line 266: overall great but

(blocking): `media-live-button`'s  should be 14px and vertically aligned against the red dot and the title

luwes (changes requested):

PR #548, opened 2023-01-17
chore(mux-uploader): move screenreader text to own component

davekiss:
- further improving modularity
- introducing an event bus for communicating between components


packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -95,8 +86,7 @@ template.innerHTML = `
   ${styles}
 </style>
 
-<div class="sr-only" id="sr-only" aria-live="polite"></div>
-
+<mux-uploader-sr-text></mux-uploader-sr-text>
 <div class=text-container>
```
cjpillsbury, line 90: idea(non-blocking): It's feeling like we should have some kind of state attribute+property for at least the upload state. Off the dome finite state possibilities: PENDING, PROGRESS (or UPLOADING etc.), SUCCEEDED (or UPLOADED etc.), FAILED (or ERROR etc.).

packages/mux-uploader/src/utils/element-utils.ts
```diff
@@ -0,0 +1,14 @@
+export const closestComposedNode = (childNode: Element, selector: string): HTMLElement | null => {
+  if (!childNode) return null;
+  const closest = childNode.closest<HTMLElement>(selector);
+  if (closest) return closest;
+  return closestComposedNode((childNode.getRootNode() as ShadowRoot).host, selector);
+};
+
+export const getMuxUploaderEl = (controlEl: Element): HTMLElement | null => {
```
cjpillsbury, line 8: praise: Good idea to abstract this. 😎

packages/mux-uploader/src/mux-uploader-sr-text.ts
```diff
@@ -0,0 +1,59 @@
...
+  width:1px;
+  height:1px;
+  overflow:hidden;
+}
+</style>
+
+<div class="sr-only" id="sr-only" aria-live="polite"></div>
+`;
+
+class MuxUploaderSrTextElement extends globalThis.HTMLElement {
```
cjpillsbury, line 22: question(non-blocking): Do we need the SR-only element to be its own component? For all of the visual components, I can think of plausible use cases, but this feels more tightly coupled to the uploader element itself?

packages/mux-uploader/src/mux-uploader-sr-text.ts
```diff
@@ -0,0 +1,59 @@
...
+  constructor() {
+    super();
+    const shadowRoot = this.attachShadow({ mode: 'open' });
+    shadowRoot.appendChild(template.content.cloneNode(true));
+
+    this.srOnlyText = this.shadowRoot?.getElementById('sr-only');
+  }
+
+  connectedCallback() {
+    this.#uploaderEl = this.shadowRoot?.host ? getMuxUploaderEl(this.shadowRoot.host) : null;
```
cjpillsbury, line 35: question(blocking): Can we just use `getMuxUploaderEl(this)`? if yes, let's simplify.

packages/mux-uploader/src/mux-uploader.ts
gkatsev: this kind of ties into the nested shadow doms issue where if we wanted a .sr-only utility class we don't have an easy way of sharing it. Both since we can't include it at the top level of an element and expect it to be available to nested shadow doms, but also, we don't currently have a way to do CSS includes (though, I believe this is solvable via esbuild). Though, that could mean we're including the same utility multiple times.

packages/mux-uploader/src/mux-uploader-sr-text.ts
gkatsev: this kind of ties into the nested shadow doms issue where if we wanted a .sr-only utility class we don't have an easy way of sharing it. Both since we can't include it at the top level of an element and expect it to be available to nested shadow doms, but also, we don't currently have a way to do CSS includes (though, I believe this is solvable via esbuild). Though, that could mean we're including the same utility multiple times.

packages/mux-uploader/src/mux-uploader.ts
davekiss: Totally, this is scoped and will be in a future revision

packages/mux-uploader/src/mux-uploader-sr-text.ts
davekiss: good callout @cjpillsbury. I could be convinced either way, but decided to move it to its own bc I plan to handle more events and didn't want to clutter up the `mux-uploader` file with that logic

If either of you feel strongly that it's better off tied to the uploader el, lmk and I can move it back in and treat this PR as a spike on eventing.

cjpillsbury: definitely not a blocker for this PR. We can revisit if/when appropriate between now and v1.0

cjpillsbury (approved):
🎉

PR #550, opened 2023-01-23
bump to latest hls.js (1.3.1)

### LINKED ISSUE #541: Example with svelte-kit not working
First thank you for providing a Player with MUX. 

I'm trying to implement the player within sveltekit so naturally I checked the docs and exemples you provided. Nevertheless, after few attempts I decided to directly run locally your svelte-kit exemple. 

But just as for my attempts, I'm facing the same issue while using the player:
Screenshot 2023-01-12 at 17 31 45

Is it because I'm running it on localhost?

dylanjha:
Might resolve #541 build issue in Svelte

### DIFF
--- a/packages/playback-core/package.json
+++ b/packages/playback-core/package.json
@@ -49,7 +49,7 @@
     "publish-release": "../../scripts/publish.sh"
   },
   "dependencies": {
-    "hls.js": "1.2.3",
+    "hls.js": "1.3.1",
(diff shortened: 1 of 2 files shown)


dylanjha:
Confirmed locally this fixes the sveltekit-demo locally (had to do a bunch of acrobatics to test it, but @cjpillsbury and I confirmed)

PR #551, opened 2023-01-23
fix: use new MC template syntax

luwes:
requires https://github.com/muxinc/media-chrome/pull/448

requires a Media Chrome upgrade that makes the theme attributes (if there are any) overwrite the ones supplied by the media controller.

`stream-type` is an example of this, we still want to be able to override this and also set it before it is computed in Media Chrome.

`audio` will have to be removed as a supplied param by media-controller because we set this ourselves, that was my bad.
https://github.com/muxinc/media-chrome/blob/main/src/js/media-theme-element.js#L12

### DIFF
--- a/packages/mux-player/src/media-theme-mux.html
+++ b/packages/mux-player/src/media-theme-mux.html
@@ -305,7 +305,9 @@
 
     <template partial="TitleDisplay">
       <template if="title">
-        <media-text-display part="top title display" class="title-display"> {{title}} </media-text-display>
+        <media-text-display part="top title display" class="title-display">
+          {{title}}
+        </media-text-display>
       </template>
     </template>
 
@@ -550,7 +552,7 @@
           </media-control-bar>
         </template>
 
-        <template if="targetLiveWindow == 0">
+        <template if="targetLiveWindow == null">
           <template if="title">
             <media-control-bar>{{>TitleDisplay}}</media-control-bar>
           </template>
@@ -639,7 +641,7 @@
 
       <template if="streamType == 'live'">
 
-        <template if="targetLiveWindow == 0">
+        <template if="targetLiveWindow == null">
 
           <template if="breakpointSm == null">
             <media-control-bar slot="top-chrome">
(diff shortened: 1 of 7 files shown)


packages/mux-player/src/media-theme-mux.html
```diff
@@ -485,157 +488,269 @@
...
-        <div class="spacer"></div>
-        {{>PlaybackRateButton}} {{>CaptionsButton}} {{>AirplayButton}} {{>CastButton}} {{>PipButton}}
-        {{>FullscreenButton}}
-      </media-control-bar>
-    </template>
+    <template if="audio == null">
+
+      <template if="streamType == 'on-demand'">
+
+        <template if="breakpointSm == null">
```
cjpillsbury, line 554: suggestion(non-blocking): Add html comments here indicating the player sizes. e.g.
```html
<!-- Extra Small UI -->
```

packages/mux-player/src/template.ts
```diff
@@ -50,9 +41,9 @@ const getHotKeys = (props: MuxTemplateProps) => {
 export const content = (props: MuxTemplateProps) => html`
   <media-theme
     template="${props.theme ?? muxTemplate.content.children[0]}"
-    class="size-${props.playerSize}${props.secondaryColor ? ' two-tone' : ''}"
-    player-size="${props.playerSize ?? false}"
-    layout="${getLayout(props)}"
+    class="${props.secondaryColor ? 'two-tone' : false}"
+    stream-type="${isLiveOrDVR(props) ? 'live' : 'on-demand'}"
```
cjpillsbury, line 45: praise: Good stopgap for the migration. 👌

cjpillsbury (approved):
One minor (non-blocking) suggestion. Otherwise 👌

packages/mux-player/src/media-theme-mux.html
luwes: thanks for the review!

I was thinking similar, only downside this would be added in the player bundle so opted to not add these.

luwes:
ooh just found a small bug, changing from live to dvr and back is breaking

PR #560, opened 2023-01-31
Mux viewer count

jaredsmith:
This is a new element that shows the viewer count for the specified video.


README.md
```diff
@@ -22,6 +22,7 @@ Elements are written to make it easy to work in a variety of web frameworks. If
 | [mux-player-react](packages/mux-player-react)     | A Mux-flavored React media player component   | React         | 1.x.x 🎉                |
 | [mux-uploader](packages/mux-uploader)             | A Mux-flavored HTML5 media uploader component | Web Component | Public Beta             |
 | [mux-uploader-react](packages/mux-uploader-react) | A Mux-flavored React media uploader component | React         | Public Beta             |
+| [mux-viewer-count](packages/mux-viewer-count)     | A Mux-flavored HTML5 viewer count component   | Web Component | Public Alpha            |
```
dylanjha, line 25: I know this is gonna feel like dropping 💣 right here when this is all done but I'm wondering if `<mux-viewer-count>` might make people think it's the _total_ views count (which people have asked for).

Should we rename to:

1. `<mux-live-viewer-count>`?
2. `<mux-active-viewer-count>`?

I think num 2 

Easier to rename it now before it's published, so worth getting it right

packages/mux-viewer-count/README.md
```diff
@@ -0,0 +1,63 @@
...
+<script type="module" src="https://cdn.jsdelivr.net/npm/@mux/mux-viewer-count@0/dist/mux-viewer-count.mjs"></script>
+```
+
+## Usage
+
+`<mux-viewer-count>` has two attributes. The first is called `token`, and it should be a signed JavaScript Web Token (JWT)
+for which you want the viewer count. (For more information on creating a JWT for a particular video, see 
+https://docs.mux.com/guides/data/see-how-many-people-are-watching.
+
+The second attribute is called `pollInterval`. It specifies the number of seconds that the component should wait between
```
dylanjha, line 62: We had _many_ _many_ long convos about attr naming on Media Chrome deciding between

- concatenated, no space attrs like `pollinterval` (which is what [HTML spec](https://www.w3.org/TR/design-principles/#casing-rules) says it should be). As I call it `smushedcase`
- the more readable `poll-interval` (kebab case)

MC decided to follow HTML

The jury is still out if Mux Elements will do that too, or not. But as of today except for a few exceptions Mux Elements does `kebab-case` (if we choose for Elements to go smushedcase it will be a large breaking change that we probably wouldn't do for a while).

For the sake of consistency between elements, I think this should be `poll-interval`. Either way, we're living with `kebab-case` now on everything else, so I think it's worth staying consistent.

If we make the decision for all elements to move to smushedcase in the future, we should move them all over at the same time to minimize the inconsistency.

dylanjha (changes requested):

README.md
jaredsmith: I think #2.  Making that change now.

packages/mux-viewer-count/README.md
jaredsmith: I'll make that change!

packages/mux-active-viewer-count/src/index.ts
```diff
@@ -0,0 +1,225 @@
...
+    this.setAttribute(Attributes.TOKEN, value);
+  }
+
+  get pollInterval() {
+    let temp = this.hasAttribute(Attributes.POLL_INTERVAL)
+      ? +(this.getAttribute(Attributes.POLL_INTERVAL) as string)
+      : DEFAULT_POLL_INTERVAL;
+    if (temp < 15) {
+      console.warn('Poll interval must be at least 15 seconds, setting to default.');
+      temp = DEFAULT_POLL_INTERVAL;
```
cjpillsbury, line 137: nitpick(blocking): should probably set `temp = 15` in this case.

packages/mux-active-viewer-count/src/index.ts
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  set token(value: string) {
+    if (this.token === value) return;
+    this.#views = undefined;
+    this.setAttribute(Attributes.TOKEN, value);
+  }
+
+  get pollInterval() {
+    let temp = this.hasAttribute(Attributes.POLL_INTERVAL)
```
cjpillsbury, line 132: suggestion(fnon-blocking): We may want to just add a private `#pollInterval` and handle this via the setter instead. Can be a followup.

packages/mux-active-viewer-count/src/index.ts
jaredsmith: Fixed!

cjpillsbury (approved):
a few fast follows for cleanup, but this LGTM!

examples/vanilla-ts-esm/public/mux-active-viewer-count.html
```diff
@@ -0,0 +1,67 @@
...
+        font-size: 22px;
+      }
+      mux-uploader-drop {
+        display: block;
+        padding: 40px;
+        background: #fef9c3;
+      }
+      mux-uploader-drop[active] {
+        background: red;
+      }
```
luwes, line 31: can these be removed?

```suggestion

```

packages/mux-active-viewer-count/src/index.ts
```diff
@@ -0,0 +1,225 @@
...
+const DEFAULT_VIEWER_COUNT_VALUE = '??';
+const toAriaLabel = ({ viewerCount }: { viewerCount?: number }) => {
+  const viewerCountStr = typeof viewerCount === 'number' ? `${viewerCount}` : 'Unknown';
+  return `${ARIA_LABEL} ${viewerCountStr}`;
+};
+
+const template = document.createElement('template');
+template.innerHTML = `
+<style>
+</style>
```
luwes, line 92:
```suggestion
```

packages/mux-active-viewer-count/src/index.ts
```diff
@@ -0,0 +1,225 @@
...
+  pollInterval: number,
+  callback: (views: number) => void,
+  errorCb: (errorMsg: string) => void
+) => {
+  const url = `https://stats.mux.com/counts?token=${token}`;
+  const controller = new AbortController();
+  const signal = controller.signal;
+  let timeoutId: number | undefined;
+  let aborted = false;
+  const fetchViewerCountPoll: () => Promise<any> = async () => {
```
luwes, line 14: nit (non-blocking): if `await` is not used in the body of the function there is no need for `async` here.

packages/mux-active-viewer-count/src/index.ts
```diff
@@ -0,0 +1,225 @@
...
+        // for re-fetching by default, but provides the opportunity
+        // to unsubscribe externally via `errorCb` if desired (CJP)
+        .catch(errorCb)
+        // Wait the duration of the polling interval before restarting
+        // the next fetch
+        .then(() => {
+          return new Promise((resolve) => {
+            timeoutId = setTimeout(async () => {
+              resolve(undefined);
+            }, pollInterval * 1000);
```
luwes, line 47: here too `async` is unneeded.

could be shortened to

```suggestion
            timeoutId = setTimeout(resolve, pollInterval * 1000);
```

README.md
```diff
@@ -12,16 +12,17 @@
...
+| [mux-audio](packages/mux-audio)                             | A Mux-flavored HTML5 audio element            | Web Component | Public Beta             |
+| [mux-audio-react](packages/mux-audio-react)                 | A Mux-flavored HTML5 audio component          | React         | Public Beta             |
+| [mux-player](packages/mux-player)                           | A Mux-flavored HTML5 media player component   | Web Component | 1.x.x 🎉                |
+| [mux-player-react](packages/mux-player-react)               | A Mux-flavored React media player component   | React         | 1.x.x 🎉                |
+| [mux-uploader](packages/mux-uploader)                       | A Mux-flavored HTML5 media uploader component | Web Component | Public Beta             |
+| [mux-uploader-react](packages/mux-uploader-react)           | A Mux-flavored React media uploader component | React         | Public Beta             |
+| [mux-active-viewer-count](packages/mux-active-viewer-count) | A Mux-flavored HTML5 viewer count component   | Web Component | Public Alpha            |
```
gkatsev, line 25: I wonder if `HTML5` is needed here. Above, I believe it's there modifying the `video` and `audio` elements.

packages/mux-active-viewer-count/src/index.ts
```diff
@@ -0,0 +1,225 @@
...
+          });
+        })
+        // Restart process of re-fetching by invoking this method again
+        // (async recursion)
+        .then(fetchViewerCountPoll)
+    );
+  };
+
+  // Kick off the polling functionality.
+  fetchViewerCountPoll();
```
luwes, line 57: (not-blocking) strictly this should handle the potential rejection of the Promise so it should have a catch.
otherwise it'll have unhandled promise rejection errors

packages/mux-active-viewer-count/src/index.ts
```diff
@@ -0,0 +1,225 @@
...
+  token: string,
+  pollInterval: number,
+  callback: (views: number) => void,
+  errorCb: (errorMsg: string) => void
+) => {
+  const url = `https://stats.mux.com/counts?token=${token}`;
+  const controller = new AbortController();
+  const signal = controller.signal;
+  let timeoutId: number | undefined;
+  let aborted = false;
```
gkatsev, line 13: Theoretically, this could be simplified to rely on `signal.aborted` instead of having a separate variable tracking `aborted` here.

cjpillsbury: @gkatsev Given the "async recursion" and the location of the `errorCb` usage, I didn't see a straightforward way to not include the `aborted` flag, but I might not have been creative enough. If you see a clean refactor that won't lead to other weirdness, I'd definitely prefer to remove the flag.

packages/mux-active-viewer-count/src/index.ts
cjpillsbury: Good call. This was an artifact from before yielding the `unsubscribe` function. I agree, it can/should be removed (even if it doesn't have any real impact, the definition is "lying").

README.md
cjpillsbury: I get what you're saying, but it's at least currently also applied to `mux-uploader`:

> A Mux-flavored HTML5 media uploader component

I'd propose we keep it for now for consistency.

examples/vanilla-ts-esm/public/mux-active-viewer-count.html
jaredsmith: Removed!

packages/mux-active-viewer-count/src/index.ts
jaredsmith: Fixed!

packages/mux-active-viewer-count/src/index.ts
jaredsmith: Fixed!

README.md
jaredsmith: Leaving for now, for the sake of consistency

packages/mux-active-viewer-count/src/index.ts
gkatsev: it's definitely not a blocker, but `signal` is in scope here, so, it seems to be as simple as `s/\baborted\b/signal.aborted/`, but I haven't tried it.

cjpillsbury: Ah I see so you're proposing still relying on a flag but just using `signal.aborted` as the flag. I thought (hoped?) you saw a way to get rid of its need altogether, but your proposal should work. @jaredsmith mind refactoring and confirming that it still works? (can help with this if needed)

luwes:
shame we didn't merge this before. the `shared-polyfills` pkg is removed. 
it should be a relative `../polyfills` import now with a NPM script for copying it over

cjpillsbury:
LGTM!

luwes (approved):
👍

PR #561, opened 2023-01-31
feat: add way to set and render MC themes

luwes:
currently packaging of MC themes is done by wrapping them in a web component see
https://github.com/muxinc/media-chrome/blob/main/src/js/themes/media-theme-youtube.js

this change adds some logic that checks (when the `theme` attribute is set) if the `media-theme-x` is defined as a custom element and uses its template for theming.

test url: https://elements-demo-nextjs-git-fork-luwes-mc-theme3-mux.vercel.app/MuxPlayerTheme

### DIFF
--- a/examples/nextjs-with-typescript/pages/MuxPlayerTheme.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayerTheme.tsx
@@ -1,0 +1,50 @@
+import Link from "next/link";
+import Head from 'next/head';
+import { useRef, useState } from "react";
+import MuxPlayer, { MuxPlayerProps } from "@mux/mux-player-react";
+import "media-chrome/dist/themes/media-theme-youtube";
+
+const INITIAL_AUTOPLAY = false;
+const INITIAL_MUTED = false;
+
+function MuxPlayerPage() {
+  const mediaElRef = useRef(null);
+  const [autoplay, setAutoplay] = useState<"muted" | boolean>(INITIAL_AUTOPLAY);
+  const [muted, setMuted] = useState(INITIAL_MUTED);
+  const [paused, setPaused] = useState<boolean | undefined>(true);
+
+  return (
+    <>
+      <Head>
+        <title>&lt;MuxPlayer/&gt; (theme) Demo</title>
+      </Head>
+
+      <MuxPlayer
+        ref={mediaElRef}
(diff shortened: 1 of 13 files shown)


examples/nextjs-with-typescript/pages/MuxPlayerTheme.tsx
```diff
@@ -0,0 +1,49 @@
+import Link from "next/link";
+import Head from 'next/head';
+import { useRef, useState } from "react";
+import MuxPlayer, { MuxPlayerProps } from "@mux/mux-player-react";
+import "media-chrome/dist/themes/media-theme-youtube";
```
cjpillsbury, line 5: question(non-blocking): Does this order matter anymore? Assuming not based on the order of imports, but just confirming. Also def fine if it does for this pass regardless.

packages/mux-player/src/index.ts
```diff
@@ -806,7 +830,15 @@ class MuxPlayerElement extends VideoApiElement {
    * Get the secondary color used by the player.
    */
   get secondaryColor() {
-    return this.getAttribute(PlayerAttributes.SECONDARY_COLOR) ?? 'rgb(0 0 0 / .75)';
+    let color = this.getAttribute(PlayerAttributes.SECONDARY_COLOR);
+    if (color != null) return color;
+
+    // Fallback to computed style if no attribute is set, causes layout.
+    // https://gist.github.com/paulirish/5d52fb081b3570c81e3a
+    if (this.mediaTheme) {
```
cjpillsbury, line 838: praise: 😎

cjpillsbury:
suggestion(non-blocking): The initial render of https://elements-demo-nextjs-git-fork-luwes-mc-theme3-mux.vercel.app/MuxPlayerTheme is janky (assuming based on timing of upgrades of levels of custom elements). As a followup, we should try to resolve that (even if simply not showing the UI until "ready").

cjpillsbury (approved):
This is amazzzzziiiing! ❤️

examples/nextjs-with-typescript/pages/MuxPlayerTheme.tsx
luwes: indeed order doesn't matter anymore 👍

luwes:
you're right, I thought initially it needed fixing in the theme itself but seems better done here

PR #563, opened 2023-02-01
chore(mux-uploader): refactor status message intro own component

davekiss:


packages/mux-uploader/src/mux-uploader-status.ts
```diff
@@ -0,0 +1,77 @@
...
+import { getMuxUploaderEl } from './utils/element-utils';
+import { MuxUploaderElementEventMap, IMuxUploaderElement } from './mux-uploader';
+
+const template = document.createElement('template');
+
+template.innerHTML = `
+<style>
+
+:host([upload-error]) .status-message {
+  color: #e22c3e;
```
luwes, line 11: is this color overridable from outside the shadow DOM? 

maybe it should be 

```css
:host([upload-error]) {
  color: #e22c3e;
```

packages/mux-uploader/src/mux-uploader-status.ts
```diff
@@ -0,0 +1,77 @@
...
+      this.#uploaderEl.addEventListener('uploadstart', this.clearStatusMessage.bind(this));
+    }
+  }
+
+  disconnectedCallback() {
+    if (this.#uploaderEl) {
+      this.#uploaderEl.removeEventListener('reset', this.clearStatusMessage.bind(this));
+      this.#uploaderEl.removeEventListener('uploaderror', this.onUploadError.bind(this));
+      this.#uploaderEl.removeEventListener('success', this.onSuccess.bind(this));
+      this.#uploaderEl.removeEventListener('uploadstart', this.clearStatusMessage.bind(this));
```
luwes, line 46: (blocking): `.bind()` returns a new function each time so this will not work. the listeners will not be removed.

I recommend using `clearStatusMessage = () => {}` and then there is no need for binding...

luwes (changes requested):

packages/mux-uploader/src/mux-uploader-status.ts
luwes: see https://github.com/muxinc/media-chrome/blob/5c7acfbdd0998f6b18e8dc2c01006abfa8ce7b58/src/js/media-time-range.js#L429-L429

packages/mux-uploader/src/mux-uploader-status.ts
```diff
@@ -0,0 +1,76 @@
...
+:host([upload-error]) {
+  color: #e22c3e;
+}
+</style>
+
+<span id="status-message" role="status" aria-live="polite"></span>
+`;
+
+class MuxUploaderStatusElement extends globalThis.HTMLElement {
+  statusMessage: HTMLElement | null | undefined;
```
cjpillsbury, line 20: suggestion(non-blocking): We should probably be consistent with:

1. naming (`#uploaderEl` vs. `statusMessage`)
2. access scope (`#uploaderEl` is private, `statusMessage` is public)

Given the impact on e.g. tests, I'd recommend treating this as a followup.

packages/mux-uploader/src/mux-uploader-status.ts
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  onSuccess = () => {
+    const successMessage = 'Upload complete!';
+
+    if (this.statusMessage) {
+      this.statusMessage.innerHTML = successMessage;
+    }
+
+    console.info(successMessage);
```
cjpillsbury, line 68: suggestion(non-blocking): we use a custom logger in Mux Player along with a `debug` prop/attr. Consider removing this and (eventually) replacing with a similar logger (can likely copy over or we can move to "shared")

cjpillsbury (approved):
😎 ship it

PR #565, opened 2023-02-02
fix(playback-core): Only initialize with setupCuePoints when using hl…

cjpillsbury:
…s for playback.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -99,8 +99,8 @@ export const initialize = (props: Partial<MuxMediaPropsInternal>, mediaEl: HTMLM
   setupMux(props, mediaEl, nextHlsInstance);
   loadMedia(props, mediaEl, nextHlsInstance);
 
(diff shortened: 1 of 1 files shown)


dylanjha (approved):

packages/playback-core/src/index.ts
```diff
@@ -87,7 +99,11 @@ export const initialize = (props: Partial<MuxMediaPropsInternal>, mediaEl: HTMLM
   setupMux(props, mediaEl, nextHlsInstance);
   loadMedia(props, mediaEl, nextHlsInstance);
 
-  setupCuePoints(mediaEl);
+  // NOTE: Safari native playback behaves differently for <track>s added. For those cases,
+  // using a workaround for now (see: MuxVideoElement::addCuePoints()) (CJP)
```
gkatsev, line 103:
```suggestion
  // NOTE: Safari native playback behaves differently for <track>s added.
  // For those cases, have the track created on demand when cues are added.
```

gkatsev (approved):

PR #578, opened 2023-02-09
fix(playback-core): Typescript + min acrobatics to make svelte and ot…

cjpillsbury:
…hers happy.

### DIFF
--- a/packages/playback-core/src/hls.ts
+++ b/packages/playback-core/src/hls.ts
@@ -1,0 +1,9 @@
+// @ts-ignore
+import HlsMin from 'hls.js/dist/hls.min.js';
+import HlsClassType from 'hls.js';
+
+type Hls = typeof HlsClassType & HlsClassType;
+const Hls: Hls = HlsMin as unknown as Hls;
+
+export default Hls;
+export type HlsInterface = HlsClassType;
--- a/packages/playback-core/src/hls.ts
+++ b/packages/playback-core/src/hls.ts
@@ -1,6 +1,6 @@
 // @ts-ignore
 import HlsMin from 'hls.js/dist/hls.min.js';
-import HlsClassType from 'hls.js';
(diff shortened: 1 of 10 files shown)


packages/playback-core/src/hls.ts
```diff
@@ -0,0 +1,9 @@
+// @ts-ignore
+import HlsMin from 'hls.js/dist/hls.min.js';
+import type HlsClassType from 'hls.js';
```
gkatsev, line 3: is it possible that we're including hls.js twice?

Also, since we're not no longer using the default import, there's no benefit of marking it as externals. Maybe that's another direction? https://github.com/muxinc/elements/blob/main/packages/playback-core/package.json#L41

gkatsev: ah, just noticed that line changed to `import type`, so, we won't be including it twice, probably.

cjpillsbury: `import type` should avoid double inclusion. Let me confirm the bundles wrt `HlsMin` though.

cjpillsbury: ah you're specifically concerned about app level "double" bundles (more specifically, bundling both `hls.js/dist/hls.min.js` and `hls.js`). Yeah, for folks who have the dependency twice, that's definitely a possibility. Given that this is a stopgap solution, and given that your proposed change won't make the "double bundle" situation any better, I recommend leaving as is.

packages/playback-core/src/types.ts
```diff
@@ -1,6 +1,5 @@
 import type { Options } from 'mux-embed';
-import type Hls from 'hls.js';
-import type { HlsConfig } from 'hls.js';
```
gkatsev, line 3: question(non-blocker): should `./hls` export an `HlsConfig` type to replace this?

gkatsev (approved):

PR #580, opened 2023-02-10
Establish pattern for default slot content

davekiss:
If nothing is passed to the `rootTemplate` slot, we need to be sure to render out the default content. This PR introduces that functionality and will render default content whether the `file-select` component is used by `mux-uploader` or outside of `mux-uploader` as a standalone component.


packages/mux-uploader/src/templates/mux-uploader.ts
```diff
@@ -0,0 +1,51 @@
+import '../mux-uploader-file-select';
+import { fileSelectFragment } from '../mux-uploader-file-select';
+
```
luwes, line 5: don't forget to add `import { document } from 'shared-polyfills';` here

packages/mux-uploader/src/templates/mux-uploader.ts
```diff
@@ -0,0 +1,41 @@
...
+<input id="hidden-file-input" type="file" />
+<mux-uploader-sr-text></mux-uploader-sr-text>
+
+<mux-uploader-status></mux-uploader-status>
+<mux-uploader-retry></mux-uploader-retry>
+
+<mux-uploader-file-select>
+  <slot name="file-select">
+    ${fileSelectFragment}
```
gkatsev, line 33: I'm not following why the template for the file-select needs to be added in here as well.

davekiss: @gkatsev if we didn't do this, the `file-select` slot would still be slotted but it would have empty contents

gkatsev: Ah, right. Because a `<slot>` element was slotted, which tells the WC to not use the default stuff.

gkatsev (approved):

packages/mux-uploader/src/mux-uploader-file-select.ts
```diff
@@ -1,46 +1,48 @@
...
+import { getMuxUploaderEl } from './utils/element-utils';
+import type MuxUploaderElement from './mux-uploader';
+
+export const fileSelectFragment = /*html*/ `
+  <style>
+  #file-select {
+    cursor: pointer;
+    line-height: 16px;
+    background: var(--button-background-color, #fff);
+    border: var(--button-border, 1px solid #000000);
```
luwes, line 11: nit (nob): use `#000`

luwes (approved):
👍

PR #581, opened 2023-02-11
feat: add attribute max-resolution on mux-player and mux-video

dylanjha:
Adds `max-resolution` attribute on `<mux-player>` and `<mux-video>`. The only valid value is `720p` and that will get added as a query param to the main manifest.

```jsx
<mux-player
  playback-id="FuJSYrK0014ec2LPnm11bzC2MAySAQPqA"
  max-resolution="720p"
 ></mux-player>

<MuxPlayer
  playbackId="FuJSYrK0014ec2LPnm11bzC2MAySAQPqA"
  maxResolution="720p"
/>

<mux-video
  playback-id="FuJSYrK0014ec2LPnm11bzC2MAySAQPqA"
  max-resolution="720p"
></mux-video>

<MuxVideo
  playbackId="FuJSYrK0014ec2LPnm11bzC2MAySAQPqA"
  maxResolution="720p"
/>
```

Right now types specify that max resolution is a string, maybe best to make that an enum with the only possible value of `720p`?


packages/mux-video/src/index.ts
```diff
@@ -460,7 +462,11 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
         break;
       case Attributes.PLAYBACK_ID:
         /** @TODO Improv+Discuss - how should playback-id update wrt src attr changes (and vice versa) (CJP) */
-        this.src = toMuxVideoURL(newValue ?? undefined, { domain: this.customDomain }) as string;
+        /* @ts-ignore */
```
dylanjha, line 465: TODO: fix @dylanjha

packages/mux-video/src/index.ts
```diff
@@ -23,37 +23,39 @@ import CustomVideoElement, { VideoEvents } from './CustomVideoElement';
...
+  MAX_RESOLUTION: 'max-resolution',
+  METADATA_URL: 'metadata-url',
+  PLAYBACK_ID: 'playback-id',
+  PLAYER_SOFTWARE_NAME: 'player-software-name',
+  PLAYER_SOFTWARE_VERSION: 'player-software-version',
   PREFER_CMCD: 'prefer-cmcd',
+  PREFER_PLAYBACK: 'prefer-playback',
+  START_TIME: 'start-time',
+  STREAM_TYPE: 'stream-type',
+  TYPE: 'type',
```
dylanjha, line 58: sorted these alphabetically

packages/mux-player/REFERENCE.md
```diff
@@ -15,6 +15,7 @@
...
 | `prefer-playback`             | `"mse" \| "native"`                                                                                         | Specify if Mux Player should try to use Media Source Extension or native playback (if available). If no value is provided, Mux Player will choose based on what's deemed optimal for content and playback environment.                                                                                                                                                                           | Varies        |
+| `max-resolution`              | `"720p"                                                                                                      | Specify if Mux Player should try to use Media Source Extension or native playback (if available). If no value is provided, Mux Player will choose based on what's deemed optimal for content and playback environment.                                                                                                                                                                          | ""        |
```
cjpillsbury, line 18: nitpick(blocking): `"720p"` should be `string`(progressive scan resolution) or similar. Also, description is inaccurate (copypasta).

cjpillsbury:
> Right now types specify that max resolution is a string, maybe best to make that an enum with the only possible value of `720p`?

I'd recommend keeping it generic unless/until we know there will be server-side constraints on this. For simplicity, we may want to just treat it as a number (height), e.g. `max-resolution="720"`.

cjpillsbury:
suggestion(non-blocking): Add this to the NextJS demo page for MuxPlayer for behavior testing.

packages/mux-player/REFERENCE.md
gkatsev: "progressive scan resolution" is too esoteric. Should just be "height of the video in pixels" or something like that. Ties into your suggestion that this should just be a number, too.

dylanjha: > height of the video in pixels

This is tough because it's actually not height. A vertical video where the height is 1280 and the width is 720 would still fall under this. It's calculated by total pixels being`<= 921,600` which is 1280x720.

The only possible value we allow is `720p`, so I wouldn't want to give people the impression that they could specify any value they want.

I took at stab at updated, lmk your thoughts

dylanjha:
> suggestion(non-blocking): Add this to the NextJS demo page for MuxPlayer for behavior testing.

done!

packages/mux-player-react/REFERENCE.md
```diff
@@ -35,6 +35,7 @@
...
 | `preferPlayback`           | `"mse" \| "native"`                                                                                                   | Specify if Mux Player should try to use Media Source Extension or native playback (if available). If no value is provided, Mux Player will choose based on what's deemed optimal for content and playback environment.                                                                                                                                                                           | Varies                     |
+| `maxResolution`              | `"720p"                                                                                                             | Specify the max-resolution you want delivered for this video. The only valid option is the string `720p`  | ""        |
```
cjpillsbury, line 38:
```suggestion
| `maxResolution`              | `"720p"`                                                                                                             | Specify the max-resolution you want delivered for this video. The only valid option is the string `720p`  | ""        |
```

packages/mux-player/src/helpers.ts
```diff
@@ -16,20 +16,19 @@ export const getPlayerVersion = () => player_version;
 
 export const getSrcFromPlaybackId = (
   playbackId?: string,
-  { token, domain = MUX_VIDEO_DOMAIN }: { token?: string; domain?: string } = {}
+  { maxResolution, token, domain = MUX_VIDEO_DOMAIN }: { maxResolution?: string; token?: string; domain?: string } = {}
```
cjpillsbury, line 19: suggestion(non-blocking): If `maxResolution` is going to be a set of string literals, either now or as a followup, we should probably type it as such (See `stream-type`/`streamType` and corresponding maps + typings for an example: https://github.com/muxinc/elements/blob/main/packages/mux-video/src/index.ts#L327-L341 & https://github.com/muxinc/elements/blob/main/packages/playback-core/src/types.ts#L39-L55)

packages/mux-player/REFERENCE.md
```diff
@@ -15,6 +15,7 @@
...
 | `prefer-playback`             | `"mse" \| "native"`                                                                                         | Specify if Mux Player should try to use Media Source Extension or native playback (if available). If no value is provided, Mux Player will choose based on what's deemed optimal for content and playback environment.                                                                                                                                                                           | Varies        |
+| `max-resolution`              | `"720p"                                                                                                     | Specify the max-resolution you want delivered for this video. The only valid option is the string `720p`  | ""        |
```
cjpillsbury, line 18:
```suggestion
| `max-resolution`              | `"720p"`                                                                                                    | Specify the max-resolution you want delivered for this video. The only valid option is the string `720p`  | ""        |
```

packages/mux-video/README.md
```diff
@@ -86,6 +86,7 @@ Attributes:
 
 <img src="../../images/env-key.png" alt="Mux Data dashboard env key"></img>
 
+| `max-resolution`: This can be used to limit the resolution of the video delivered. When using signed URLs this attribute will not work and instead you will need to include the `max_resolution` parameter in your signed token.
```
cjpillsbury, line 89: praise: Good callout, re: signed token.

packages/mux-player/src/helpers.ts
```diff
@@ -16,20 +16,19 @@ export const getPlayerVersion = () => player_version;
...
 ) => {
   /*
-   * 2022-04-01 djhaveri
+   * 2023-02-11 djhaveri
```
luwes, line 22: can we use `git blame` for this? 

if we all start adding our username / date it's gonna get noisy

packages/playback-core/src/index.ts
```diff
@@ -53,10 +53,18 @@ export const generatePlayerInitTime = () => {
...
-  return `https://stream.${domain}/${idPart}.m3u8${queryPart}`;
+  let queryWithMaxRes;
+  if (maxResolution) {
+    if (queryPart && queryPart.length) {
+      queryWithMaxRes = `${queryPart}&max_resolution=${maxResolution}`;
+    } else {
+      queryWithMaxRes = `?max_resolution=${maxResolution}`;
+    }
+  }
+  return `https://stream.${domain}/${idPart}.m3u8${queryWithMaxRes || queryPart}`;
```
luwes, line 67: can we make use of [URLSearchParams()](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams) ?

we shouldn't have to need conditionals here

luwes (changes requested):
overall looks great.
added 1 blocker because I don't think we'll come back to that if we don't do it now

packages/playback-core/src/index.ts
gkatsev: potentially in conjunction with [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) as well

luwes (approved):

packages/playback-core/src/index.ts
```diff
@@ -53,10 +53,14 @@ export const generatePlayerInitTime = () => {
...
-export const toMuxVideoURL = (playbackId?: string, { domain = MUX_VIDEO_DOMAIN } = {}) => {
+export const toMuxVideoURL = (playbackId?: string, { domain = MUX_VIDEO_DOMAIN, maxResolution = '' } = {}) => {
   if (!playbackId) return undefined;
   const [idPart, queryPart = ''] = toPlaybackIdParts(playbackId);
-  return `https://stream.${domain}/${idPart}.m3u8${queryPart}`;
+  const url = new URL(`https://stream.${domain}/${idPart}.m3u8${queryPart}`);
+  if (maxResolution) {
+    url.searchParams.set('max_resolution', maxResolution);
+  }
+  return url.toString();
```
luwes, line 63: 👍

dylanjha: much better!

cjpillsbury (approved):
🎉

PR #586, opened 2023-02-18
Cross browser tests

gkatsev:
Fixes the big blocker for running tests in Firefox and Safari. This is an error from usage of ResizeObserver in media-chrome. For whatever reason, Firefox and Safari throw an error around the loop limit being exceeded. Currently having those errors ignored in `player.test.js`.

Enabled running tests in firefox and webkit via `playwright`. Can be done by running `yarn test --all`.

TODO, in no particular order:
- [x] setup tests on CI with playwright https://modern-web.dev/docs/test-runner/browser-launchers/playwright/#using-with-github-actions
- [x] fix outstanding tests in safari

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,8 @@ jobs:
       # esbuild requires --ignore-scripts to NOT be added here.
       - run: yarn install --frozen-lockfile
       - run: yarn clean
+      - name: Install playwright browsers
+        run: npx playwright install --with-deps
       # run yarn and pass through `-- --all` to turborepo
       - run: yarn test -- -- --all
       - name: Upload coverage to Codecov
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,8 @@ jobs:
       # esbuild requires --ignore-scripts to NOT be added here.
       - run: yarn install --frozen-lockfile
       - run: yarn clean
-      - run: yarn test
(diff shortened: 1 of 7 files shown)


gkatsev:
running all 3 via playwright takes ~60s, compared to ~25 in just default chrome. It'll likely improve when the tests are fixed because there were a bunch of timeout errors.

luwes:
awesome! 
60s is not too bad 
maybe something can also be done with Turborepo, I don't think we use the cache properly yet in GH actions

gkatsev:
turborepo could be used to reduce the overall build time on CI, but probably not the test times themselves. The 60s is from running `time yarn test --all`. I think once the tests are fixed it'll be less since a bunch of the tests timed out. Since I have it set up to run the browsers in parallel, it should be as fast as the slowest browser. On CI, we could even potentially set it up in a matrix to run each browser in a separate build.

gkatsev:
With the fixed tests, they now run in ~30s, so, only about 5 seconds slower for all 3 browsers!

gkatsev:
Also, can run in specific browsers with playwright, via the built-in flags like `yarn test --playwright --browsers webkit chromium`

gkatsev:
It also works with `--watch` flag.

cjpillsbury (approved):
probably will want to make some of this stuff cleaner to impl but I don't think any of that should block merging this pr. Approved / thanks for the hard work

PR #592, opened 2023-02-27
feat: inferred stream type

cjpillsbury:
Implementing inferred stream types and related (`targetLiveWindow`, `liveEdgeStart`).

This PR removes the requirement for setting stream type, which can now be inferred directly from the media. Note that, since this process is asynchronous and does not begin until the media metadata is requested (aka will not immediately begin if `preload="none"`), this means that, for our current UI designs, folks will see an "on demand" UI until the `streamType` and related has been inferred (at least by default).

Additionally, for this iteration, we will continue to support/translate "invalid" stream types to their appropriate corresponding properties/attributes. See https://github.com/cjpillsbury/elements/blob/feat/inferred-stream-type/errors/deprecated-stream-type.md for how to migrate now-deprecated stream types.

Finally, there is now a distinct `default-stream-type`/`defaultStreamType` attribute/property for cases where we want a different UI to show by default. I've treated any corresponding "`default-target-live-window`" attr/prop as out of scope for this iteration, as there is no corresponding implementation within Media Chrome.


packages/mux-player/src/media-theme-mux.html
```diff
@@ -265,6 +265,7 @@
     }
   </style>
   <media-controller
+    default-stream-type="on-demand"
```
cjpillsbury, line 268: should we allow this to be user-configurable?

packages/mux-player/src/media-theme-mux.html
```diff
@@ -530,7 +531,7 @@
           </media-control-bar>
         </template>
 
-        <template if="targetLiveWindow == null">
+        <template if="targetLiveWindow == 0">
```
cjpillsbury, line 534: note: This change is based on the m-ui-ext spec (See: https://github.com/video-dev/media-ui-extensions/issues/4#issuecomment-1344924246). We may want to discuss theme compatibility complexities as a result. i.e.: This is no longer compatible with slotted media elements that don't implement `targetLiveWindow`. Fine for mux-video/mux-player but has implications for our new themes that "live" in media-chrome (e.g. micro theme)

packages/playback-core/src/index.ts
```diff
@@ -331,12 +354,39 @@ export const loadMedia = (
...
+          const mediaPlaylistUrl = multivariantPlaylistStr.split('\n').find((_line, idx, lines) => {
+            return idx && lines[idx - 1].startsWith('#EXT-X-STREAM-INF');
+          }) as string;
+
+          return fetch(mediaPlaylistUrl).then((resp) => resp.text());
+        })
+        .then((mediaPlaylistStr) => {
+          const typeLine = mediaPlaylistStr.split('\n').find((line) => line.startsWith('#EXT-X-PLAYLIST-TYPE')) ?? '';
+
+          const playlistType = typeLine.split(':')[1]?.trim() as HlsPlaylistTypes;
```
cjpillsbury, line 376: the code here and in the hls scenario is identical. Break into a shared function?

yarn.lock
```diff
@@ -10595,10 +10595,10 @@ mdn-data@2.0.4:
   resolved "https://registry.yarnpkg.com/mdn-data/-/mdn-data-2.0.4.tgz#699b3c38ac6f1d728091a64650b65d388502fd5b"
   integrity sha512-iV3XNKw06j5Q7mi6h+9vbx23Tv7JkjEVgKHW4pimwyDGWm0OIQntJJ+u1C6mg6mK1EaTv42XQ7w76yuzH7M2cA==
 
-media-chrome@0.18.4:
-  version "0.18.4"
-  resolved "https://registry.yarnpkg.com/media-chrome/-/media-chrome-0.18.4.tgz#50438f0635d9027a5d23c69e8baf75358b22dfcd"
-  integrity sha512-ftG68jxSUraibyjOhyNEueEW94jVA87wJRYPylS8b/k6/nm07yURRPmk0O321NaI+eQ6i15wpA8UFgKSoBOqnQ==
+media-chrome@0.18.5-canary.5-2ed0630:
```
cjpillsbury, line 10598: locally using a linked version of media-chrome pending a canary/"official" release of https://github.com/muxinc/media-chrome/pull/480

packages/mux-player/src/index.ts
```diff
@@ -700,24 +707,22 @@ class MuxPlayerElement extends VideoApiElement {
    * we aren't an audio player and the stream-type isn't live.
    */
   get storyboard() {
+    // If the storyboardSrc has been explicitly set, assume it should be used
```
cjpillsbury, line 710: Intentionally changed the rules here a bit to always prefer the `storyboardSrc` if it's explicitly set to unlock e.g. `audio` themes that have storyboards.

packages/mux-player/src/template.ts
```diff
@@ -41,7 +43,7 @@ const getHotKeys = (props: MuxTemplateProps) => {
 export const content = (props: MuxTemplateProps) => html`
   <media-theme
     template="${props.themeTemplate ?? muxTemplate.content.children[0]}"
-    stream-type="${isLiveOrDVR(props) ? 'live' : 'on-demand'}"
```
luwes, line 44: (blocking): I think we still require this here because when the user manually sets the stream-type on mux-player that one should be used.

luwes (changes requested):

packages/mux-player/src/template.ts
cjpillsbury: for sure/good callout. I removed all of the explicit setting until I resolved all of the assumptions so everything could work with the inferred case. With the new architecture/latest push, I'm solving it slightly differently by overriding it on `<mux-video>` and letting the values "flow" through media chrome. I have, however, added a `default-stream-type` to correspond with overriding the newly added corresponding attribute in `<media-controller>`.

packages/mux-player/src/template.ts
```diff
@@ -78,7 +82,6 @@ export const content = (props: MuxTemplateProps) => html`
       player-software-name="${props.playerSoftwareName ?? false}"
       player-software-version="${props.playerSoftwareVersion ?? false}"
       env-key="${props.envKey ?? false}"
-      stream-type="${props.streamType ?? false}"
       custom-domain="${props.customDomain ?? false}"
       src="${!!props.src
```
cjpillsbury, line 86: still need to solve `cast-stream-type` below

packages/mux-player/src/template.ts
```diff
@@ -41,8 +43,8 @@ const getHotKeys = (props: MuxTemplateProps) => {
 export const content = (props: MuxTemplateProps) => html`
   <media-theme
     template="${props.themeTemplate ?? muxTemplate.content.children[0]}"
-    stream-type="${isLiveOrDVR(props) ? 'live' : 'on-demand'}"
-    target-live-window="${isDVR(props) ? 1 : false}"
+    default-stream-type="${props.defaultStreamType}"
+    class="${props.secondaryColor ? 'two-tone' : false}"
```
cjpillsbury, line 47: even though this was broken in the old impl, should fix the "seek to live on play" for "simulated live" cases.

packages/mux-player/src/media-theme-mux.html
cjpillsbury: current implementation answers: yes

packages/mux-player/src/template.ts
cjpillsbury: Resolved by adding support to `<castable-video>` and migrating to the "real" version 🎉!

packages/mux-player/src/media-theme-mux.html
cjpillsbury: Solved in media chrome per recommendation 🎉

packages/mux-player/src/template.ts
cjpillsbury: Resolved for "DVR". Will treat as followup for "simulated live".

packages/mux-player/src/template.ts
```diff
@@ -41,10 +36,11 @@ const getHotKeys = (props: MuxTemplateProps) => {
 export const content = (props: MuxTemplateProps) => html`
   <media-theme
     template="${props.themeTemplate ?? muxTemplate.content.children[0]}"
-    stream-type="${isLiveOrDVR(props) ? 'live' : 'on-demand'}"
-    target-live-window="${isDVR(props) ? 1 : false}"
+    default-stream-type="${props.defaultStreamType ?? false}"
+    class="${props.secondaryColor ? 'two-tone' : false}"
     hotkeys="${getHotKeys(props) || false}"
     nohotkeys="${props.noHotKeys || !props.hasSrc || props.isDialogOpen || false}"
+    noautoseektolive="${props.targetLiveWindow !== 0}"
```
cjpillsbury, line 43: disables "auto seek to live" on play requests for non "standard live" cases. This fixes an issue with pause+play for "DVR" but does not fully resolve "simulated live" cases (where the content is "on demand" but we want to "treat it as live," which were likewise already broken).

errors/invalid-stream-type.md
```diff
@@ -9,11 +9,13 @@ The provided `stream-type` is invalid.
 If you are a viewer of this video there is not much you can do. The owner of the
 video will have to fix this issue.
 
-If you are the owner of this video make sure the `stream-type` is one of the
-following: `on-demand`, `live`, `ll-live`, `live:dvr`, or `ll-live:dvr`.
+If you are the owner of this video, `stream-type`is no longer required, so you may consider removing it.
+If you still have a use case where you would like to explicitly define the `stream-type`, make sure it or
+`default-stream-type` is one of the following: `on-demand`, `live`.
```
luwes, line 14: `it or` is quick to read over 

```suggestion
If you are the owner of this video, `stream-type` is no longer required, so you may consider removing it.
If you still have a use case where you would like to explicitly define the `stream-type`, make sure the `stream-type` or
`default-stream-type` attribute is one of the following: `on-demand`, `live`.
```

packages/mux-player/README.md
```diff
@@ -30,7 +30,6 @@ yarn add @mux/mux-player
   playback-id="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"
   metadata-video-title="Big Buck Bunny"
   metadata-viewer-user-id="user-id-1234"
-  stream-type="on-demand"
```
luwes, line 33: ooh yeah!

packages/mux-player/src/index.ts
```diff
@@ -379,7 +386,7 @@ class MuxPlayerElement extends VideoApiElement {
       // - native fullscreen on iPhones
       return (
         this.streamType &&
-        [StreamTypes.LIVE, StreamTypes.LL_LIVE].includes(this.streamType as any) &&
+        [StreamTypes.LIVE].includes(this.streamType as any) &&
```
luwes, line 389: nit (nonb): this.streamType === StreamTypes.LIVE

cjpillsbury, line 389: lol this and below should probably just be a strict equality check... will refactor.

packages/mux-player/src/index.ts
```diff
@@ -540,26 +547,33 @@ class MuxPlayerElement extends VideoApiElement {
...
+      case MuxVideoAttributes.STREAM_TYPE: {
+        if (newValue && ![StreamTypes.LIVE, StreamTypes.ON_DEMAND, StreamTypes.UNKNOWN].includes(newValue as any)) {
+          // Handle deprecated values by translating to new properties for the time being.
+          // NOTE: The value of `streamType` / `stream-type` will be translated at the template
+          // level. See template.ts for more information (CJP).
+          if (['ll-live', 'live:dvr', 'll-live:dvr'].includes(this.streamType as any)) {
+            logger.devlog({
+              file: 'deprecated-stream-type.md',
+              message: i18n(
+                `The stream type is currently deprecated: \`{streamType}\`. Please provide stream-type as either: \`on-demand\`, \`live\`. For DVR, please use \`target-live-window="Infinity"\``
```
luwes, line 561: `currently` seems confusing?

```suggestion
                `The stream type is deprecated: \`{streamType}\`. Please provide stream-type as either: \`on-demand\`, \`live\`. For DVR, please use \`target-live-window="Infinity"\``
```

packages/mux-player/src/index.ts
```diff
@@ -1081,6 +1096,47 @@ class MuxPlayerElement extends VideoApiElement {
...
+
+    if (typeof val === 'number') {
+      this.setAttribute(PlayerAttributes.TARGET_LIVE_WINDOW, `${val}`);
+    } else {
+      this.removeAttribute(PlayerAttributes.TARGET_LIVE_WINDOW);
+    }
+  }
+
+  get liveEdgeStart() {
+    return this.media?.liveEdgeOffset;
```
cjpillsbury, line 1137: issue(blocker): this is wrong. Will update.

packages/playback-core/src/index.ts
```diff
@@ -87,6 +98,31 @@ export const getError = (mediaEl: HTMLMediaElement) => {
...
 
+export const getStreamType = (mediaEl: HTMLMediaElement) => {
+  return muxMediaState.get(mediaEl)?.streamType ?? StreamTypes.UNKNOWN;
+};
+
+export const getTargetLiveWindow = (mediaEl: HTMLMediaElement) => {
+  return muxMediaState.get(mediaEl)?.targetLiveWindow ?? Number.NaN;
+};
+
+// export const getLowLatency = (mediaEl: HTMLMediaElement) => {
```
cjpillsbury, line 109: nitpick(blocking): Not needed. remove.

packages/playback-core/src/index.ts
```diff
@@ -204,24 +240,14 @@ export const setupHls = (
 
 export const getStreamTypeConfig = (streamType?: ValueOf<StreamTypes>) => {
   // for regular live videos, set backBufferLength to 8
-  if ([StreamTypes.LIVE, StreamTypes.DVR].includes(streamType as any)) {
+  if ([StreamTypes.LIVE].includes(streamType as any)) {
```
cjpillsbury, line 243: nitpick(blocking): refactor to simple equality compare

packages/playback-core/src/index.ts
```diff
@@ -332,12 +358,74 @@ export const loadMedia = (
...
+                // so parsing and casting presumptuously here. See spec link above for more info. (CJP)
+                const targetDuration = +targetDurationLine.split(':')[1];
+                liveEdgeStartOffset = targetDuration * 3;
+              }
+              (muxMediaState.get(mediaEl) ?? {}).liveEdgeStartOffset = liveEdgeStartOffset;
+            }
+          });
+      };
+
+      if (mediaEl.preload === 'none') {
```
cjpillsbury, line 423: respect preload prefs here for native playback cases.

packages/mux-video/src/index.ts
```diff
@@ -501,6 +554,21 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
             .catch(() => console.error(`Unable to load or parse metadata JSON from metadata-url ${newValue}!`));
         }
         break;
+      case Attributes.STREAM_TYPE:
+        // If the newValue is unset
+        if (newValue == null || newValue !== oldValue) {
+          this.dispatchEvent(
+            new CustomEvent('streamtypechange', { composed: true, bubbles: true, detail: this.streamType })
```
luwes, line 561: native video events don't have detail / data attached. not sure if it's the best idea to change expectations for our custom events (and I see I did that too for the error event 😄🔫) 

```suggestion
            new CustomEvent('streamtypechange', { composed: true, bubbles: true })
```

cjpillsbury: Accepting change, but I believe this one should be a followup conversation. It is valid to have native events with specific instance-based properties (see many examples here https://developer.mozilla.org/en-US/docs/Web/API/Event#interfaces_based_on_event), but, if we want to implement those (for media-ui-extension proposals like this, for example), we should probably avoid `CustomEvent` and instead implement our own event class

packages/playback-core/src/index.ts
```diff
@@ -332,12 +358,74 @@ export const loadMedia = (
...
+                // This is required for all media playlists. See: https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-12#section-4.4.3.1 (CJP)
+                const targetDurationLine = playlistLines.find((line) =>
+                  line.startsWith('#EXT-X-TARGETDURATION')
+                ) as string;
+                // EXT-X-TARGETDURATION has exactly one unnamed attribute that represents the target duration value, which is required,
+                // so parsing and casting presumptuously here. See spec link above for more info. (CJP)
+                const targetDuration = +targetDurationLine.split(':')[1];
+                liveEdgeStartOffset = targetDuration * 3;
+              }
+              (muxMediaState.get(mediaEl) ?? {}).liveEdgeStartOffset = liveEdgeStartOffset;
```
luwes, line 418: would be really good to have some tests here of parsing out the values from the m3u8

cjpillsbury: cool good call.

packages/playback-core/src/index.ts
```diff
@@ -332,12 +358,74 @@ export const loadMedia = (
...
     | 'subtitleTracks'
     | 'subtitleTrack'
+    | 'userConfig'
   >
 ) => {
   const shouldUseNative = useNative(props, mediaEl);
   const { src } = props;
   if (mediaEl && shouldUseNative) {
     if (typeof src === 'string') {
+      const getStreamInfo = () => {
```
gkatsev, line 368: This method seems like it should be in a separate module and accepts a URL to fetch. Even better would be to separate it out into multiple methods. Methods that do the fetch and methods that do the data processing. It would then make writing tests for the data processing methods a lot easier, since they won't be all inline and could short-circuit the fetching with hard-coded data.

packages/playback-core/src/util.ts
```diff
@@ -69,6 +70,16 @@ export const getType = (props: Partial<Pick<MuxMediaProps, 'type' | 'src'>>) =>
...
 };
 
+export const toStreamTypeFromPlaylistType = (playlistType: HlsPlaylistTypes) => {
+  return playlistType === 'VOD' ? StreamTypes.ON_DEMAND : StreamTypes.LIVE;
+};
+
+export const toTargetLiveWindowFromPlaylistType = (playlistType: HlsPlaylistTypes) => {
+  if (playlistType === 'EVENT') return Number.POSITIVE_INFINITY;
+  if (playlistType === 'VOD') return Number.NaN;
+  return 0;
```
gkatsev, line 80: 👍🏻

packages/mux-video/package.json
```diff
@@ -49,7 +49,8 @@
     "publish-release": "../../scripts/publish.sh"
   },
   "dependencies": {
-    "@mux/playback-core": "0.17.0"
+    "@mux/playback-core": "0.17.0",
+    "castable-video": "^0.5.2"
```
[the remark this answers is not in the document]
cjpillsbury: I think it's a dependency, at least without changing expectations. Before we were relying on our local copy of the element, so this feels like the "right" equivalent. cc @luwes for thoughts.

packages/playback-core/src/index.ts
cjpillsbury: @gkatsev Thoughts on sharing code between this and the hls.js permutation? There's def overlap, but folks may find the functions a bit too abstracted at that point?

dylanjha:
I think in order to get this work merged we should:

- Not make any API changes. We should keep advising folks to use the existing stream types (`on-demand`, `live`, `ll-live`, `live:dvr`, `ll-live:dvr`)
- Under the hood, we parse those stream types and do whatever we need to

If no stream-type is provided:
- The player will show `on-demand` while it resolves the correct stream-type. When the correct stream-type is resolved then the UI will be updated accordingly.
- We all agree this is not ideal, which is why we are still advising people to provide stream-type and keeping our examples & documentation consistent with that.

Later on, we'll solve the problem of preventing controls from doing a show/hide flash while the stream-type is being inferred. When that happens, THEN we can tell people `stream-type` is not required.

packages/mux-player/src/index.ts
```diff
@@ -1194,7 +1204,10 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
    * Get stream type.
    */
   get streamType() {
-    return getVideoAttribute(this, MuxVideoAttributes.STREAM_TYPE);
+    if (this.hasAttribute(MuxVideoAttributes.STREAM_TYPE)) {
+      return this.getAttribute(MuxVideoAttributes.STREAM_TYPE);
+    }
+    return this.media?.streamType ?? StreamTypes.UNKNOWN;
```
luwes, line 1210:
```suggestion
    return this.getAttribute(MuxVideoAttributes.STREAM_TYPE) ?? this.media?.streamType ?? StreamTypes.UNKNOWN;
```

a bit shorter

packages/mux-player/src/index.ts
```diff
@@ -1204,6 +1217,47 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
...
+      return +(this.getAttribute(PlayerAttributes.TARGET_LIVE_WINDOW) as string) as number;
+    }
+    return this.media?.targetLiveWindow ?? Number.NaN;
+  }
+
+  set targetLiveWindow(val: number | undefined) {
+    // don't cause an infinite loop and avoid change event dispatching
+    if (val === this.targetLiveWindow) return;
+
+    if (typeof val === 'number') {
```
luwes, line 1250: (blocking): use isNumeric here. a string number should be okay

packages/mux-video/src/index.ts
```diff
@@ -340,12 +348,15 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
   }
 
   get streamType(): ValueOf<StreamTypes> | undefined {
-    // getAttribute doesn't know that this attribute is well defined. Should explore extending for MuxVideo (CJP)
-    return (this.getAttribute(Attributes.STREAM_TYPE) as ValueOf<StreamTypes>) ?? undefined;
+    // Allow overriding inferred `streamType`
+    if (this.hasAttribute(Attributes.STREAM_TYPE)) {
+      return this.getAttribute(Attributes.STREAM_TYPE) as ValueOf<StreamTypes>;
+    }
+    return getStreamType(this.nativeEl);
```
luwes, line 355:
```suggestion
    return (this.getAttribute(Attributes.STREAM_TYPE) as ValueOf<StreamTypes>) ?? getStreamType(this.nativeEl);
```

packages/mux-video/src/index.ts
```diff
@@ -355,6 +366,55 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
+  get liveEdgeOffset() {
+    if (!this.hasAttribute(Attributes.LIVE_EDGE_OFFSET)) return undefined;
+    return +(this.getAttribute(Attributes.LIVE_EDGE_OFFSET) as string) as number;
+  }
+
+  set liveEdgeOffset(val: number | undefined) {
+    // don't cause an infinite loop and avoid change event dispatching
+    if (val === this.liveEdgeOffset) return;
+
+    if (typeof val === 'number') {
```
luwes, line 407: same numeric check

packages/mux-video/test/index.test.js
```diff
@@ -256,4 +256,84 @@ describe('<mux-video>', () => {
...
+
+    it('infers liveEdgeStart >= 0 for live content', async () => {
+      const playbackId = 'v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM';
+      const muxVideoEl = await fixture(`<mux-video
+        playback-id="${playbackId}"
+        preload="metadata"
+      ></mux-video>`);
+      // Wait for this event simply to guarantee inferred values have been computed
+      await oneEvent(muxVideoEl, 'streamtypechange');
+      assert(muxVideoEl.liveEdgeStart >= 0, 'liveEdgeStart should be a positive number');
```
luwes, line 320: 👍

packages/playback-core/src/index.ts
```diff
@@ -41,9 +49,143 @@ export {
...
+
+  (muxMediaState.get(mediaEl) ?? {}).liveEdgeStartOffset = liveEdgeStartOffset;
+
+  (muxMediaState.get(mediaEl) ?? {}).targetLiveWindow = targetLiveWindow;
+  mediaEl.dispatchEvent(
+    new CustomEvent('targetlivewindowchange', { composed: true, bubbles: true, detail: targetLiveWindow })
+  );
+
+  (muxMediaState.get(mediaEl) ?? {}).streamType = streamType;
+  mediaEl.dispatchEvent(new CustomEvent('streamtypechange', { composed: true, bubbles: true, detail: streamType }));
```
luwes, line 115: I wouldn't add `detail` here because that will never be added to the native HTML video tag

packages/mux-video/src/index.ts
```diff
@@ -501,6 +559,19 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
+      case Attributes.STREAM_TYPE:
+        // If the newValue is unset
+        if (newValue == null || newValue !== oldValue) {
+          this.dispatchEvent(new CustomEvent('streamtypechange', { composed: true, bubbles: true }));
+        }
+        break;
+      case Attributes.TARGET_LIVE_WINDOW:
+        if (newValue == null || newValue !== oldValue) {
+          this.dispatchEvent(
+            new CustomEvent('targetlivewindowchange', { composed: true, bubbles: true, detail: this.targetLiveWindow })
```
luwes, line 571: I wouldn't add `detail` here because that will never be added to the native HTML video tag

packages/playback-core/src/index.ts
```diff
@@ -41,9 +49,143 @@ export {
...
+    const seekable: TimeRanges = Object.freeze({
+      get length() {
+        return mediaEl.seekable.length;
+      },
+      start(index: number) {
+        return mediaEl.seekable.start(index);
+      },
+      end(index: number) {
+        if (index > this.length) return mediaEl.seekable.end(index);
+        return hls.liveSyncPosition ?? mediaEl.seekable.end(index);
```
luwes, line 168: is this correct? all indexes will return liveSyncPosition if it exists

packages/playback-core/src/index.ts
luwes: maybe check readyState or networkState here to be sure `loadstart` will fire in the future

packages/playback-core/src/index.ts
cjpillsbury: Yes that's by design per https://github.com/video-dev/media-ui-extensions/blob/main/proposals/0007-live-edge.md#property-constraint-on-htmlmediaelementseekableend-to-model-seekable-live-edge

packages/mux-video/src/index.ts
```diff
@@ -355,6 +366,55 @@ class MuxVideoElement extends CustomVideoElement<HTMLVideoElement> implements Pa
...
+      return +(this.getAttribute(Attributes.TARGET_LIVE_WINDOW) as string) as number;
+    }
+    return getTargetLiveWindow(this.nativeEl);
+  }
+
+  set targetLiveWindow(val: number | undefined) {
+    // don't cause an infinite loop and avoid change event dispatching
+    if (val === this.targetLiveWindow) return;
+
+    if (typeof val === 'number') {
```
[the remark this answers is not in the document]
cjpillsbury: updated in slightly different way per out of band convo

packages/mux-player/src/index.ts
cjpillsbury: updated in slightly different way per out of band convo

luwes (approved):
LGTM!

one thing I also have to remind myself of is keep the PR small enough. 
ideally something like half the size of this, sometimes it's difficult but this one could have been split up in a few I think. would also help with merge conflicts. faster reviews / merges

gkatsev (approved):
Tested and verified that changing the stream type from DVR to others is fixed.

errors/deprecated-stream-type.md
```diff
@@ -0,0 +1,23 @@
+# Deprecated Stream Type
```
dylanjha, line 1: (non-blocking) -- open to other opinions.

Can we put this on a branch for safe keeping? Feels weird to have it in the repo if it's never linked.

Might also want to put the commented code on a branch. Feels a little sloppy to have commented out code making it into `main`

PR #595, opened 2023-02-28
fix: switch to media-captions-selectmenu

gkatsev:
Depends on https://github.com/muxinc/media-chrome/pull/471


packages/mux-player/src/styles.css
```diff
@@ -112,7 +109,7 @@ media-theme {
   display: var(--media-controls, var(--media-mute-button));
 }
 ::part(captions menu-button) {
-  display: var(--media-controls, var(--media-captions-menu-button, contents));
```
gkatsev, line 115: the default display value is no longer `contents`

packages/mux-player/package.json
```diff
@@ -52,7 +52,7 @@
   "dependencies": {
     "@mux/mux-video": "0.14.0",
     "@mux/playback-core": "0.17.0",
-    "media-chrome": "0.18.4"
```
gkatsev, line 55: PR is blocked until the next media-chrome release.

cjpillsbury:
suggestion(non-blocking): Update css vars in nextjs MuxPlayer page. Update REFERENCE.md for parts + css vars.

gkatsev:
The failing tests are the `Error: Timeout: playback did not start inLiveWindow` tests.

packages/mux-player/src/media-theme-mux.html
```diff
@@ -379,9 +379,9 @@
     </template>
 
     <template partial="CaptionsMenuButton">
-      <media-captions-menu-button
+      <media-captions-selectmenu
         default-showing="{{defaultShowingCaptions}}"
-        part="bottom captions menu-button"
```
gkatsev, line 384: Updated from `menu-button` to select menu. I also added `button`, but should I remove it?

gkatsev:
@cjpillsbury thanks. Forgot that existed. Updated. That also made me notice that the parts needed to be updated. I re-added `button`, but maybe I shouldn't?

packages/mux-player/src/media-theme-mux.html
cjpillsbury: I'd say no, it shouldn't have a `button` part, but can we export the underlying media-captions-button as its own part in mux player? To channel prior convos we had w/@heff, would be good for folks to not need to know implementation details of the selectmenu

packages/mux-player/src/styles.css
cjpillsbury: praise: glad you caught this

packages/mux-player/src/media-theme-mux.html
gkatsev: If we switch from captions-selectmenu to chrome-selectmenu and slot the captions-button manually, we could export it as a `button`

cjpillsbury: ah right it's in the light DOM now. We could also do it programmatically (since it's on the light DOM wrt `media-captions-selectmenu`) but that feels a bit too scope creepy for this iteration.

packages/mux-player/src/media-theme-mux.html
```diff
@@ -58,16 +58,15 @@
       cursor: not-allowed;
     }
 
-    :host(:not([audio])) media-captions-menu-button::part(listbox) {
-      bottom: 42px;
```
gkatsev, line 62: bottom is no longer necessary

packages/mux-player/src/media-theme-mux.html
```diff
@@ -58,16 +58,15 @@
       cursor: not-allowed;
     }
 
-    :host(:not([audio])) media-captions-menu-button::part(listbox) {
```
gkatsev, line 61: I didn't actually try to see if I can add a part to the listbox, but since in this implementation it's in the lightdom, we don't need it. Likely to change again in the next selectmenu update.

cjpillsbury (approved):
Smoke tested cross browser + using css vars. 👌

PR #598, opened 2023-03-03
don't skip lib check in ts

gkatsev:


cjpillsbury:
Merged as part of #601

PR #600, opened 2023-03-03
chore: copy custom-types into dist during types build

gkatsev:


.github/workflows/ci.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       - name: Install playwright browsers
         run: npx playwright install --with-deps
       # run yarn and pass through `-- --all` to turborepo
-      - run: yarn test -- -- --all
+      - run: yarn test -- --concurrency=1 -- --all
```
gkatsev, line 66: this unblocks the PR.
Locally, for me just `yarn test` was mostly succeeded. Also, running individual projects succeeded without issues. Only running `yarn test -- -- --all` failed.
I think it's a reasonable change to unblock this PR. We'd still want to look at addressing the root cause and fixing it to be able to remove this concurrency limitation.
Test timing goes from ~35s to ~70s.

packages/mux-audio-react/package.json
```diff
@@ -33,7 +33,7 @@
     "dev": "npm-run-all --parallel dev:types dev:cjs dev:esm",
     "build:cjs": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/cjs.json --format=cjs --outdir=dist --out-extension:.js=.cjs.js --external:react --external:prop-types --external:@mux/* --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:esm": "esbuild src/index.tsx --target=es2019 --bundle --sourcemap --metafile=./dist/esm.json --format=esm --outdir=dist --out-extension:.js=.mjs --external:react --external:prop-types --external:@mux/* --define:PLAYER_VERSION=\"'$npm_package_version'\"",
-    "build:types": "tsc --declaration --emitDeclarationOnly --outDir './dist/types'",
+    "build:types": "tsc",
```
gkatsev, line 36: for all the projects, everything was moved into the tsconfig so that the types command is only `tsc`

packages/mux-audio-react/tsconfig.json
```diff
@@ -17,7 +18,7 @@
     "esModuleInterop": true,
     "skipLibCheck": false,
     "forceConsistentCasingInFileNames": true,
-    "typeRoots": ["node_modules/@types", "../../types"]
+    "typeRoots": ["node_modules/@types"]
```
gkatsev, line 21: we no longer need `../../types` in `typeRoots` and `include`

packages/mux-audio/package.json
```diff
@@ -28,8 +28,9 @@
   "author": "Mux, Inc.",
   "license": "MIT",
   "scripts": {
-    "clean": "shx rm -rf dist/",
+    "clean": "shx rm -rf dist/ && shx rm -rf src/polyfills",
```
gkatsev, line 31: clean should remove the local polyfills too.

packages/mux-audio/package.json
```diff
@@ -42,8 +43,11 @@
     "build:esm-module": "esbuild src/index.ts --target=es2019 --bundle --sourcemap --metafile=./dist/module.json --format=esm --outfile=./dist/mux-audio.mjs --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:cjs": "esbuild src/index.ts --target=es2019 --bundle --sourcemap --metafile=./dist/cjs.json --format=cjs --outdir=dist --out-extension:.js=.cjs.js --external:@mux/* --define:PLAYER_VERSION=\"'$npm_package_version'\"",
     "build:iife": "esbuild src/index.ts --target=es2019 --bundle --sourcemap --metafile=./dist/iife.json --format=iife --outfile=./dist/mux-audio.js --define:PLAYER_VERSION=\"'$npm_package_version'\"",
-    "build:types": "tsc --declaration --emitDeclarationOnly --outDir './dist/types'",
+    "copypolyfills": "shx mkdir -p src/polyfills && shx cp ../../shared/polyfills/index.ts ./src/polyfills/index.ts",
```
gkatsev, line 46: need to run this command on pretest, prebuild:types, and prebuild

packages/mux-audio/test/player.test.js
```diff
@@ -35,6 +34,9 @@ describe('<mux-audio>', () => {
       });
     });
 
+    player.playbackId = 'vDpm5ygrRJgfIEPNIc02IJR4Trf3z00AiP';
```
gkatsev, line 37: by setting the playback id after, it makes the test more reliable to be able to get a `loadstart` event.

packages/mux-player/src/media-chrome/dialog.ts
```diff
@@ -122,7 +122,8 @@ function focus(el: MediaDialog) {
 
   // Find element with `autofocus` attribute, or fall back to the first form/tabindex control.
   let target: Element | null | undefined = el.querySelector('[autofocus]:not([disabled])');
-  if (!target && (el as HTMLElement).tabIndex >= 0) {
+  if (!target && (el as unknown as HTMLElement).tabIndex >= 0) {
+    // @ts-ignore
```
gkatsev, line 126: wasn't sure exactly how to type this properly.

shared/polyfills/index.ts
```diff
@@ -0,0 +1,70 @@
+/* eslint @typescript-eslint/no-empty-function: "off", @typescript-eslint/no-unused-vars: "off" */
```
gkatsev, line 1: this file is what index.js used to be with additional typings.

types/mux-embed.d.ts
```diff
@@ -548,6 +548,16 @@ declare module 'mux-embed' {
...
 
+  type MuxType = (task: any) => void;
+  type muxProps = {
+    utils: publicUtils;
+    monitor: typeof monitor;
+  };
+  interface MuxGlobal extends MuxType, muxProps {}
+
+  export var mux: MuxGlobal;
+  export default mux;
```
gkatsev, line 559: some minimal typings to get a default export based on the types in mux-embed itself.

packages/mux-audio/tsconfig.json
```diff
@@ -2,11 +2,13 @@
...
     "rootDir": "src",
-    "outDir": "dist",
+    "outDir": "./dist/types",
+    "declaration": true,
+    "emitDeclarationOnly": true,
     "incremental": true,
     "target": "ES2019",
     "module": "es6",
-    "declaration": true,
+    "jsx": "react",
```
cjpillsbury, line 11: nitpick(non-blocking): Unneeded config prop.

packages/mux-player/tsconfig.json
```diff
@@ -2,21 +2,23 @@
...
     "rootDir": "src",
-    "outDir": "dist",
+    "outDir": "./dist/types",
+    "declaration": true,
+    "emitDeclarationOnly": true,
     "incremental": true,
     "target": "ES2019",
     "module": "es6",
-    "declaration": true,
+    "jsx": "react",
```
cjpillsbury, line 11: nitpick(non-blocking): unneeded.

packages/mux-player/tsconfig.json
```diff
@@ -2,21 +2,23 @@
...
     "sourceMap": true,
     "composite": true,
     "strict": true,
     "noImplicitAny": true,
-    "moduleResolution": "Node",
+    "moduleResolution": "node",
```
cjpillsbury, line 16: praise: 😎 nice catch

packages/mux-audio/tsconfig.json
gkatsev: I was thinking about it, the main reason it's here is to keep the tsconfigs across all the repos as close to each other as possible. However, I can def remove it if you think it's not worth having for the non-react packages.

types/mux-embed.d.ts
```diff
@@ -548,6 +548,16 @@ declare module 'mux-embed' {
...
   export const events: MuxEvents;
 
+  type MuxType = (task: any) => void;
+  type muxProps = {
+    utils: publicUtils;
+    monitor: typeof monitor;
+  };
+  interface MuxGlobal extends MuxType, muxProps {}
+
+  export var mux: MuxGlobal;
```
cjpillsbury, line 558: nitpick(non-blocking): `export const`

types/mux-embed.d.ts
```diff
@@ -548,6 +548,16 @@ declare module 'mux-embed' {
 
   export const events: MuxEvents;
 
+  type MuxType = (task: any) => void;
+  type muxProps = {
```
cjpillsbury, line 552: nitpick(non-blocking): We should be able to just add every named `export` defined in this ambient module to the `muxProps` type to avoid any future surprises.

.gitignore
```diff
@@ -127,4 +127,7 @@ dist
 **/*/iife.json
 **/*/module.json
 
+# ignore polyfills copy
+**/*/polyfills
```
cjpillsbury, line 131: Since this is copied to a predictable location per package, recommend:

```suggestion
packages/*/src/polyfills
```

packages/mux-audio/tsconfig.json
cjpillsbury: nah I'm cool with keeping it around, esp since tsconfig can inherit.

.gitignore
gkatsev: ah, yup, that makes sense. Otherwise, potential new files in shared/polyfills would also be ignored.

packages/mux-audio/tsconfig.json
gkatsev: actually, that's something to consider for the future. Have the top level tsconfig with like 90% of the stuff and have each package inherit from that one.

types/mux-embed.d.ts
gkatsev: Updated.

packages/mux-audio/package.json
gkatsev: We don't do this anymore. It's handled via turborepo.

cjpillsbury (approved):
😅 achievement unlocked.

luwes (approved):
LGTM 👍

PR #601, opened 2023-03-03
Fix/typedefs for lib checks

cjpillsbury:
includes @gkatsev's updates for `skipLibCheck`


types/mux-embed.d.ts
```diff
@@ -1,43 +1,577 @@
 declare module 'mux-embed' {
```
cjpillsbury, line 1: had to move defs into ambient module declaration.

packages/playback-core/src/types.ts
```diff
@@ -131,20 +133,21 @@ export type MuxMediaPropTypes = {
...
-  playerInitTime: Options['data']['player_init_time'];
+  playerInitTime: MetaData['player_init_time'];
   preferPlayback: ValueOf<PlaybackTypes> | undefined;
   type: MediaTypes;
   streamType: ValueOf<StreamTypes>;
   startTime: Hls['config']['startPosition'];
   autoPlay?: Autoplay;
   autoplay?: Autoplay;
   preferCmcd: ValueOf<CmcdTypes> | undefined;
+  error?: HTMLMediaElement['error'] | MediaError;
```
cjpillsbury, line 144: this one is technically a lie since we should _**always**_ yield our custom `MediaError` (or `null`) in `MuxVideoElement`. However, overriding accessor with completely different type signatures in subclasses gets messy, so went with an expanding type instead.

gkatsev (approved):
LGTM. Since it passes, I say merge unless @luwes had any thoughts on the changes.

luwes (approved):
👍

PR #607, opened 2023-03-13
fix: use default import for React

gkatsev:


packages/mux-uploader-react/src/useObjectPropEffect.ts
```diff
@@ -55,7 +55,7 @@ export const useObjectPropEffect = <T extends { [k: string]: any }, V>(
   updateValue = defaultUpdateValue,
   hasChanged = defaultHasChanged
 ) => {
-  return useEffect(() => {
+  return React.useEffect(() => {
```
cjpillsbury, line 58: nitpick(non-blocking): instead of changing usage of hooks, consider a module-level destructuring, e.g.:
```ts
import React from 'react';
const { useEffect } = React;
```

cjpillsbury (approved):
One (non-blocking) suggestion for an alternative refactor. 👌

gkatsev:
This change isn't necessary. We've decided that with CRA4 (and potentially simple webpack 4 configs) you'll need to do a deep import of the cjs file manually `import MuxUploader from "@mux/mux-uploader-react/dist/index.cjs.js";`

PR #608, opened 2023-03-14
fix: theme attributes

luwes:
this change makes it so we can forward attributes to the themes from the Mux player interface.

`theme-` prefixed makes sense I think because we don't really know which attributes the custom theme could support.


cjpillsbury:
@luwes should we move this to a draft PR or close the PR per our out of band convo?

luwes:
sure lets close, I really wish we could do this with CSS vars.

PR #613, opened 2023-03-17
fix: use Media Chrome display CSS vars

luwes:
Requires https://github.com/muxinc/media-chrome/pull/502


cjpillsbury (approved):
nothing to see here. All is well.

PR #616, opened 2023-03-20
fix: Cannot assume activeCues removed precisely in Chromium browsers …

cjpillsbury:
…(note: should file chromium bug).


gkatsev (approved):

luwes (approved):
LGTM

PR #618, opened 2023-03-21
Default theme enhancements

luwes:
- use poster slot
- remove unneeded default theme styles


packages/mux-player/src/template.ts
```diff
@@ -57,8 +57,6 @@ export const content = (props: MuxTemplateProps) => html`
     playback-rates="${props.playbackRates ?? false}"
     default-show-remaining-time="${props.defaultShowRemainingTime ?? false}"
     title="${props.title ?? false}"
-    poster="${props.poster === '' ? false : props.poster ?? false}"
-    placeholder="${props.placeholder ?? false}"
```
gkatsev, line 61: does mux-player-react/lazy use this? If so, it would need to be updated as well.

luwes: I think this is all under the hood for those pkgs because they interface with the `<mux-player>` WC

packages/mux-player/src/media-theme-mux.html
```diff
@@ -218,10 +214,10 @@
     media-loading-indicator {
       --media-loading-icon-width: 100%;
       --media-button-icon-height: auto;
+      display: var(--media-control-display, var(--media-loading-indicator-display, flex));
```
cjpillsbury, line 217: question(blocking): Is this order right? Wouldn't we want the more specific `--media-loading-indicator-display` to take precedence over the more generic `--media-control-display`?

packages/mux-player/src/media-theme-mux.html
```diff
@@ -263,6 +259,214 @@
       font-size: 21px;
     }
   </style>
+
+  <template partial="TitleDisplay">
```
cjpillsbury, line 263: praise: Pulling these up into partials 👌

packages/mux-player/src/media-theme-mux.html
```diff
@@ -528,7 +520,7 @@
           </media-control-bar>
         </template>
 
-        <template if="targetLiveWindow == 0">
+        <template if="!targetLiveWindow">
```
cjpillsbury, line 523: praise: !!🎉

AdamJaggard: Love this

packages/mux-player/src/template.ts
```diff
@@ -105,6 +103,12 @@ export const content = (props: MuxTemplateProps) => html`
         ? html`<track label="thumbnails" default kind="metadata" src="${props.storyboard}" />`
         : html``}
     </mux-video>
+    <media-poster-image
```
cjpillsbury, line 106: question(non-blocking): From the looks of this refactor, is the expectation that (non-`<mux-player>`) users of our themes will now have to slot a component to be able to see a poster image? If yes, that feels like a code sniff, but I don't want to hold up merging this, so we can treat it as a followup.

cjpillsbury (changes requested):
One blocking question, otherwise 👍

cjpillsbury (approved):
scratch that. Display css var precedence makes sense for wide -> narrow. Approved

PR #623, opened 2023-03-23
fix(mux-uploader): conditionally render mux-uploader-drop

davekiss:
If a user prefers to use `<mux-uploader-drop>` outside of the `<mux-uploader>` element, we should allow them to disable the default `<mux-uploader-drop>` element rendered as a child of `<mux-uploader>`

This adds a `noDrop` attribute which would modifiy the defauly layout to render a `div` element as a wrapper instead of `<mux-uploader-drop>`

### DIFF
--- a/packages/mux-uploader/src/layouts/block.ts
+++ b/packages/mux-uploader/src/layouts/block.ts
@@ -3,25 +3,26 @@ import { document } from '../polyfills';
 import '../mux-uploader-file-select';
 import { fileSelectFragment } from '../mux-uploader-file-select';
 
-const template = document.createElement('template');
+export default function blockLayout(noDrop: boolean) {
+  const wrapper = noDrop ? 'div' : 'mux-uploader-drop overlay';
 
-template.innerHTML = /*html*/ `
-<style>
-</style>
+  return document.createRange().createContextualFragment(`
+    <style>
+      /* Add your styles here */
+    </style>
 
-<mux-uploader-drop overlay>
-  <mux-uploader-status></mux-uploader-status>
-  <mux-uploader-retry></mux-uploader-retry>
+    <${wrapper}>
+      <mux-uploader-status></mux-uploader-status>
+      <mux-uploader-retry></mux-uploader-retry>
 
-  <mux-uploader-file-select>
-    <slot name="file-select">
-      ${fileSelectFragment}
-    </slot>
-  </mux-uploader-file-select>
+      <mux-uploader-file-select>
+        <slot name="file-select">
+          ${fileSelectFragment}
+        </slot>
+      </mux-uploader-file-select>
 
-  <mux-uploader-progress type="percentage"></mux-uploader-progress>
-  <mux-uploader-progress></mux-uploader-progress>
-</mux-uploader-drop>
-`;
-
-export default template;
+      <mux-uploader-progress type="percentage"></mux-uploader-progress>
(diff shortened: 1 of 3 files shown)


packages/mux-uploader/src/layouts/block.ts
```diff
@@ -3,25 +3,26 @@ import { document } from '../polyfills';
...
+      <mux-uploader-retry></mux-uploader-retry>
 
-  <mux-uploader-file-select>
-    <slot name="file-select">
-      ${fileSelectFragment}
-    </slot>
-  </mux-uploader-file-select>
+      <mux-uploader-file-select>
+        <slot name="file-select">
+          ${fileSelectFragment}
```
luwes, line 20: for now this is okay but I find it counter intuitive we need to repeat almost all HTML contents of mux-uploader-file-select here. I will help look at this later.

packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -135,6 +134,28 @@ class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderEl
     this._endpoint = value;
   }
 
+  get noDrop(): boolean {
+    return this.hasAttribute('noDrop');
```
luwes, line 138: lets keep the design guidelines to keep attributes lowercase.
it has no effect but still it's unusual

luwes, line 138:
```suggestion
    return this.hasAttribute('nodrop');
```

lets keep the design guidelines to keep attributes lowercase.
it has no effect but still it's unusual

packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -135,6 +134,28 @@ class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderEl
...
+  }
+
+  set noDrop(value: boolean) {
+    if (value) {
+      this.setAttribute('noDrop', '');
+    } else {
+      this.removeAttribute('noDrop');
+    }
```
luwes, line 146:
```suggestion
    this.toggleAttribute('nodrop', Boolean(value));
```

we can avoid if / else and use lowercase attribute naming

packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -135,6 +134,28 @@ class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderEl
...
+    this.updateLayout();
```
luwes, line 147: this should be called on in the `attributeChangedCallback` I think if we like attribute changes to have effect on the layout as well. not only setting the `noDrop` prop. requires `nodrop` to be added to the `static observedAttributes: []` as well

packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -135,6 +134,28 @@ class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderEl
...
+  }
+
+  updateLayout() {
+    const oldLayout = this.shadowRoot!.querySelector('mux-uploader-drop, div');
+    if (oldLayout) {
+      oldLayout.remove();
+    }
+    const newLayout = blockLayout(this.noDrop);
+    this.shadowRoot!.appendChild(newLayout);
+  }
```
luwes, line 157: 👍

luwes (approved):
👍 lgtm

PR #625, opened 2023-03-24
fix(mux-uploader): fixing dropzone styling regressions

davekiss:
Introducing the `<mux-uploader-drop>` element as a default presentation brought a few problems with how the elements were styled as the upload progressed. This PR fixes those regressions.

### DIFF
--- a/packages/mux-uploader/src/mux-uploader-progress.ts
+++ b/packages/mux-uploader/src/mux-uploader-progress.ts
@@ -9,14 +9,16 @@ const ariaDescription = 'Media upload progress bar';
 template.innerHTML = /*html*/ `
 <style>
   :host {
-    position: relative;
-    display: contents;
+    width: 100%;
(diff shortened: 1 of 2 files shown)


packages/mux-uploader/src/mux-uploader-progress.ts
```diff
@@ -6,10 +6,11 @@ import { formatProgress } from './utils/progress';
 const template = document.createElement('template');
 const ariaDescription = 'Media upload progress bar';
 
-template.innerHTML = `
+template.innerHTML = /*html*/ `
 <style>
   :host {
     position: relative;
+    display: contents;
```
luwes, line 13: I would try to avoid setting a custom element to `display: contents`.
it's surprising the host would not have dimensions (box)

luwes (approved):
LGTM

PR #628, opened 2023-03-28
fix: capLevel to player size & on fps drop

dylanjha:


luwes (approved):
LGTM
tested and fullscreen makes the player load the big boi segments

gkatsev (approved):

PR #631, opened 2023-03-30
fix: upgrade MC & integrate media-time-display

luwes:


packages/mux-player/src/index.ts
```diff
@@ -135,6 +136,16 @@ function getThemeTemplate(el: MuxPlayerElement) {
   }
 }
 
+function getHideDuration(el: MuxPlayerElement) {
```
cjpillsbury, line 139: thought(non-blocking): Given these acrobatics, I'm wondering if on the Media Chrome side we should break up the display for duration vs. current/remaining time into separate nodes with their own css var for display. That way we can avoid these acrobatics in player.

packages/mux-player/src/template.ts
```diff
@@ -56,6 +55,7 @@ export const content = (props: MuxTemplateProps) => html`
     backward-seek-offset="${props.backwardSeekOffset ?? false}"
     playback-rates="${props.playbackRates ?? false}"
     default-show-remaining-time="${props.defaultShowRemainingTime ?? false}"
+    hide-duration="${props.hideDuration ?? false}"
```
cjpillsbury, line 58: 👌

packages/mux-player/src/index.ts
luwes: yes, it's worth exploring again. I have done this in the past but for some reason dropped that effort

PR #640, opened 2023-04-03
Refactor blockLayout function to conditionally render elements

davekiss:
This PR introduces several improvements to the `blockLayout` to prevent unnecessary rendering of elements when they already exist in the DOM.

1. Introduced the `shouldRenderElement` function, which checks if an element with the given selector exists in the DOM. If it does not exist, the function returns `true`, indicating that the element should be rendered.

2. Introduced the `renderElement` function, which utilizes `shouldRenderElement` to conditionally render elements based on their presence in the DOM. This function simplifies the process of adding new elements without duplicating the logic for checking whether they should be rendered.

3. Refactored the `blockLayout` function to use the `renderElement` function for conditionally rendering the following elements:
   - `mux-uploader-status`
   - `mux-uploader-retry`
   - `mux-uploader-progress[type="percentage"]`
   - `mux-uploader-file-select`
   - `mux-uploader-progress:not([type])`

4. Introduced the `shouldRenderWrapper` function to conditionally render the `mux-uploader-drop overlay` element based on the `nodrop` value and its presence in the DOM. If `nodrop` is `true` or the element already exists in the DOM, a `div` will be used as a wrapper instead.

### DIFF
--- a/packages/mux-uploader/src/layouts/block.ts
+++ b/packages/mux-uploader/src/layouts/block.ts
@@ -3,26 +3,47 @@ import { document } from '../polyfills';
 import '../mux-uploader-file-select';
 import { fileSelectFragment } from '../mux-uploader-file-select';
 
+function shouldRenderElement(selector: string): boolean {
+  return !document.querySelector(selector);
+}
+
+function renderElement(selector: string, template: string): string {
+  return shouldRenderElement(selector) ? template : '';
+}
+
+function shouldRenderWrapper(nodrop: boolean, selector: string): string {
+  return !nodrop && !document.querySelector(selector) ? 'mux-uploader-drop overlay' : 'div';
+}
+
 export default function blockLayout(nodrop: boolean) {
-  const wrapper = nodrop ? 'div' : 'mux-uploader-drop overlay';
+  const wrapper = shouldRenderWrapper(nodrop, 'mux-uploader-drop');
 
-  return document.createRange().createContextualFragment(`
-    <style>
-      /* Add your styles here */
-    </style>
-
-    <${wrapper}>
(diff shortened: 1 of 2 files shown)


packages/mux-uploader/src/layouts/block.ts
```diff
@@ -3,26 +3,47 @@ import { document } from '../polyfills';
 import '../mux-uploader-file-select';
 import { fileSelectFragment } from '../mux-uploader-file-select';
 
-export default function blockLayout(nodrop: boolean) {
-  const wrapper = nodrop ? 'div' : 'mux-uploader-drop overlay';
+function shouldRenderElement(selector: string): boolean {
+  return !document.querySelector(selector);
```
gkatsev, line 7: If the elements are inside a Web Component, this won't find them and assume that the elements need to get added.
If this can get a reference to the mux-uploader element, we can at least check to see if the same document context is that contains it via `muxUploader.getRootNode()` instead of `document`.
Seems like potentially could pass `this` here https://github.com/muxinc/elements/blob/4e45cc07b03d29ed03fe09b94e28591f8ff9b886/packages/mux-uploader/src/mux-uploader.ts#L158

davekiss:
@gkatsev what do you think of the refactored approach?

luwes (changes requested):
I can't get on board with this approach I think. isn't this exactly what `slot`'s are for?
they can have a default inner html and this overridden by adding a slotted element in the light dom

davekiss:
As per discussion in eng sync, we've decided against this approach in favor of providing explicit attributes to prevent rendering specific components (ie. `noprogress`)

PR #646, opened 2023-04-05
Player `theme-*` attributes & `themeProps` property

luwes:
adds `theme-` prefixed attributes support for forwarding to the underlying `media-theme` element.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -66,6 +66,25 @@ const PlayerAttributes = {
   THEME: 'theme',
 };
 
+const ThemeAttributeNames = [
+  'audio',
+  'backward-seek-offset',
+  'default-show-remaining-time',
+  'default-showing-captions',
+  'disabled',
+  'exportparts',
+  'forward-seek-offset',
+  'hide-duration',
+  'hotkeys',
+  'nohotkeys',
+  'playbackrates',
+  'stream-type',
+  'style',
+  'target-live-window',
+  'template',
+  'title',
+];
+
 function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
   const props = {
     // Give priority to playbackId derrived asset URL's if playbackId is set.
@@ -327,6 +346,8 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
       if (!attributeName?.startsWith('theme-')) return;
 
       const themeAttrName = attributeName.replace(/^theme-/, '');
+      if (ThemeAttributeNames.includes(themeAttrName)) return;
+
       const value = this.getAttribute(attributeName);
       if (value != null) {
         this.mediaTheme?.setAttribute(themeAttrName, value);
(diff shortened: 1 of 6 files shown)


packages/mux-player/src/index.ts
```diff
@@ -290,22 +304,10 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
     return this.mediaTheme?.shadowRoot?.querySelector('media-controller');
   }
 
-  get metadataFromAttrs() {
```
gkatsev, line 293: Probably low risk, but technically a breaking change. I'm fine with making the change, but we'd want to revert if there's any complaints.

gkatsev (approved):
Works great in the example!

Is the react stuff coming in this PR? I can re-review at that point.

packages/mux-player/src/index.ts
luwes: yes, I don't think this was the intention to make public and we didn't document it. sounds good

luwes:
@gkatsev the React prop is added as well. example here
https://elements-demo-nextjs-git-fork-luwes-theme-attributes-mux.vercel.app/MuxPlayerTheme

packages/mux-player/src/index.ts
```diff
@@ -683,6 +681,41 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
...
+   * Set the theme attributes via a plain object.
+   */
+  set themeProps(props) {
+    this.#init();
+
+    for (const name in props) {
+      const value: string | boolean | null | undefined = props[name];
+
+      if (typeof value === 'boolean' || value == null) {
+        this.mediaTheme?.toggleAttribute(kebabCase(name), Boolean(value));
```
gkatsev, line 712: `toggleAttribute` can simplify so much of our attribute usage, especially with the forcing parameter. 👍🏻

packages/mux-player/src/index.ts
```diff
@@ -656,6 +681,41 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
     this.setAttribute(PlayerAttributes.THEME, `${val}`);
   }
 
+  /**
+   * Get the theme attributes in a plain object (camelCase keys).
+   * This includes already defined attributes. e.g. streamType, disabled, etc.
+   */
+  get themeProps() {
```
gkatsev, line 688: It's a bit confusing that if I set something via `themeProps`, if I get it again, I don't get the same object back.

gkatsev: Should we be filtering out some of the "well-known" attributes like `style` and `exportparts`? Because, I get set `theme-style` or `theme-exportparts` and it'll override the already set attribute, which seems dangerous.

packages/mux-player/src/index.ts
```diff
@@ -318,6 +339,33 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
...
+
+      const value = this.getAttribute(attributeName);
+      if (value != null) {
+        this.mediaTheme?.setAttribute(themeAttrName, value);
+      } else {
+        this.mediaTheme?.removeAttribute(themeAttrName);
+      }
+    };
+
+    const observer = new MutationObserver((mutationList) => {
```
gkatsev, line 359: should we store this observer in a private var so that we can disconnect it on `disconnectedCallback`?

gkatsev (approved):

packages/mux-player/src/index.ts
luwes: briefly discussed internally, opt to keep the attribute forwarding working even if disconnected for now.
if the element is released from memory, the observer should get cleaned up 
https://github.com/WICG/webcomponents/issues/550#issuecomment-460015529

PR #658, opened 2023-04-11
feat(mux-uploader): support max file size

davekiss:
This PR adds support for the `maxFileSize` feature provided by UpChunk. We introduce a new attribute called `max-file-size` for the `mux-uploader` custom element. This attribute allows users to set a maximum file size limit for the uploaded files. When the `max-file-size` attribute is set, its value will be passed down to UpChunk's `createUpload` method as the `maxFileSize` property.

### Changes

- Added a `maxFileSize` getter and setter to the `MuxUploaderElement` class
- Updated the `handleUpload` method to conditionally include the `maxFileSize` property in the options object passed to UpChunk's `createUpload` method

### Benefits

By supporting UpChunk's `maxFileSize` feature, developers can now easily enforce file size restrictions directly from the `mux-uploader` custom element, improving the user experience and reducing the need for additional client-side validation.

### DIFF
--- a/packages/mux-uploader/src/mux-uploader.ts
+++ b/packages/mux-uploader/src/mux-uploader.ts
@@ -13,7 +13,7 @@ rootTemplate.innerHTML = /*html*/ `
   }
 </style>
 
-<input id="hidden-file-input" type="file" />
+<input id="hidden-file-input" type="file" accept="video/*" />
 <mux-uploader-sr-text></mux-uploader-sr-text>
 `;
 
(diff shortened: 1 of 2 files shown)


luwes (approved):
LGTM

packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -190,6 +190,24 @@ class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderEl
     }
   }
 
+  get maxFileSize(): number | undefined {
+    const maxFileSize = this.getAttribute('max-file-size');
+    return maxFileSize !== null ? parseInt(maxFileSize) : undefined;
```
gkatsev, line 195: It's worth adding a radix to `parseInt`, otherwise, the parsed value could be unexpected. It could parse with a different radix depending on the string, and we probably don't want that.
```js
return maxFileSize !== null ? parseInt(maxFileSize, 10) : undefined;
```

Issue #506, opened 2022-11-30
Bug: MuxPlayer React component resets playback time to 0 after arbitrary setState happens 


dylanjha:
### Checklist

- [x] Verified whether your issue exists in the latest version Mux Element(s) release: `npm install @mux/[element_name]` or `yarn @mux/[element_name]`.
  - (NOTE: If the issue is not occurring in the latest version, we may not resolve it.)
- [x] Confirmed that this issue hasn't [already been filed](https://github.com/muxinc/elements/issues?q=is%3Aopen+is%3Aissue+label%3Abug).

### Which Mux Elements/Packages has a bug?

- mux-player-react

### Which browser(s) are you using?

- Chrome

### Which operating system(s) are you using?

macOS

### How are you using Mux Elements?

npm packate

### Describe the bug

Calling `setState` in a React component is causing the player to reset back to time `0`

## Steps To Reproduce

See codesandbox: https://codesandbox.io/s/debug-set-state-vz0j3p

1. start playing
1. cick "set some state"
1. player resets back to beginning of the video
1. seems to only happen on the first time clicking "set some state" subsequent clicks on "set some state" do not reset the player back to 0

## The current behavior

Player resets back to 0

## The expected behavior

Player should not reset back to 0, playback should continue uninterrupted

raytiley:
I think I found a work around by simply calling `useEffect` during initial render and setting a throw away state like `mounted`. This will force the player to `0` before the user has had a chance to interact with it, and seems to work ok for my use case.

https://codesandbox.io/s/debug-set-state-forked-oemyii?file=/App.js

dylanjha:
Well that feels a little silly, but a worthy work-around for the time being.

Thanks for sharing, we'll have to dig into this to find a permanent fix

PR #663, opened 2023-04-13
fix(mux-player-react): Have smarter value checking for currentTime to…

cjpillsbury:
… avoid unintentional 'resets' of currentTime to 0.


packages/mux-player-react/src/index.tsx
```diff
@@ -194,9 +194,26 @@ const usePlayer = (
...
+  // 1. Whether the prop value is/was "unset" to e.g. `undefined`
+  //    - To make sure we don't try to update currentTime on these changes
+  // 2. Whether the *prop* value was changed ("from the outside")
+  //    - To only apply the `currentTime` if it actually changed from the last provided value
+  // (CJP)
+  const [internalCurrentTime, setInternalCurrentTime] = useState(currentTime);
+  useEffect(() => {
+    if (internalCurrentTime == currentTime) return;
+    setInternalCurrentTime(currentTime);
+  }, [currentTime, internalCurrentTime]);
```
luwes, line 207:
```suggestion
  }, [currentTime, internalCurrentTime]);
  
```

a newline here would be nice. was difficult to read

packages/mux-player-react/src/index.tsx
```diff
@@ -194,9 +194,27 @@ const usePlayer = (
...
```
luwes, line 207: digging a bit more, I believe this code snippet is unneeded because React's effect compare array does the same thing https://github.com/muxinc/elements/blob/741d607521ca9578cfad9f0a9216a6565b4c56a1/packages/mux-player-react/src/useObjectPropEffect.ts#L63-L63

if the previous currentTime was not changed from the current, the effect will not run.

the check `if (currentTimeVal == null) return;` fixes the bug reported here
https://github.com/muxinc/elements/issues/506

does that seem right or am I missing something?

cjpillsbury: Ah so since we only care about the (immediately) previous state as a comparison for the current prop state and since react necessarily stores and compares that "under the hood" for `useEffect` (which we use in `useObjectPropEffect`), that "does all the work needed" for what I'm trying to solve via the `useState` + effect. So between my updated bail + no longer setting the `currentTime = 0` by default, yeah, I believe that should cover it! Brilliant!

luwes (approved):
👍

PR #665, opened 2023-04-15
Turborepo related chores

smorimoto:
Our team uses Mux Player, Next.js and Turborepo in production, and we realized that we were uploading a large amount of useless remote cache to Vercel because we didn't ignore this one 😱
I believe this reduces post-build cache upload time and build cache download time somewhat 🤷‍♂️


luwes (approved):
also LGTM! thanks for the heads up, that makes sense

PR #668, opened 2023-04-17
chore: Prepare default theme from Mux Player for export

AdamJaggard:
The goal of this PR is to package the existing default theme such that it can be exported as it's own module. This can then be easily swapped out in the future with a new theme.

Todo
- [x] Extract default theme files into it's own folder
- [x] Build the theme so that it can be imported by consumers of the mux-player package
- [x] Create all needed formats for each theme we expose (esm, commonjs, iife)
- [x] Figure out if we need IIFE builds (they're not currently exposed by package json)

Theme source files now reside in `src/themes` (themes based in this repo and re-exported from media chrome). Build artefacts for themes are in `dist/themes`.


packages/mux-player/src/template.ts
```diff
@@ -1,21 +1,14 @@
...
 import { getSrcFromPlaybackId, getStreamTypeFromAttr } from './helpers';
 import { html } from './html';
 import { i18n, stylePropsToString } from './utils';
-
+import classicTheme from './themes/classic';
 import type { MuxTemplateProps } from './types';
 import { StreamTypes } from '@mux/playback-core';
 
-const muxTemplate = document.createElement('template');
-if ('innerHTML' in muxTemplate) muxTemplate.innerHTML = muxTheme;
```
cjpillsbury, line 16: question(blocking): Why did you remove this code? I believe it will add slightly faster render time than the refactored version (cc @luwes). If so, I recommend reverting.

packages/mux-player/src/template.ts
```diff
@@ -37,7 +30,7 @@ const getHotKeys = (props: MuxTemplateProps) => {
 
 export const content = (props: MuxTemplateProps) => html`
   <media-theme
-    template="${props.themeTemplate ?? muxTemplate.content.children[0]}"
+    template="${props.themeTemplate ?? classicTheme.template}"
```
cjpillsbury, line 33: question(blocking): See above (related).

packages/mux-player/src/template.ts
AdamJaggard: This has been moved here [`src/themes/classic.ts`](https://github.com/muxinc/elements/pull/668/files#diff-5f2f1a5ac88395921da3dd66c63088cb4e4bcbe6552be1782212e3330f2c8e70R7). It seemed like it belongs to the theme as opposed to the consuming code, is that backwards?

packages/mux-player/src/themes/classic/index.ts
```diff
@@ -0,0 +1,18 @@
+// @ts-ignore
+import theme from './theme.html';
```
gkatsev, line 2: this probably should point at `./classic.html`?

packages/mux-player/src/template.ts
gkatsev: I think it's reasonable to keep in classic.ts, since this template element is available on the exported object.

luwes: this is fine yes, we want to have a template object that can be set via the MediaTheme#template prop

packages/mux-player/src/themes/classic/index.ts
AdamJaggard: Oops yup, got lost in renaming lots of stuff. But now I'm worried that my local running dev example page didn't show any issues 🤔...

packages/mux-player/src/template.ts
luwes: this could be removed ` ?? classicTheme.template` then if you make `theme` prop `classic` by default

packages/mux-player/src/themes/classic/index.ts
```diff
@@ -0,0 +1,18 @@
+// @ts-ignore
+import theme from './classic.html';
+import { document } from '../../polyfills';
+import { MediaThemeElement } from 'media-chrome/dist/media-theme-element';
+import { window } from 'media-chrome/dist/utils/server-safe-globals.js';
```
gkatsev, line 5: I believe this is causing all the typescript issues. We probably want to use the elements polyfill rather than importing media-chrome's server-safe-globals.

packages/mux-player/src/themes/classic/index.ts
```diff
@@ -0,0 +1,17 @@
+// @ts-ignore
+import theme from './classic.html';
+import { document } from '../../polyfills';
```
gkatsev, line 3: Should use the `globalThis` from the polyfills too.
```suggestion
import { document, globalThis } from '../../polyfills';
```

packages/mux-player/package.json
```diff
@@ -9,12 +9,29 @@
...
+      "require": "./dist/themes/minimal/index.cjs.js",
+      "browser": "./dist/themes/minimal/index.mjs",
+      "default": "./dist/themes/minimal/index.cjs.js"
+    },
+    "./themes/classic": {
+      "import": "./dist/themes/classic/index.mjs",
+      "require": "./dist/themes/classic/index.cjs.js",
+      "browser": "./dist/themes/classic/index.mjs",
+      "default": "./dist/themes/classic/index.cjs.js"
+    }
```
luwes, line 34: this looks good!

I hope we can deprecate `.mjs` soon. it's really a stone in my garden

we figured out a way to only use `.js` in Media Chrome

luwes (approved):
LGTM 👍

PR #670, opened 2023-04-19
fix: force theme to be ltr direction

gkatsev:
if the player is set to rtl direction, some functionality breaks, so, we want to force it to always be ltr

If we set it on the :host, then if someone sets direction on mux-player itself, it'll get overridden. So, instead, to force it to always be ltr, we set it on the media-theme element itself.

### DIFF
--- a/packages/mux-player/src/styles.css
+++ b/packages/mux-player/src/styles.css
@@ -38,6 +38,7 @@ a {
 media-theme {
   width: 100%;
   height: 100%;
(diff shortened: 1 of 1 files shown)


gkatsev:
To test, try adding `direction: rtl` to mux-player, or a parent element in the preview of this PR and compare to a preview from the main branch.

cjpillsbury (approved):
👌

PR #672, opened 2023-04-20
chore: add capture names for jq canary script

gkatsev:


cjpillsbury (approved):
If a tree falls in the yadda yadda...

PR #674, opened 2023-04-20
fix: use webkit pseudo element for captions movement, where available

### LINKED ISSUE #660: Bug: Captions jump to top on hover

mux-player-react

Chrome, Safari, Firefox

None

### Description

Sometimes when I hover over the player captions jump to the top for a second and then come back.
This is not the default behavior that happens on hover when closed captions get bottom padding.

This can also be seen on Mux Player Docs customisation page: https://docs.mux.com/guides/player/customize-look-and-feel

### Reduced test case

https://docs.mux.com/guides/player/customize-look-and-feel

### Steps to reproduce

1. Play the video with captions
2. Hover over and out a couple of times while captions are playing

### Current Behavior

Captions go all the way to the top as in the top position is set to 0.

### Expected Behavior

Captions get only the bottom padding.

### Errors

_No response_

v1.3.0

gkatsev:
While a single solution for all browsers is nice, we've had some reports of captions briefly showing up at the top of the video display area, which is unexpected. Therefore, we should switch back to the pseudo element technique for Chrome and Safari while keeping the programmatic line technique in use for Firefox, since it doesn't have anything else available.

This uses a transform on the text track container pseudo element and will transition it at 0.433s so that the transition is nice and smooth. This number was borrowed from the WebVTT specification. Also, when hiding, add a slight delay.

Fixes #660.

### DIFF
--- a/packages/mux-player/src/media-theme-mux.html
+++ b/packages/mux-player/src/media-theme-mux.html
@@ -73,7 +73,7 @@
     media-controller ::slotted([slot='media']) {
       --media-webkit-text-track-transition: transform 0.433s ease-out 0.3s;
     }
-    media-controller:is(:not([user-inactive])) ::slotted([slot='media']) {
+    media-controller:is([media-paused],:not([user-inactive])) ::slotted([slot='media']) {
       /* 42px is the height of the control bar and progress bar
        * with an additional 5px as a buffer, to get 47px */
       --media-webkit-text-track-transform: translateY(-47px);
--- a/packages/mux-player/src/media-theme-mux.html
+++ b/packages/mux-player/src/media-theme-mux.html
@@ -71,11 +71,13 @@
     /* 0.433s is the transition duration for VTT Regions.
      * Borrowed here, so the captions don't move too fast. */
     media-controller ::slotted([slot='media']) {
-      --media-webkit-text-track-transform: transform 0.433s 0.15s;
+      --media-webkit-text-track-transition: transform 0.433s ease-out 0.3s;
     }
-    media-controller:is([media-paused],:not([user-inactive])) ::slotted([slot='media']) {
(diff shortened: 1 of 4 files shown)


packages/mux-player/src/media-theme-mux.html
```diff
@@ -67,6 +67,19 @@
       --captions-selectmenu: none;
     }
 
+
+    /* 0.433s is the transition duration for VTT Regions.
+     * Borrowed here, so the captions don't move too fast. */
+    media-controller ::slotted([slot='media']) {
+      --media-webkit-text-track-transition: transform 0.433s ease-out 0.3s;
+    }
+    media-controller:is(:not([user-inactive])) ::slotted([slot='media']) {
```
luwes, line 76:
```suggestion
    media-controller:is(:not([user-inactive]),[media-paused]) ::slotted([slot='media']) {
```

packages/mux-player/src/media-theme-mux.html
```diff
@@ -67,6 +67,19 @@
       --captions-selectmenu: none;
     }
 
+
+    /* 0.433s is the transition duration for VTT Regions.
+     * Borrowed here, so the captions don't move too fast. */
+    media-controller ::slotted([slot='media']) {
+      --media-webkit-text-track-transition: transform 0.433s ease-out 0.3s;
```
luwes, line 74: it feels better but it's still touching the control bar fading out, the issue seems that the media-chrome control bar takes a full second to fade out. that's too long I think.

gkatsev: Yeah, we can probably increase the delay a tiny bit more, but probably not a lot more.

luwes (approved):
LGTM!

PR #678, opened 2023-04-28
chore(mux-uploader): Reflect state across multiple components

davekiss:
This allows for additional styles to be applied on more components based on the current upload state of the selected file.


cjpillsbury (commented):
Approving, though I think we should add/update tests to confirm these states are updated as expected on the components.

PR #681, opened 2023-05-03
chore: update mux-video-react example to use ref to toggle playback

gkatsev:


examples/nextjs-with-typescript/pages/MuxVideo.tsx
```diff
@@ -45,7 +45,7 @@ function MuxVideoPage() {
           <input
             id="paused-control"
             type="checkbox"
-            onChange={() => setPaused(!paused)}
+            onChange={() => paused ? mediaElRef.current.play() : mediaElRef.current.pause()}
```
cjpillsbury, line 48: suggestion(non-blocking): consider keeping this as is and instead doing the `play()`/`pause()` in a `useEffect(() => {}, [paused]);`

cjpillsbury (approved):
Approved with one suggestion for more predictable state + side effects.

PR #686, opened 2023-05-05
feat: add a no-volume-pref attribute to turn of storing volume in localStorage

gkatsev:


luwes (approved):
LGTM!

packages/mux-player/src/index.ts
```diff
@@ -1169,6 +1172,24 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
...
+
+  /**
+   * Set video engine debug flag.
+   */
+  set noVolumePref(val) {
+    if (val) {
+      this.setAttribute(PlayerAttributes.NO_VOLUME_PREF, '');
+    } else {
+      this.removeAttribute(PlayerAttributes.NO_VOLUME_PREF);
+    }
```
cjpillsbury, line 1190: nitpick(non-blocking): We're moving in the direction of preferring `toggleAttribute()` for these cases, e.g.

```suggestion
this.toggleAttribute(PlayerAttributes.NO_VOLUME_PREF, !!val);
```
Can also add a "changed check" to avoid unnecessary re-rendering, e.g.:
```ts
const boolVal = !!val;
if (boolVal === this.noVolumePref) return;
this.toggleAttribute(PlayerAttributes.NO_VOLUME_PREF, boolVal);
```

cjpillsbury (approved):
Approved with one nb suggestion!

packages/mux-player/src/index.ts
gkatsev: Feels a bit weird to be the odd one out.

PR #691, opened 2023-05-16
Feat/mc v1 upgrade

cjpillsbury:
Good to go. If we need to do a hot fix release before mc@1.0 is out of rc, we're comfortable releasing against the latest rc.

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -69,24 +69,24 @@ const PlayerAttributes = {
 
 const ThemeAttributeNames = [
   'audio',
-  'backward-seek-offset',
-  'default-show-remaining-time',
-  'default-showing-captions',
+  'backwardseekoffset',
+  'defaultshowremainingtime',
+  'defaultshowingcaptions',
   'noautoseektolive',
   'disabled',
   'exportparts',
-  'forward-seek-offset',
-  'hide-duration',
+  'forwardseekoffset',
+  'hideduration',
   'hotkeys',
   'nohotkeys',
   'playbackrates',
-  'default-stream-type',
-  'stream-type',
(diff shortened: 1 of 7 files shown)


packages/mux-player/src/template.ts
```diff
@@ -16,12 +16,14 @@ const muxTemplate = document.createElement('template');
 if ('innerHTML' in muxTemplate) muxTemplate.innerHTML = muxTheme;
 
 // prettier-ignore
-export const template = (props: MuxTemplateProps) => html`
+export const template = (props: MuxTemplateProps) => {
+  console.log('props!', props);
```
luwes, line 20: don't forget to remove

packages/mux-player/src/media-theme-mux.html
```diff
@@ -53,8 +53,8 @@
       width: 100%;
     }
 
-    [disabled],
-    [aria-disabled='true'] {
+    [disabled]:not(media-live-button),
+    [aria-disabled='true']:not(media-live-button) {
```
cjpillsbury, line 57: NOTE: Unrelated fix for styling of live button (which is disabled when at the live edge)

packages/mux-player/src/index.ts
```diff
@@ -69,24 +69,24 @@ const PlayerAttributes = {
 
 const ThemeAttributeNames = [
   'audio',
-  'backward-seek-offset',
-  'default-show-remaining-time',
-  'default-showing-captions',
+  'backwardseekoffset',
+  'defaultshowremainingtime',
+  'defaultshowingcaptions',
```
luwes, line 74:
```suggestion
  'defaultshowingsubtitles',
```

PR #697, opened 2023-05-24
fix: use custom-media-element pkg

luwes:
https://github.com/muxinc/custom-media-element


cjpillsbury (approved):
Pre-approving on the caveat that we do testing using our web build tests on the canary release shortly after the merge. ❤️ This is great.

PR #702, opened 2023-06-01
Fix/stream info for mp4

cjpillsbury:
Also adds some tests for better regression testing and footprint.
Exports the `muxMediaState` map for better test side effect cleanup (NOTE: Should probably apply this more generally to test beforeEach/afterEach as a followup).


luwes (approved):
LGTM

PR #705, opened 2023-06-06
fix(playback-core, mux-video): Handle native playback edge cases wher…

cjpillsbury:
…e ended state doesn't work as expected based on small timeline discrepancies.

### DIFF
--- a/packages/mux-video/src/index.ts
+++ b/packages/mux-video/src/index.ts
@@ -21,6 +21,8 @@ import {
   getTargetLiveWindow,
   getLiveEdgeStart,
   getSeekable,
+  getEnded,
+  getPlayPromise,
 } from '@mux/playback-core';
 import type { PlaybackCore, PlaybackEngine, Autoplay, ExtensionMimeTypeMap, ValueOf } from '@mux/playback-core';
 import { getPlayerVersion } from './env';
@@ -328,6 +330,12 @@ class MuxVideoElement extends CustomVideoElement implements Partial<MuxMediaProp
     }
   }
 
+  get ended() {
+    // This ensures that edge case media that doesn't properly end will
+    // still announce itself as "ended".
+    return getEnded(this.nativeEl, this._hls);
+  }
+
   get envKey(): string | undefined {
     return this.getAttribute(Attributes.ENV_KEY) ?? undefined;
   }
@@ -490,6 +498,12 @@ class MuxVideoElement extends CustomVideoElement implements Partial<MuxMediaProp
     await (this.#loadRequested = Promise.resolve());
     this.#loadRequested = null;
     this.load();
+  }
+
+  play() {
+    // This accounts for "replay" behavior for edge case media that doesn't properly enter
+    // the "ended" state. See also: get ended(), above.
+    return getPlayPromise(this.nativeEl, this._hls);
   }
 
   load() {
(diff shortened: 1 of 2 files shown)


packages/playback-core/src/index.ts
```diff
@@ -281,6 +281,43 @@ export const getLiveEdgeStart = (mediaEl: HTMLMediaElement) => {
...
+};
+
+export const getPlayPromise = (mediaEl: HTMLMediaElement, hls?: HlsInterface) => {
+  if (hls || mediaEl.ended || !getEnded(mediaEl, hls)) return mediaEl.play();
+  const mediaState = muxMediaState.get(mediaEl) ?? {};
+  if (!mediaState?.pendingPlayPromise) {
+    mediaState.pendingPlayPromise = createPlayPromise(mediaEl);
+  }
+  return mediaState.pendingPlayPromise;
+};
```
luwes, line 319: I think this play promise logic is not absolutely needed. can it be removed or why is it a hard requirement?

packages/mux-video/src/index.ts
```diff
@@ -492,6 +500,12 @@ class MuxVideoElement extends CustomVideoElement implements Partial<MuxMediaProp
     this.load();
   }
 
+  play() {
+    // This accounts for "replay" behavior for edge case media that doesn't properly enter
+    // the "ended" state. See also: get ended(), above.
+    return getPlayPromise(this.nativeEl, this._hls);
+  }
```
luwes, line 507: consider removing, use `play` event callback

packages/mux-video/src/index.ts
```diff
@@ -328,6 +330,12 @@ class MuxVideoElement extends CustomVideoElement implements Partial<MuxMediaProp
     }
   }
 
+  get ended() {
+    // This ensures that edge case media that doesn't properly end will
+    // still announce itself as "ended".
+    return getEnded(this.nativeEl, this._hls);
+  }
```
luwes, line 337: 👍 I reproduced the `ended` property being false right after the ended event callback so this makes sense.

this would also be easily patched in Media Chrome `media.ended || time > duration - 0.1` or something since we can't shim all the media elements that would be used in Media Chrome.

packages/playback-core/src/index.ts
```diff
@@ -281,6 +281,43 @@ export const getLiveEdgeStart = (mediaEl: HTMLMediaElement) => {
   return seekable.end(seekable.length - 1) - liveEdgeStartOffset;
 };
 
+const isApproximatelyEqual = (x: number, y: number, moe = 0.001) => Math.abs(x - y) <= moe;
+const isApproximatelyGTE = (x: number, y: number, moe = 0.001) => x > y || isApproximatelyEqual(x, y, moe);
```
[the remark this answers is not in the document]
luwes: greater than equals, I should have got that the first time haha

luwes (approved):
👍 
I traced and repro'd the fake restart on my end

packages/playback-core/src/index.ts
```diff
@@ -554,6 +566,20 @@ export const loadMedia = (
...
+      mediaEl.dispatchEvent(new Event('ended'));
+    });
+
+    addEventListenerWithTeardown(mediaEl, 'play', () => {
+      if (mediaEl.ended) return;
+      if (!isApproximatelyGTE(mediaEl.currentTime, mediaEl.duration)) return;
+      // If we were "pseudo-ended" before playback was attempted, seek back to the
+      // beginning to "replay", like "real" ended behavior.
+      mediaEl.currentTime = mediaEl.seekable.start(0);
+    });
```
luwes, line 582: this could also be easily added to the React versions but yeah I would be fully on board to make those wrappers around mux-video and mux-audio

PR #707, opened 2023-06-14
fix(playback-core): Apply pseudo-ended logic on seeked, even if paused.

cjpillsbury:
NOTE: This is actually a bug in browsers based on a previously ambiguous definition in the HTML spec. For context, see:
- https://github.com/video-dev/hls.js/issues/5168
- https://html.spec.whatwg.org/multipage/media.html#reaches-the-end (referenced in the issue, above)

Additionally, I've updated the "pseudo-ended" logic to not apply when the media element is looping, since (again, per spec), that media should not ever "end".

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -289,7 +289,8 @@ export const isPseudoEnded = (mediaEl: HTMLMediaElement) => {
 };
 
 export const getEnded = (mediaEl: HTMLMediaElement, hls?: HlsInterface) => {
-  if (!!hls) return mediaEl.ended;
+  // Since looping media never truly ends, don't apply pseudo-ended logic
+  if (mediaEl.loop || !!hls) return mediaEl.ended;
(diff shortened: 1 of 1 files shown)


packages/playback-core/src/index.ts
```diff
@@ -566,12 +567,19 @@ export const loadMedia = (
       },
       { once: true }
     );
-    addEventListenerWithTeardown(mediaEl, 'pause', () => {
-      if (!getEnded(mediaEl)) return;
+    const maybeDispatchEndedCallback = () => {
+      // Do the "cheaper" check first
       if (mediaEl.ended) return;
```
AdamJaggard, line 572: non-blocking: is this line needed? it looks like `getEnded()` does some very cheap checks and `mediaEl.ended` is evaluated before `isPseudoEnded()` in that function also. Just wondering if the redundancy is needed. Either way it's fine 🙂

AdamJaggard (approved):
LGTM

PR #708, opened 2023-06-15
feat: implement renditions API

luwes:
Media Tracks + renditions API at https://github.com/muxinc/media-tracks

test url: https://elements-demo-vanilla-git-fork-luwes-renditions-mux.vercel.app/mux-player-renditions.html

### DIFF
--- a/packages/mux-video/package.json
+++ b/packages/mux-video/package.json
@@ -62,7 +62,8 @@
   "dependencies": {
     "@mux/playback-core": "0.18.5",
     "castable-video": "0.5.2",
-    "custom-media-element": "0.2.0"
+    "custom-media-element": "0.2.0",
+    "media-tracks": "0.1.2"
   },
   "devDependencies": {
     "@mux/test-esm-exports": "0.1.0",
--- a/packages/mux-video/package.json
+++ b/packages/mux-video/package.json
@@ -63,7 +63,7 @@
     "@mux/playback-core": "0.18.5",
     "castable-video": "0.5.2",
     "custom-media-element": "0.2.0",
-    "media-tracks": ">=0.1.2"
+    "media-tracks": ">=0.2.1"
   },
   "devDependencies": {
     "@mux/test-esm-exports": "0.1.0",
--- a/packages/mux-video/package.json
+++ b/packages/mux-video/package.json
@@ -63,7 +63,7 @@
     "@mux/playback-core": "0.18.5",
     "castable-video": "0.5.2",
     "custom-media-element": "0.2.0",
-    "media-tracks": "0.1.2"
+    "media-tracks": ">=0.1.2"
   },
   "devDependencies": {
     "@mux/test-esm-exports": "0.1.0",
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -166,7 +166,7 @@ export const getStreamInfoFromHlsjsLevelDetails = (levelDetails: any) => {
 export const updateStreamInfoFromHlsjsLevelDetails = (
   levelDetails: any,
   mediaEl: HTMLMediaElement,
-  hls: HlsInterface
+  hls: Pick<Hls, 'config' | 'userConfig' | 'liveSyncPosition'>
 ) => {
   const { streamType, targetLiveWindow, liveEdgeStartOffset, lowLatency } =
     getStreamInfoFromHlsjsLevelDetails(levelDetails);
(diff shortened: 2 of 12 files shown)


packages/playback-core/src/renditions.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const videoTrack = customMediaEl.addVideoTrack('main');
+    videoTrack.selected = true;
+
+    for (const [id, level] of data.levels.entries()) {
+      const videoRendition = videoTrack.addRendition(
+        level.url[0],
+        level.width,
+        level.height,
+        level.videoCodec,
```
cjpillsbury, line 25: this should be the composed mimecodec str, correct?

packages/playback-core/src/renditions.ts
```diff
@@ -0,0 +1,103 @@
...
+        level.bitrate
+      );
+
+      // The returned levels all have an id of `0`, save the id in a WeakMap.
+      levelIdMap.set(level, `${id}`);
+
+      videoRendition.id = `${id}`;
+      videoRendition.enabled = hls.autoLevelEnabled;
```
cjpillsbury, line 33: I believe this isn't quite right. enabled wrt the media-ui-extensions means it is selectable. autoLevelEnabled means the HLS ABR level switching is enabled.

packages/playback-core/src/index.ts
```diff
@@ -514,26 +516,7 @@ export const setupMux = (
...
-    | 'destroy'
-    | 'loadSource'
-    | 'attachMedia'
-    | 'liveSyncPosition'
-    | 'subtitleTracks'
-    | 'subtitleTrack'
-    | 'userConfig'
-  >
-) => {
+export const loadMedia = (props: Partial<MuxMediaProps>, mediaEl: HTMLMediaElement, hls?: HlsInterface) => {
```
cjpillsbury, line 519: I'm cool with simplifying these type defs, though the original intent was to make sure we were being intentional about our API dependencies.

packages/playback-core/src/renditions.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  hls.on(Hls.Events.LEVEL_SWITCHED, function (event, data) {
+    const videoTrack = customMediaEl.videoTracks[0];
+    if (!videoTrack) return;
+
+    // data.level returns an index of the level, not an id.
+    const levelId = levelIdMap.get(hls.levels[data.level]);
+    const rendition = videoTrack?.renditions.getRenditionById(levelId);
+    if (rendition) {
+      rendition.active = true;
```
cjpillsbury, line 64: question(blocking): Shouldn't we also be setting any previously active rendition to false here?

packages/playback-core/src/renditions.ts
luwes: discussed internally, in this iteration `enabled` is used as a way to select 1, lock to 2, 3, or multiple renditions.

this will be discussed further in the future if there is a need for an API to select just 1 or a need for a flag to indicate the rendition is hidden / not possible to use.

packages/playback-core/src/renditions.ts
luwes: that's done internally 
https://github.com/muxinc/media-tracks/blob/2cf67e680301a23100a17c669ad50ac8989f6771/src/video-rendition.ts#L56-L58

packages/playback-core/src/renditions.ts
luwes: I'd rather do this in a next iteration. it's not required for the renditions UI selector afaik

cjpillsbury (approved):
mentioned out of band but not a blocker: there are some inefficiencies around buffer flushes and segment refetches wrt the current rendition selection impl (mostly due to hls.js's default behaviors/API that we're relying on) that we should try to resolve, but those should not be a blocker for this initial merge. Approved/great spike on this, @luwes! 🎉

packages/mux-video/package.json
```diff
@@ -62,7 +62,8 @@
   "dependencies": {
     "@mux/playback-core": "0.18.5",
     "castable-video": "0.5.2",
-    "custom-media-element": "0.2.0"
+    "custom-media-element": "0.2.0",
+    "media-tracks": ">=0.2.1"
```
cjpillsbury, line 66: praise: Good call/attention to detail on the semver qualifier here (re: playback-core version consistency) 🎉

cjpillsbury (approved):
gnarly 🤙 😎

PR #709, opened 2023-06-16
feat: New default theme

AdamJaggard:
This PR retains the original (`classic`) theme as the default theme for the time being. We can merge this and test further without it being 100% ready. The new theme is named `2023` and is exported alongside all other themes.

Unfinished bits:
- [ ] Centered mute button when autoplay is enabled. This element exists commented out in the theme template but we need to figure out A: how to make it trigger unmute behaviour, and B: how to combine the right set of template variables to show it at the right time.
- [X] Select menu tweaks to more closely match design (tick icon)
- [ ] Integrate rendition selector when ready
- [ ] General testing

image


luwes:
Love the animations! 

some visual feedback:

- it seems the play icon is not volumetric centered 
https://stackoverflow.com/questions/44071333/formula-for-calculating-the-volumetric-center-or-optical-alignment-centroid-of
SCR-20230802-kgi
- the control bar icons seem too small and cramped, especially on the right. 
  I would add 1 or 2px height if possible and some space between them.

SCR-20230802-kex

AdamJaggard:
@luwes agreed regarding control sizes, it was the first thing I noticed. Because we are planning to merge this with the old default theme remaining the default for the time being, we can get Jason to have a play with it and get his feedback on changing the size.

luwes (approved):
LGTM for merging but not enabling as default
requires more testing in live, DVR, etc mode

PR #712, opened 2023-06-26
feat: add Mux uploader rsc for Next 13

luwes:


packages/mux-uploader-react/src/rsc.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const { Video } = new Mux();
+
+export default async function MuxUploaderServer(props: any) {
+  const upload = await Video.Uploads.create({
+    cors_origin: '*',
+    new_asset_settings: {
+      playback_policy: 'public',
+    },
+  });
```
dylanjha, line 13: We'll an ability to pass in props to the `Uploads.Create` function, right?

PR #718, opened 2023-07-07
fix(playback-core): Switch back to using the unminified hls.js module…

cjpillsbury:
…, since this appears to cause edge case issues with initial times after ts segment transmuxing.

To test Next.js usage:
- confirm playbackId `8xhTU2GpKuskv1eCx8rdp3PodlGb6dABCf27f4BsmW8` does not continually load the 0th segment before playback
- confirm that playback begins at time 0
- confirm that you can seek back to time 0 after playback has begun

**NOTE**: We originally used the minified version of hls.js due to issues with svelte-kit and Vue minified builds, with a root cause of how its web worker was integrated. This appears to be resolved since hls.js has migrated its build setup to rollup. Confirmed locally that there were no issues with svelte-kit, but make sure to do additional smoke testing for our Vue and Svelte apps before any approval. For reference, see:

- https://github.com/muxinc/elements/issues/541
- https://github.com/video-dev/hls.js/issues/5107

### DIFF
--- a/packages/playback-core/src/hls.ts
+++ b/packages/playback-core/src/hls.ts
@@ -1,9 +1,10 @@
-// @ts-ignore
-import HlsMin from 'hls.js/dist/hls.min.js';
+import Hls from 'hls.js';
+// // @ts-ignore
+// import HlsMin from 'hls.js/dist/hls.min.js';
 import type HlsClassType from 'hls.js';
 
-type Hls = typeof HlsClassType & HlsClassType;
-const Hls: Hls = HlsMin as unknown as Hls;
+// type Hls = typeof HlsClassType & HlsClassType;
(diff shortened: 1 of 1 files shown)


packages/playback-core/src/hls.ts
```diff
@@ -1,9 +1,10 @@
-// @ts-ignore
-import HlsMin from 'hls.js/dist/hls.min.js';
+import Hls from 'hls.js';
```
cjpillsbury, line 1: NOTE: Keeping the changes narrow for now for both expediency and just in case we need to revert anything. We can minimally remove the commented out code pre-merge if preferred.

dylanjha:
Confirmed the fix! LGTM

Do we understand _why_ the minified version was causing this?

dylanjha (approved):

PR #724, opened 2023-07-25
Add missing transitive peer dependencies

smorimoto:


smorimoto:
@luwes This fixes a dependency issue that was missed due to some rude npm flaws. There is no direct impact to the user, just to keep the codebase right.

luwes (approved):

smorimoto:
@luwes Unrelated to this PR, my goal this week is to share a PR about the build infrastructure around TypeScript! It should probably fix all the problems from last time nicely!

PR #725, opened 2023-07-25
Fix type field path

smorimoto:
If I understand correctly, I think the previous PR had the wrong type field... I completely forgot to fix that. I believe this will eliminate any errors.

### DIFF
--- a/packages/mux-active-viewer-count/package.json
+++ b/packages/mux-active-viewer-count/package.json
@@ -27,7 +27,7 @@
     "types@<4.3.5": "./dist/types-ts3.4/index.d.ts",
     "types": "./dist/types/index.d.ts"
   },
-  "types": "./dist/types-ts3.4/index.d.ts",
+  "types": "./dist/types/index.d.ts",
   "repository": {
     "type": "git",
     "url": "https://github.com/muxinc/elements",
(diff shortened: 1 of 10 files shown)


luwes (approved):
👍

luwes:
thanks @smorimoto ! it seems better 
https://arethetypeswrong.github.io/?p=%40mux%2Fmux-player%401.11.4

but not all is good yet somehow
https://arethetypeswrong.github.io/?p=%40mux%2Fmux-player%401.11.4-canary.0-2197108

our web build checks passed. was a bit worried about the change to `type: module`
https://github.com/muxinc/web-build-checks/actions/runs/5659725527/job/15333912154

smorimoto:
@luwes As for the web build checks, it seems like a private repository and I don't have access to it.

luwes:
I think we might have to update moduleResolution. I'm doing some research

https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/InternalResolutionError.md#declarations-and-javascript-match-but-are-not-checked-with-correct-settings

luwes:
it looks like it's trying to resolve `/helpers` in ESM mode but that requires `/helpers.js` to access the file in ESM.

SCR-20230725-imb

it's actually good practice to add the `.js` extension always, even in TS so ESM works when not bundling.

smorimoto:
@luwes Is explicit support for older TypeScript mandatory in Mux?

luwes:
I think version 4 is probably sufficient. @cjpillsbury do you know how far back we support?

PR #726, opened 2023-07-25
Transition from destructured named imports to namespace imports

smorimoto:
This PR does the groundwork to enable the new jsx transform in the future.

I recommend this blog post because the details are very clear: https://epicreact.dev/importing-react-through-the-ages

After merging this, the Mux team should be able to enable it by changing tsconfig to something like:

```diff
- "jsx": "react"
+ "jsx": "react-jsx"
```

esbuild:

```
--jsx
Set to "automatic" to use React's automatic runtime or to "preserve" to disable transforming JSX to JS
```


cjpillsbury:
Hey @smorimoto thanks for the eager contributions! We're going to hold off on merging this one for now, since we don't currently plan on supporting alternative jsx/"h" transforms such as `preact`. If we ever do prioritize this in the future, we'd still want to keep the code changes as narrow as possible, so, e.g.:

- only make these changes to our library `/packages` (and not, e.g. our `/examples` projects in the monorepo)
- don't change unrelated things such as formatting
- destructure the `React` module "namespace" at the top of the corresponding module to keep changes narrow, so e.g.

not:
```tsx
import * as React from 'react';

// ...
React.useEffect(/* ... */);
```

instead:
```tsx
import * as React from 'react';
const {
  useEffect
  // ... other dependencies that were imported by name
} = React;

// ...
useEffect(/* ... */);
```

Still appreciate all of the engagement! Always feel free to start a [discussion](https://github.com/muxinc/elements/discussions/new/choose) or open an [issue](https://github.com/muxinc/elements/issues/new/choose) if you want to avoid doing a bunch of work that may not end up getting merged.

PR #730, opened 2023-08-02
chore: bump media chrome to 1.2.1 for player

AdamJaggard:
Bumps the player package media-chrome dependency to 1.2.1.

This would allow using recent media chrome features (like list box enhancements) in the new default player theme.


luwes (approved):
LGTM!

PR #733, opened 2023-08-04
fix(mux-player): exporting underlying poster img part for advanced st…

### LINKED ISSUE #717: Feature Request: Add `exportparts` to the poster `<img>` element so we can style it

mux-player-react

### Description

This exists for the `<video>` element so I'm not certain why there isn't another one for the `` element. I've spent nearly a day trying to add a border-radius to both elements but the docs don't cover some of these examples and were slightly misleading. The description seemed to imply that you could build something like the below example, in order to style a child element:

```css
  mux-player::part(poster-layer img) {
    border-radius: 1.5rem;
  }
```

I'm testing this out in work to see if we'd like to use purchase your video service so I'd appreciate a quick release with this if possible (I have some design guidelines I need to meet to demo). I noticed https://github.com/muxinc/elements/pull/384 which might be of some help.

Might also be worth an update to the docs to include an example such as mine for future reference.

### Expected Behavior

This or similar should work:
```css
  mux-player::part(video),
  mux-player::part(image) {
    border-radius: 1.5rem;
  }
```

cjpillsbury:
…yling.

depends on https://github.com/muxinc/media-chrome/pull/712
fixes #717

### DIFF
--- a/packages/mux-player/package.json
+++ b/packages/mux-player/package.json
@@ -83,7 +83,7 @@
   "dependencies": {
     "@mux/mux-video": "0.15.6",
     "@mux/playback-core": "0.18.5",
-    "media-chrome": "^1.2.1"
+    "media-chrome": "^1.2.2"
   },
   "devDependencies": {
     "@mux/esbuilder": "0.1.0",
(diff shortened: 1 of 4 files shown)


AdamJaggard (approved):
LGTM

PR #734, opened 2023-08-08
feat: add quality selector

### LINKED ISSUE #572: Docs: Quality Controls
# Which documentation could use some work?

- mux-video
- general mux-elements documentation
- hls

## 🖊 Describe what you'd like to see a change in the documentation

HLS is great, but it feels a little odd that there is no way to natively select video quality. For example, if the content of the video is quality-critical, IMHO the user should be able to select the quality explicitly.

Currently, I have not been able to find a way how to enable the quality controls neither in `mux-player-react` nor make `plyr` work with Mux source :( This is a major pain point, and I hope that I'm simply missing something. Am I?

luwes:
https://elements-demo-vanilla-git-fork-luwes-quality-selector-mux.vercel.app/mux-player.html

https://elements-demo-nextjs-git-fork-luwes-quality-selector-mux.vercel.app/MuxPlayer?title=%22Landscape+VOD+%28White+Background%29%22&playbackId=%22VcmKA6aqzIzlg3MayLJDnbF55kX00mds028Z65QxvBYaA%22&metadata=%7B%22video_id%22%3A%22videoId1%22%2C%22video_title%22%3A%22Title%3A+Landscape+VOD+%28White+Background%29%22%7D

https://elements-demo-nextjs-git-fork-luwes-quality-selector-mux.vercel.app/MuxPlayer?title=%22Long+Lived+LL-HLS%22&playbackId=%22v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM%22&metadata=%7B%22video_id%22%3A%22videoId11%22%2C%22video_title%22%3A%22Title%3A+Long+Lived+LL-HLS%22%7D

resolves #572


examples/vanilla-ts-esm/public/mux-player-renditions.html
```diff
@@ -46,7 +46,7 @@ <h1><a href="/">Elements</a></h1>
     <mux-player
       id="muxPlayer"
       stream-type="on-demand"
-      playback-id="23s11nz72DsoN657h4314PjKKjsF2JG33eBQQt6B95I"
+      playback-id="VcmKA6aqzIzlg3MayLJDnbF55kX00mds028Z65QxvBYaA"
```
cjpillsbury, line 49: question(non-blocking): Is this a better playback-id to demo renditions or was this just for some testing? if the latter, consider reverting.

packages/mux-player/src/index.ts
```diff
@@ -1,6 +1,6 @@
 import { globalThis, document } from './polyfills';
 import { MediaController } from 'media-chrome';
-import 'media-chrome/dist/experimental/media-captions-selectmenu.js';
+import 'media-chrome/dist/experimental/index.js';
```
cjpillsbury, line 3: nitpick(blocking): We should directly import each experimental feature we use so we aren't unnecessarily adding to our bundles and are being explicit about the dependencies we are using.

packages/mux-player/src/themes/classic/classic.html
```diff
@@ -80,14 +79,6 @@
       --media-webkit-text-track-transition: transform 0.15s ease;
     }
 
-    media-captions-selectmenu {
```
cjpillsbury, line 83: praise: love this additional cleanup.

cjpillsbury (changes requested):
one nitpick, Otherwise, 💪

examples/vanilla-ts-esm/public/mux-player-renditions.html
luwes: I prefer this playback id for inspecting the quality selector UI. it doesn't have a black background.

packages/mux-player/src/index.ts
luwes: I had it first like that but then I thought what about custom themes.
it's nice to have all the experimental elements to drop in without needing imports.
the 1 left is playback-rate-selectmenu and dependencies, comparative to the Mux player bundle it's a drop in a bucket. thoughts?

cjpillsbury: hmmm... I'm sold as solution for now. Maybe next week in media chrome office hours we can try to come up with a solution to make sure themes can include any of their dependencies? This has come up in the past and would hopefully avoid your concern here.

cjpillsbury (approved):
Sold on the current solution per @luwes's immediate concerns, re: theme dependencies.

PR #735, opened 2023-08-09
fix: types allow CSS vars in React player style

luwes:
SCR-20230809-gq9

### DIFF
--- a/packages/mux-player-react/src/index.tsx
+++ b/packages/mux-player-react/src/index.tsx
@@ -17,6 +17,11 @@ interface GenericEventListener<T extends Event = CustomEvent> {
   (evt: T): void;
 }
 
+export interface MuxPlayerCSSProperties extends CSSProperties {
+  // Allow any CSS Custom Properties
+  [index: `--${string}`]: any;
+}
+
 export type MuxPlayerRefAttributes = MuxPlayerElement;
 type VideoApiAttributes = {
   currentTime: number;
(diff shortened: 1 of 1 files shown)


packages/mux-player-react/src/index.tsx
```diff
@@ -17,6 +17,11 @@ interface GenericEventListener<T extends Event = CustomEvent> {
   (evt: T): void;
 }
 
+export interface MuxPlayerCSSProperties extends CSSProperties {
```
cjpillsbury, line 20: Since CSS Custom properties are inheritable via scope, we want to treat them as type-valid for any React element. Ideally, we'd enumerate all of our CSS vars in the type declaration, but it may be fine to use a similar "buckshot" templated version like you do here. For more, see: https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors

Here's an example that is produced for media-chrome's react type definitions:

```ts
import type * as CSS from 'csstype'; // A dependency from `@types/react`

// other crud...

declare module 'csstype' {
  interface Properties {
    // Should add generic support for any CSS variables
    [index: `--${string}`]: any;
  }
}
```

cjpillsbury (changes requested):
suggestion for alternative solution. Happy to get arguments for keeping as is, though it is not my S1.

luwes:
makes sense! maybe it's better for the consuming app to do that then, it feels strange to affect the whole app if the import our player

PR #737, opened 2023-08-11
feat(mux-player-react): re-export themes from player package

AdamJaggard:
Re-export themes from `mux-player-react` so that you don't need an explicit dependency on `mux-player` to import the optional themes.

image


luwes (approved):
awesome!

PR #738, opened 2023-08-14
feat: add a poster slot for progressive enhancement

luwes:
for progressive enhancement and using responsive images. it will also load
the poster faster as it doesn't have to wait on
the mux-player JS bundle to load


AdamJaggard (approved):
Seems like a nice improvement. Are there any docs updates that need to coincide with it?

luwes:
good call, yes that would be good to document!

luwes:
I'm gonna merge it already though because it would be very useful in the next stuff

PR #747, opened 2023-08-16
feat: add poster slot for progressive enhancement

### LINKED ISSUE #590: Show a static constant image in MuxPlayer for audio only sources

mux-player-react

### Description

When I use `<MuxPlayer>` with audio only files, I can use a `poster` prop to display an image. When the user clicks play, the poster image disappears, and the video player goes black. I'd like the ability to show a constant static video image in the player box from an audio only source. 

### Expected Behavior

1. Could there be a setting to use the `poster` prop of `<MuxPlayer>` to display permanently?
or
2. Can the `poster` prop be used as a permanent display when `audio` is `true`?

luwes:
related PR https://github.com/muxinc/media-chrome/pull/721

this change makes it possible to set a poster slot. it results in quicker loading of the thumbnail because the image url can be loaded in parallel with the player JS bundle. this also allows users to better make use of responsive images etc.

```html
<mux-player playback-id="VcmKA6aqzIzlg3MayLJDnbF55kX00mds028Z65QxvBYaA">
  <img slot="poster" src="https://image.mux.com/VcmKA6aqzIzlg3MayLJDnbF55kX00mds028Z65QxvBYaA/thumbnail.webp">
</mux-player>
```

this change also enables setting a poster for an audio player. by default we don't set a poster src if the mux player is in audio mode but the user can explicitly set the poster attribute and then the poster image will show.

fixes #590


examples/vanilla-ts-esm/public/mux-player-audio.html
```diff
@@ -48,6 +48,16 @@ <h1><a href="/">Elements</a></h1>
...
     ></mux-player>
 
+    <br>
+
+    <mux-player
+      audio
+      stream-type="on-demand"
+      playback-id="O4h5z00885HEucNNa1rV02wZapcGp01FXXoJd35AHmGX7g"
+      style="max-width: 480px"
+      poster="https://image.mux.com/23s11nz72DsoN657h4314PjKKjsF2JG33eBQQt6B95I/thumbnail.webp?width=640"
```
cjpillsbury, line 58: suggestion(non-blocking): maybe have two examples here: one with poster and one without? (can be on the same page imo)

cjpillsbury (approved):
🚢 🇮🇹 🎉

PR #750, opened 2023-08-17
feat: add audio track select menu

luwes:
- prevent duplicate audio tracks from different groups
- adds audio track selectmenu


cjpillsbury (approved):
Looks great!

PR #757, opened 2023-08-25
Fix/audio selectmenu 2023 theme

cjpillsbury:
* Adds `media-audio-track-selectmenu` to 2023 theme.
* Adds 2023 theme and includes some cleanup for MuxPlayer Next.JS demo page.


packages/mux-player/src/themes/2023/2023.html
```diff
@@ -732,6 +734,48 @@
...
   </template>
 
+  <template partial="AudioTrackSelect">
+    <media-audio-track-selectmenu
+      part="bottom audio-track selectmenu"
+      disabled="{{disabled}}"
+      aria-disabled="{{disabled}}"
+    >
+      <media-audio-track-button slot="button" part="bottom audio-track button">
+        <svg aria-hidden="true" slot="icon" viewBox="0 0 24 24">
```
AdamJaggard, line 744: All of our current theme icons are 18 x 14 so this has a ratio that doesn't match. I don't think it's a problem for merging it in now but I'll make a note to get an icon from Jason, I think we need a new icon for rendition selector anyway as we're currently using the very general triple dot one.

AdamJaggard (approved):
looks good! made a comment about the icon but we can sort it out in a separate PR

packages/mux-player/src/themes/2023/2023.html
cjpillsbury: cool agree with everything. Also the icon may change based on out of band convo, so probably don't need to belabor it too much on this PR as long as it's resolved before switching 2023 to default.

PR #760, opened 2023-08-31
chore: Add NPM keywords

mmcc:
I didn't spend a ton of time brainstorming what other keywords should be in these lists, but figured at least getting the ball rolling could help. Noticed we were missing these when I didn't see us come up in some searches, so hopefully this helps with some discoverability.


AdamJaggard (approved):
Looks like a good first pass.

PR #765, opened 2023-09-05
fix(mux-video,mux-audio): reload core on DOM connect

cjpillsbury:
Currently, we do an "unload"/teardown of core (and everything that happens in it/is created by it "under the hood") whenever our custom media elements are disconnected from the DOM. Ideally, this would not be the way we handle cleanups, as we're currently treating disconnect as a "`destroy()`" method. This causes issues with more advanced UIs/use cases, especially common in app frameworks where an element may be moved around the DOM for UI/UX purposes without the intention of destroying it. Unfortunately, it looks like we have some dangling references that result in a fairly large memory footprint per media element if we don't do this teardown. As a "middle ground" solution until we can identify a reasonable and reliable resolution of this larger memory management GC issue, we will simply (conditionally) "reload"/re-initialize core when the media element is (re)connected to the DOM.

### DIFF
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -375,6 +375,13 @@ class MuxAudioElement extends CustomAudioElement implements Partial<MuxMediaProp
     }
   }
 
+  connectedCallback(): void {
+    super.connectedCallback?.();
+    if (this.nativeEl && this.src && !this.#core) {
+      this.#requestLoad();
+    }
+  }
+
   disconnectedCallback() {
     this.unload();
   }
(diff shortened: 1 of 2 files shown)


AdamJaggard:
Unrelated to code, the title needs a `:` instead of a `/` 🙃

AdamJaggard (approved):
LGTM. Obviously not ideal but does the job and hopefully in the general case people aren't constantly disconnecting and re-connecting their components.

cjpillsbury:
> Unrelated to code, the title needs a `:` instead of a `/` 🙃

Yeah that's derived from my branch name, not my intended squash commit msg. That said, since we've moved to squash merges instead of rebase merges, I've def unintentionally merged with the wrong commit msg before.

PR #771, opened 2023-09-15
feat: Rename new default theme

AdamJaggard:
~~`classic` is now `daniel`, `2023` is now `gerwig`~~

~~This is technically a breaking change but because no one is likely aware that these recent exports exist we can decide how we want to handle merging. Seeing as switching to the new default theme is going to be a major version bump anyway we may want to hold off merging until we are preparing for that release?~~

Old theme is remaining as `classic`, new theme is now `gerwig`.

### DIFF
--- a/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
@@ -240,11 +240,8 @@ const ControlCustomizationCSSVars = [
   "--bottom-mute-button",
   "--volume-range",
   "--bottom-volume-range",
-  "--audio-track-selectmenu",
   "--rendition-selectmenu",
-  "--top-audio-track-selectmenu",
   "--top-rendition-selectmenu",
-  "--bottom-audio-track-selectmenu",
   "--bottom-rendition-selectmenu",
   "--playback-rate-button",
   "--bottom-playback-rate-button",
@@ -475,7 +472,7 @@ function MuxPlayerPage({ location }: Props) {
           value={state.theme}
           name="theme"
           onChange={genericOnChange}
-          values={['default', 'microvideo', 'minimal', '2023']}
+          values={['classic', 'microvideo', 'minimal', '2023']}
         />
         <TextRenderer
           value={state.envKey}
--- a/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
@@ -4,7 +4,7 @@ import Script from 'next/script';
 import MuxPlayer, { MuxPlayerProps } from "@mux/mux-player-react";
 import "@mux/mux-player/themes/minimal";
 import "@mux/mux-player/themes/microvideo";
-import "@mux/mux-player/themes/2023";
+import "@mux/mux-player/themes/gerwig";
 import { useEffect, useReducer, useRef, useState } from "react";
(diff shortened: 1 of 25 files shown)


luwes (changes requested):
could we have a few more people review this? @heff @mmcc 

it's in fun spirit which I like but it feels disconnected to me.
to me it didn't ring any bells, daniel or gerwig.

also how do these fit with the other themes `minimal` and `microvideo`?
in the future in player.style?

I'm leaning more to keep classic and just a new catchy name for the new theme.

it's pink and bubbly. gerwig doesn't feel very catchy to me

cjpillsbury:
> could we have a few more people review this? @heff @mmcc

Happy to get input from 👆, but since this is naming the official elements themes, I believe Dylan would be a a more appropriate reviewer/decision maker to loop in.

> it's in fun spirit which I like but it feels disconnected to me. to me it didn't ring any bells, daniel or gerwig.

This was based on a conversation with @dylanjha and I had that some other folks chimed in on out of band. The idea was taking our own spin on codenames like e.g. Android's, where the name is honoring a writer/director of a movie from the year that the theme was created (to that end, I'm not sure if we settled on "daniel" vs. "kwan" out of band). Like the Android or other naming conventions, they typically aren't going to have hard associations outside of the place that's establishing them (what does "eclair" or "froyo" mean outside of the fact that folks expect Android OS codenames to be desert-y? Will it tell you which version something is, or when it was made, etc.?).

> also how do these fit with the other themes minimal and microvideo?
> in the future in player.style?

We've previously decided that we were going to treat the "official" mux player themes differently from "generic media chrome themes," not just in where the code for the themes live (part of elements/mux player package as opposed to e.g. media chrome), but in how we will/won't treat them as "readily usable for things other than mux player". Even though we designed things like the `minimal` and `microvideo` themes, these were created as generic media chrome themes, so I wouldn't say we need to worry about naming consistency there.

cjpillsbury:
It's also worth noting that we could _**also**_ treat these as "aliases" and simply keep the year or some other versioning in place so folks could use either/both to reference their theme, whatever "fun codename" we decide to assign.

dylanjha:
> I'm leaning more to keep classic and just a new catchy name for the new theme.

@luwes -- can you make a suggestion for a new name?

> it's pink and bubbly. gerwig doesn't feel very catchy to me

Gerwig is a reference to Greta Gerwig, which is the director of Barbie (pink and bubbly), seems okay to me?

I'm open to other suggestions.

AdamJaggard:
I think keeping `classic` and naming the new one `gerwig` is a good middle ground. Adding aliases feels a bit messy.

luwes:
@cjpillsbury yep, I saw the Slack convo's. that's why I didn't cc Dylan because he's in the loop.

pink, bubbly -> barbie -> gerwig is a step too far IMO.
I don't think 50% gets the reference. at least with eclair, froyo something could be visualized

why not call it barbie then? easier to remember 

other suggestions: bubblegum, rosa (rose), pinkpop, pinkfizz, jellyfish, 
https://www.google.com/search?q=what%27s+pink+in+nature

cjpillsbury:
Cool. No notes on my side.

AdamJaggard:
@cjpillsbury, @dylanjha, @luwes, I've reverted old theme back to `classic`, new theme is still `gerwig`.

As this now only affects the new theme, it's a safer merge.

luwes (approved):
LGTM 👍

PR #772, opened 2023-09-16
feat: add startLevel prop to mux-video-react

staniwasd:
Hey team,

A small PR that allows you to add the startLevel prop in the MuxVideo react component.

API doc - https://github.com/video-dev/hls.js/blob/master/docs/API.md#startlevel

### DIFF
--- a/packages/mux-video-react/src/index.tsx
+++ b/packages/mux-video-react/src/index.tsx
@@ -24,7 +24,7 @@ const playerSoftwareVersion = getPlayerVersion();
 const playerSoftwareName = 'mux-video-react';
 
 const MuxVideo = React.forwardRef<HTMLVideoElement | undefined, Partial<Props>>((props, ref) => {
-  const { playbackId, src: outerSrc, children, autoPlay, preload, streamType, ...restProps } = props;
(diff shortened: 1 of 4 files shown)


cjpillsbury:
Hey @staniwasd,

Appreciate the eager and proactive engagement with the Open Elements project! Some way of programmatically stipulating a preferred initial rendition/media playlist sounds like a good feature add. For these kinds of changes, we'll want to discuss with the team before merging them, including details on how we'd want to solve that general problem.

Just in case we don't accept this PR, I'd recommend starting with filing a [Feature Request Issue](https://github.com/muxinc/elements/issues/new?assignees=&labels=enhancement&projects=&template=feature_request.yml&title=Feature+Request%3A+) to avoid any wasted effort on your part. Also, when it comes to directly exposing `hls.js` API details (including config), we've been very intentional about the shape of this API, as we don't want to be overly-coupled to relying on this as our underlying "playback engine". Additionally, if we do end up exposing some API for starting rendition/level/playlist, we'll probably want that to be implemented for the other elements as well.

Let me check in with the team and get back to you on how we'd like to approach this! Assuming folks from the team are on board, if you're still interested in contributing, we can let you know what changes we'd want for a more general solution here.

Thanks again!

cjpillsbury:
Closing this. Intended purpose of the changes should be covered by #833

PR #774, opened 2023-09-19
fix: improve 2023 icons

luwes:
after:

SCR-20230919-q6t

before:

SCR-20230919-q6i


packages/mux-player/src/themes/2023/2023.html
```diff
@@ -255,8 +261,6 @@
       --media-control-hover-background: transparent;
       --media-control-padding: 0;
       width: 40px;
-      transform-origin: center center;
-      transform: translate(3px, 0);
```
luwes, line 259: @AdamJaggard this was causing a 3px shift when pressing the center play button because the pre-play-hide animation transform would take over. 

I didn't see the need for this right away.

AdamJaggard (approved):
Looks good!

PR #759, opened 2023-08-29
Fix/mta ll hls bugs

cjpillsbury:
This resolves an issue where hangs were occurring with initialization of MTA on LL-HLS streams. Also added:
- some quick fixes for presumptuous existence
- patch version bumps for hls.js and mux-embed
- refactor the monorepo/workspace-level dev dependencies on ☝️ to only target major versions for easier maintainability

Validating:
- https://elements-demo-nextjs-git-fork-cjpillsbury-fix-mta-ll-d0bce1-mux.vercel.app/MuxPlayer?muted=true&autoPlay=true&title=%22Long+Lived+LL-HLS%22&playbackId=%22v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM%22&metadata=%7B%22video_id%22%3A%22videoId11%22%2C%22video_title%22%3A%22Title%3A+Long+Lived+LL-HLS%22%7D should load without playback getting "stuck" loading
- Compare with equivalent current deploy: https://elements-demo-nextjs-mux.vercel.app/MuxPlayer?muted=true&autoPlay=true&title=%22Long+Lived+LL-HLS%22&playbackId=%22v69RSHhFelSm4701snP22dYz2jICy4E4FUyk02rW4gxRM%22&metadata=%7B%22video_id%22%3A%22videoId11%22%2C%22video_title%22%3A%22Title%3A+Long+Lived+LL-HLS%22%7D

PR #776, opened 2023-09-22
fix: audio track API for 1 track

luwes:
reverts a bug fix in https://github.com/muxinc/elements/pull/759

it was a good patch but it prevented the audio tracks API to be populated if there was just 1 audio track.
it's still expected that with 1 audio track, the API returns an item for video.audioTracks[0].

this change adds that back in.

the root issue in hls.js I believe was that we were trying to select an audio track with a string ID and hls.js works with numeric ID's.
this change also adds a check if hls.js audioTracks has the ID to which the player tries to switch.


cjpillsbury (approved):
👌

PR #780, opened 2023-10-02
fix(playback-core): Use optional chaining and nullish defaulting for …

cjpillsbury:
…edge case RTE.

A Mux Video customer is currently seeing some edge case RTEs reported into their error reporting system on iPhones. Although the root cause hasn't been identified, we've narrowed it to the line changed in this PR. Added a code comment explaining the change, even though the condition should never be met to cause the RTE in the first place.


AdamJaggard (approved):
LGTM

luwes (approved):
LGTM

PR #784, opened 2023-10-03
feat(mux-player)!: switch to new default theme

AdamJaggard:
also fixes a couple css selector regressions


cjpillsbury (approved):
The future is now.

PR #795, opened 2023-10-12
feat(playback-core, mux-video): Initial impl of manifest manipulation…

cjpillsbury:
… support.

This PR will add corresponding properties/attributes to the media elements to support Mux Video's new manifest manipulation query params features. These include:
* expanding the supported range of values for `maxResolution` (prop) / `max-resolution` (attr) (query param `max_resolution`)
* a new `minResolution` (prop) / `min-resolution` (attr) (query param `min_resolution`), which will remove any renditions below a particular resolution
* a new `renditionOrder` (prop) / `rendition-order` (attr) = `"desc"` (query param `rendition_order`), which will ensure that the renditions are sorted in descending order (which mostly impacts certain default initial rendition/segment selection behavior)

Additionally, it adds closer parity of supported and respected props/attrs related to mux video url across other media elements/packages, does some "under the hood" reduction of code redundancy and unnecessary complexity (see inline notes in the code comments for examples).


packages/mux-player/test/helpers.test.js
```diff
@@ -1,37 +1,12 @@
...
-} from '../src/helpers.ts';
+import { getPosterURLFromPlaybackId, getStoryboardURLFromPlaybackId, AttributeTokenList } from '../src/helpers.ts';
 
 // token key Key-Must-Be-at-least-32-bytes-in-length!
 // generated via https://dinochiesa.github.io/jwt/
 const THUMBNAIL_TOKEN = 'eyJhbGciOiJIUzI1NiJ9.eyJhdWQiOiJ0In0.BUhYQ9GTTFrrqyfT_7d2hb7FCnpmdRnzHPJulYU_yUo';
 const STORYBOARD_TOKEN = 'eyJhbGciOiJIUzI1NiJ9.eyJhdWQiOiJzIn0.aYYPU6VETdQOF04OEbhDYPOCgIxmngVBIfC6BQXbgI8';
 
 describe('helpers', () => {
-  describe('getSrcFromPlaybackId', () => {
```
cjpillsbury, line 15: NOTE: Using playback core's `toMuxVideoURL` directly now, as it has functional parity.

packages/mux-player/test/player.test.js
```diff
@@ -523,10 +523,14 @@ describe('<mux-player>', () => {
     const muxVideo = player.media;
 
     assert.equal(player.maxResolution, '720p');
-    assert.equal(
-      muxVideo.src,
+    const actualSrcUrl = new URL(muxVideo.src);
+    const expectedSrcUrl = new URL(
       'https://stream.mux.com/r4rOE02cc95tbe3I00302nlrHfT023Q3IedFJW029w018KxZA.m3u8?redundant_streams=true&max_resolution=720p'
     );
+    assert.equal(actualSrcUrl.searchParams.size, expectedSrcUrl.searchParams.size);
```
cjpillsbury, line 530: NOTE: order of query params changed by refactor, so updated test to reflect what it's actually asserting: that the query param keys and values match.

packages/playback-core/src/index.ts
```diff
@@ -230,12 +228,66 @@ export const generatePlayerInitTime = () => {
...
+  renditionOrder: RenditionOrderValue;
+  tokens: Partial<{
+    playback: string;
+    storyboard: string;
+    thumbnail: string;
+  }>;
+  extraPlaylistParams: Record<string, any>;
+}>;
+
+export const toMuxVideoURL = ({
```
cjpillsbury, line 245: NOTE: Updated the params here so we can directly pass in react props or the web component instance, hopefully improving maintainability, particularly across components/elements.

packages/mux-player/src/index.ts
```diff
@@ -144,6 +151,7 @@ function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
     customDomain: el.getAttribute(MuxVideoAttributes.CUSTOM_DOMAIN) ?? undefined,
     title: el.getAttribute(PlayerAttributes.TITLE),
     novolumepref: el.hasAttribute(PlayerAttributes.NO_VOLUME_PREF),
+    extraPlaylistParams: { redundant_streams: true },
```
cjpillsbury, line 154: NOTE: This will be made into a prop (and possibly corresponding attr) in a followup PR to allow for arbitrary query param additions, where `{ redundant_streams: true }` will be treated as the default value of the prop.

packages/mux-player/src/template.ts
```diff
@@ -71,24 +71,8 @@ export const content = (props: MuxTemplateProps) => html`
...
-      cast-src="${!!props.src
-        ? props.src
-        : props.playbackId
-        ? getSrcFromPlaybackId(props.playbackId, {
-            maxResolution: props.maxResolution,
-            domain: props.customDomain,
-            token: props.tokens.playback,
-          })
-        : false}"
+      src="${!!props.src ? props.src : props.playbackId ? toMuxVideoURL(props) : false}"
```
cjpillsbury, line 74: NOTE: with both the refactored function signature and removing the similarly-purposed `getSrcFromPlaybackId`, this should make code maintenance easier.

packages/mux-video/src/index.ts
```diff
@@ -47,34 +50,15 @@ const CustomVideoElement = MediaTracksMixin(
...
-  PLAYER_SOFTWARE_VERSION: 'player-software-version';
-  PREFER_CMCD: 'prefer-cmcd';
-  PREFER_PLAYBACK: 'prefer-playback';
-  START_TIME: 'start-time';
-  STREAM_TYPE: 'stream-type';
-  TARGET_LIVE_WINDOW: 'target-live-window';
-  LIVE_EDGE_OFFSET: 'live-edge-offset';
-  TYPE: 'type';
-};
-
```
luwes, line 70: ❤️

packages/playback-core/src/index.ts
```diff
@@ -230,12 +228,66 @@ export const generatePlayerInitTime = () => {
...
+   * All identified query params here can only be added to public
+   * playback IDs. In order to use these features with signed URLs
+   * the query param must be added to the signing token.
+   *
+   * */
+  if (token || url.searchParams.has('token')) {
+    url.searchParams.forEach((_, key) => {
+      if (key != 'token') url.searchParams.delete(key);
+    });
+    if (token) url.searchParams.set('token', token);
```
luwes, line 267: here's some unneeded work done I think.
either drop this line `if (token) url.searchParams.set('token', token);`
OR reset the whole `url.searchParams = new URLSearchParams()` and leave the set token line

packages/mux-player/src/index.ts
luwes: since there is already `playbackid?token=dsdsds`
should this work similarly?

luwes (approved):
LGTM, I have 1 request removing a potential unneeded line.

packages/mux-player/src/index.ts
cjpillsbury: I started there with this impl actually, but ran into issues with our state update infrastructure (bottoms out in another `...state` overwrite issue in `getProps()` that we've encountered before with other props). Happy to brainstorm together/revisit though if you'd like. Keeping it separate does feel overall cleaner/easier/less complex imo though.

packages/playback-core/src/index.ts
cjpillsbury: `url.searchParams` is read only 💔 I actually tried that for the `forEach()` that deletes any additional query params above. We could do **_different_** work to replace the `url.searchParams` with a refactor for sure. One complexity though is that in our current component implementations, there are two different ways of receiving a token value:
1. the `token` value may come in via `playbackId=`MYPLAYBACKID?token=MYTOKEN`
2. the `token` value may come in via `token.playback=MYTOKEN`
3. the `token` may be specified on both (in which case we should have some rule on behavior)
We could certainly refactor the conditional paths also, but we will still need to set the `token` in some form for case at least case (2) (here, it's 2 & 3, mostly because that was my S1 on what the rule should be). Lmk your thoughts here.

PR #797, opened 2023-10-17
fix: Don't allow clicks on bg gradients

AdamJaggard:
The bg gradients can sometimes swallow clicks. At small sizes the bottom one starts to encroach on the center controls and can swallow clicks if they're towards the bottom of the center play and skip buttons. This could also be fixed by tweaking z-indexes but that could have other effects.


luwes (approved):
LGTM

PR #800, opened 2023-10-20
feat: Add volume slider to live controls

AdamJaggard:
The live version of the new theme has a mute button but no volume slider. This is actually how it is in the designs but I think it's an omission and a customer asked where it was. Have added it back in and uncommented out the live preview in our vanilla example.


luwes (approved):
LGTM

PR #805, opened 2023-10-31
fix: upgrade MC + style tweaks

luwes:
The listbox options of Gerwig were not the same height when selected vs not selected.
The new range requires padding values that can be calculated.

### DIFF
--- a/packages/mux-player/src/themes/gerwig/gerwig.html
+++ b/packages/mux-player/src/themes/gerwig/gerwig.html
@@ -90,9 +90,10 @@
         var(--_accent-color) 25%,
         var(--_accent-color)
       );
-      --media-range-thumb-width: 0;
-      --media-range-thumb-height: 0;
-      --media-range-thumb-transition: all 0.3s;
+      --media-range-thumb-width: 12px;
+      --media-range-thumb-height: 12px;
+      --media-range-thumb-transform: scale(0);
+      --media-range-thumb-transition: transform 0.3s;
       --media-range-thumb-opacity: 1;
       --media-preview-thumbnail-background: var(--_primary-color);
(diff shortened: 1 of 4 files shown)


packages/mux-player/src/themes/gerwig/gerwig.html
```diff
@@ -90,9 +90,10 @@
         var(--_accent-color) 25%,
         var(--_accent-color)
       );
-      --media-range-thumb-width: 0;
-      --media-range-thumb-height: 0;
-      --media-range-thumb-transition: all 0.3s;
+      --media-range-thumb-width: 12px;
+      --media-range-thumb-height: 12px;
+      --media-range-thumb-transform: scale(0);
+      --media-range-thumb-transition: transform 0.3s;
```
luwes, line 96: transform scale is preferred over width, height because it's centered by default and performs better.

packages/mux-player/src/themes/gerwig/gerwig.html
```diff
@@ -132,40 +132,33 @@
...
     }
 
-    :is(media-captions-listbox, media-rendition-listbox, media-audio-track-listbox, media-playback-rate-listbox)::part(
-        indicator
-      ) {
+    [part~='listbox'] {
+      border-radius: 2px;
+      border: 1px solid rgba(0, 0, 0, 0.1);
+      bottom: calc(100% + 16px);
+      padding: 2.5px 10px;
```
luwes, line 139: the internal container already had 7.5px vertical padding so just add 2.5px here for 10px like in the design

AdamJaggard (approved):
Listbox selector cleanup looks much cleaner!

PR #807, opened 2023-10-31
Update Yarn to v4

smorimoto:
At the moment, multiple versions of yarn can be mixed, so it's quite tricky to develop. So in this PR use [corepack](https://github.com/nodejs/corepack) to make it possible to use yarn v4 without much effort. All you or the CI machine need to do is execute `corepack enable`.


luwes:
@smorimoto thanks for the contributions! we discussed this internally briefly and we're going to stick with yarn v1 for the time being as it doesn't give any huge benefits and it has the potential to lead to other things breaking.

personally I haven't seen much adoption of corepack. is this being used more and more in open source?

PR #816, opened 2023-11-10
Fix/media chrome attr name mismatches

cjpillsbury:
a few missed cases of switching to lowercase/"smushedcase". Moved inline "magic value" strings for attr names to rely on media-chrome constants instead to better avoid this in the future.


luwes (approved):
Nice!

Issue #821, opened 2023-11-14
Missing playback, thumbnail and storyboard token parameters in mux-player-react


mmvsk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player-react

### Which browsers are you using?

Chrome

### Which operating systems are you using?

Other (add details below)

### Description

MuxPlayerReact is missing the following component props:

- `playbackToken` (passed as `playback-token` to `mux-player`)
- `thumbnailToken` (passed as `thumbnail-token` to `mux-player`)
- `storyboardToken` (passed as `storyboard-token` to `mux-player`)

### Steps to reproduce

1. Try to add `playbackToken`, `thumbnailToken` or `storyboardToken`
2. Those props does not exist

### Current Behavior

Not implemented.

### Expected Behavior

Should be implemented.

### What version of the package are you using?

v2.2.0

PR #822, opened 2023-11-14
Fix (mux-player-react): missing token props (#821)

mmvsk:
Add missing `playbackToken`, `thumbnailToken` and `storyboardToken` to MuxPlayer react component.

### DIFF
--- a/packages/mux-player-react/src/index.tsx
+++ b/packages/mux-player-react/src/index.tsx
@@ -63,6 +63,9 @@ type MuxMediaPropTypes = {
   storyboardSrc: string;
   preferCmcd: ValueOf<CmcdTypes> | undefined;
   children?: React.ReactNode;
+  playbackToken?: string;
+  thumbnailToken?: string;
+  storyboardToken?: string;
 };
 
 export type MuxPlayerProps = {
@@ -178,6 +181,9 @@ const usePlayer = (
     paused,
     playbackId,
     playbackRates,
+    playbackToken,
+    thumbnailToken,
+    storyboardToken,
     currentTime,
     themeProps,
     ...remainingProps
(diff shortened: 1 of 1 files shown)


dylanjha:
Hi @mmvsk -- good eye.

`<MuxPlayer />` the React component actually has a `tokens` prop that takes an object: `{ playback: string; thumbnail: string; storyboard: string; }`

Since React allows for setting objects as props, this felt like a cleaner API compared to the web component which requires attribute values to be strings. That's why we went with:

Web component:

```html
<!-- web component -->
<mux-player
  playback-token="eyJhbGciOiJSUzI1NiI..."
  thumbnail-token="eyJhbGciOiJSUzI1N..."
  storyboard-token="eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsI..."
  ></mux-player>
```

React:

```jsx
// React component
<MuxPlayer
  tokens={{
    playback: "eyJhbGciOiJSUzI1NiI...",
    thumbnail: "eyJhbGciOiJSUzI1N...",
    storyboard: "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsI...",
  }}
/>
```

Sidenote for the web component usage:, when using the `<mux-player>` HTMLL element, it exposes a similar _property_ in Javascript where the `tokens=` value can be set to an object:

```js
// web component
const muxPlayer = document.querySelector("mux-player");
muxPlayer.tokens = {
  playback: "eyJhbGciOiJSUzI1NiI...",
  thumbnail: "eyJhbGciOiJSUzI1N...",
  storyboard: "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsI...",
};
```

We have an [example here in the docs](https://docs.mux.com/guides/player-advanced-usage#using-with-signed-urls)

If you have a suggestion for where we could better document this, please let me know! Closing b/c this is the intended behavior -- thank you for the PR though, appreciate the effort!

mmvsk:
That's right, I totally missed the relevant part in the docs.

Now I see, it's pretty clear :)

PR #824, opened 2023-11-15
fix(mux-player): upgrade media chrome to get fix for android tap issu…

cjpillsbury:
…es on controls show and hide.


luwes (approved):
LGTM

PR #825, opened 2023-11-15
Fix/narrow cmcd for cache

cjpillsbury:
For the time being, only send `sid` and `cid` CMCD k/v pairs on requests (if cmcd is on and using query params). This should improve cache hits during e.g. seeks and loops due to buffer flushes + refetches.


luwes (approved):
lgtm!

PR #833, opened 2023-11-27
fix: Adding unofficial _hlsConfig to media elements and playback core.

### LINKED ISSUE #792: Feature Request: Add startLevel prop to control the quality of first video fragment

mux-video, mux-video-react, mux-player, mux-player-react

### Description

Hey!

A fairly common problem these days is that a stream starts with a lower video quality, and after checking your internet connection, the next segment is generally of better quality. The easiest way to reproduce it is to open any stream on Twitch. This may be fine for long streams, but in the case of a short video, especially when you loop it, it becomes a more serious problem because this low quality fragment remains in the cache.

The current library ([hls.js](https://github.com/video-dev/hls.js)) on which the MUX components are built has a **startLevel** parameter that allows you to control the quality of the first fragment. Having this capability will help improve the user experience by providing a better quality stream from the start.

This can be a separate prop **startLevel**, or a general prop, for example **hlsConfig**, which allows you to override all params in hls.

```
// React example

<MuxVideo
     src={src}
     startLevel={999} <-- this maximizes the quality of the first fragment
     ...
/>
```

hls.js API doc - https://github.com/video-dev/hls.js/blob/master/docs/API.md#startlevel
Pull request with the change - https://github.com/muxinc/elements/pull/772

### Expected Behavior

Ability to control the quality of the first video fragment through the startLevel property.

cjpillsbury:
🤫 seeeecret config.
fixes #792

### DIFF
--- a/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayer.tsx
@@ -351,6 +351,11 @@ function MuxPlayerPage({ location }: Props) {
         theme={state.theme}
         envKey={state.envKey}
         metadata={state.metadata}
+        // Test _hlsConfig for MuxPlayer (react) (Note: This also indirectly tests <mux-player> & <mux-video>)
+        // _hlsConfig={{
+        //   startLevel: 2,
+        //   debug: true,
+        // }}
         title={state.title}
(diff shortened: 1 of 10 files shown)


luwes (approved):
LGTM

PR #834, opened 2023-11-27
Feat/arbitrary src params

cjpillsbury:
Mux Player `extraPlaylistParams` (prop) / `extra-playlist-params` (attr) allows users to specify any arbitrary search/query params for the resultant `src` URL (based on, e.g. `playbackId` and other specific props/attrs used to construct the URL).

### DIFF
--- a/packages/mux-player/src/index.ts
+++ b/packages/mux-player/src/index.ts
@@ -74,7 +74,7 @@ const PlayerAttributes = {
   THEME: 'theme',
   DEFAULT_STREAM_TYPE: 'default-stream-type',
   TARGET_LIVE_WINDOW: 'target-live-window',
-  EXTRA_PLAYLIST_PARAMS: 'extra-playlist-params',
+  EXTRA_SOURCE_PARAMS: 'extra-source-params',
   NO_VOLUME_PREF: 'no-volume-pref',
 };
 
@@ -157,7 +157,7 @@ function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
     ...state,
     // NOTE: since the attribute value is used as the "source of truth" for the property getter,
     // moving this below the `...state` spread so it resolves to the default value when unset (CJP)
-    extraPlaylistParams: el.extraPlaylistParams,
+    extraSourceParams: el.extraSourceParams,
   };
 
   return props;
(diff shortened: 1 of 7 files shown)


packages/mux-player/src/index.ts
```diff
@@ -1221,6 +1224,27 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
     }
   }
 
+  get extraPlaylistParams() {
+    // NOTE: the official type definition for the 0th `URLSearchParams` constructor argument is stricter than what is actually
+    // allowed and commonly used (e.g. a key with a boolean or number value). Ignoring to allow these cases. (CJP)
+    // @ts-ignore
+    if (!this.hasAttribute(PlayerAttributes.EXTRA_PLAYLIST_PARAMS)) return DEFAULT_EXTRA_PLAYLIST_PARAMS;
+    return [
+      ...new URLSearchParams(this.getAttribute(PlayerAttributes.EXTRA_PLAYLIST_PARAMS) as string).entries(),
```
cjpillsbury, line 1233: NOTE: Uses `URLSearchParams` constructor to more easily/reliably parse the attribute string, which acts as the "source of truth" for the value.

packages/mux-player/src/index.ts
```diff
@@ -153,8 +154,10 @@ function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
     customDomain: el.getAttribute(MuxVideoAttributes.CUSTOM_DOMAIN) ?? undefined,
     title: el.getAttribute(PlayerAttributes.TITLE),
     novolumepref: el.hasAttribute(PlayerAttributes.NO_VOLUME_PREF),
-    extraPlaylistParams: { redundant_streams: true },
     ...state,
+    // NOTE: since the attribute value is used as the "source of truth" for the property getter,
+    // moving this below the `...state` spread so it resolves to the default value when unset (CJP)
+    extraPlaylistParams: el.extraPlaylistParams,
```
cjpillsbury, line 160: We should consider this convention for all of the acrobatics we've had with our generic state update from attr values in `attributeChangedCallback()`. tl;dr - if our mux-player prop value is always derived from the corresponding mux-player attr value, we should move those below our `state` spread. (cc @luwes)

PR #835, opened 2023-11-28
Add disablePictureInPicture to player prop types

### LINKED ISSUE #831: Bug: `disablePictureInPicture` prop not allowed

mux-video-react

Chrome, Safari

None

### Description

Hey team 👋 

Just a little one, but I'm seeing a TypeScript error when using the [`disablePictureInPicture`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/disablePictureInPicture) prop

It works as expected, so the error is only at the typing level. 

I've worked around it for now with this dirty little hack, just to silence the error:

```tsx
 <MuxPlayer {...{ disablePictureInPicture: true }} />
```

Thanks!

### Reduced test case

_No response_

### Steps to reproduce

1. `import MuxPlayer from "@mux/mux-player-react";`
2. Render `<MuxPlayer />`
3. Add the `disablePictureInPicture` prop

### Current Behavior

image

### Expected Behavior

No TS errors

### Errors

_No response_

2.2.0

heff:
Fixes #831

Not sure if this is the right place to add this.

### DIFF
--- a/packages/mux-player-react/src/index.tsx
+++ b/packages/mux-player-react/src/index.tsx
@@ -70,6 +70,7 @@ export type MuxPlayerProps = {
   hotkeys?: string;
   nohotkeys?: boolean;
   defaultHiddenCaptions?: boolean;
+  disablePictureInPicture?: boolean;
   playerSoftwareVersion?: string;
   playerSoftwareName?: string;
   forwardSeekOffset?: number;


packages/mux-player-react/src/index.tsx
```diff
@@ -70,6 +70,7 @@ export type MuxPlayerProps = {
   hotkeys?: string;
   nohotkeys?: boolean;
   defaultHiddenCaptions?: boolean;
+  disablePictureInPicture?: boolean;
```
cjpillsbury, line 73: nitpick: Should be added to `MuxMediaPropTypes`, just above (which is part of the `MuxPlayerProps` type def, found just below)

cjpillsbury (changes requested):
minor nitpick but general sentiment was right

packages/mux-player-react/src/index.tsx
heff: Fixed

cjpillsbury (approved):
👌

PR #840, opened 2023-12-06
fix: upgrade media-chrome v2 & castable-video v1

luwes:


yarn.lock
```diff
@@ -6400,7 +6402,7 @@ csstype@^3.0.2:
   resolved "https://registry.yarnpkg.com/csstype/-/csstype-3.1.1.tgz#841b532c45c758ee546a11d5bd7b7b473c8c30b9"
   integrity sha512-DJR/VvkAvSZW9bTouZue2sSxDwdTN92uHjqeKVm+0dAqdfNykRzQ95tay8aXMBAAPpUiq4Qcug2L7neoRh2Egw==
 
-custom-media-element@1.1.0:
+custom-media-element@1.1.0, custom-media-element@^1.0.2:
```
cjpillsbury, line 6405: thought(non-blocking): double version here. Looks like it's due to `castable-video.dependencies`. Trying to think of ways we can avoid this, especially since the only reason it's a dependency is for the convenience use case of `castable-video` definition https://github.com/muxinc/castable-video/blob/362939f5cd2445759c0080ae746cb6300cecb0c9/castable-video.js#L4-L11. I don't think there's anything immediately actionable here, just something to chew on.

cjpillsbury (approved):
👌

PR #841, opened 2023-12-06
fix(mux-player): gerwig theme misapplying opt in controls selector.

cjpillsbury:
Also fixes audio only styling issues for gerwig.


cjpillsbury:
Closing. Better fix for theme colors from @luwes in #843

PR #845, opened 2023-12-07
examples: remove unneeded cast framework script

luwes:
this script is not needed anymore, castable-video loads it automatically.

also if the script is loaded twice it causes an error in the console but doesn't break the player or page.

SCR-20231206-rbt

this is only an issue when the users script is loaded after castable-video already appended its own script.
Next.js script tag is loaded afterInteractive, that's why it is causing the error log.

if it's loaded before castable-video checks that and prevents loading
see https://github.com/muxinc/castable-video/blob/0e5f322d4d1b7dd53215e4102d5712f8aa2721aa/castable-utils.js#L50-L50


cjpillsbury (approved):
LGTM for the fix. We should follow up to ideally keep this an "opt in" implementation for chromecast.

PR #851, opened 2023-12-13
fix: start-time for live streams

luwes:
this change also removes the code that would explicitly jump to the live edge in autoplay.ts
It was part of the bug but it is actually not needed, hls.js and native HLS will jump to the live edge automatically.

I tested this change with all the `preload` and `autoplay` options and it worked as expected.


luwes:
closing this fix for now until it comes up again

PR #857, opened 2023-12-22
Correct unused or unlisted dependencies

smorimoto:
I checked them with my own eyes after scanning them with knip, but just in case, could you give me a review?


smorimoto:
Ah, knip is https://knip.dev!

PR #867, opened 2024-02-06
fix/pseudo ended stalled last frag

cjpillsbury:
Adds logic to see if we've stalled at the end of playback, even if we've already buffered (part of) the last segment. Accounts for time alignment issues and potential time drift caused by hls.js transmuxing logic.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -686,8 +686,9 @@ export const loadMedia = (
     if (isStuckOnLastFragment(mediaEl, hls)) {
       // Nudge the playhead in this case.
       mediaEl.currentTime = mediaEl.buffered.end(mediaEl.buffered.length - 1);
+    } else {
+      mediaEl.dispatchEvent(new Event('ended'));
     }
-    mediaEl.dispatchEvent(new Event('ended'));
   };
 
   if (mediaEl && shouldUseNative) {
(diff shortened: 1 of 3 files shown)


packages/playback-core/src/index.ts
```diff
@@ -337,17 +337,76 @@ export const getLiveEdgeStart = (mediaEl: HTMLMediaElement) => {
   return seekable.end(seekable.length - 1) - liveEdgeStartOffset;
 };
 
-const isApproximatelyEqual = (x: number, y: number, moe = 0.001) => Math.abs(x - y) <= moe;
-const isApproximatelyGTE = (x: number, y: number, moe = 0.001) => x > y || isApproximatelyEqual(x, y, moe);
+const DEFAULT_ENDED_MOE = 0.034;
```
AdamJaggard, line 340: What does MOE stand for?

cjpillsbury: "margin of error". MOE is a common acronym for it, but this has come up in the past since not all devs are familiar with the terminology.

AdamJaggard: Ah, that makes sense. Familiar with the term, but had never come across it abbreviated before somehow

cjpillsbury: I just like naming my variables after Simpsons characters. 😄 jk

luwes (approved):
LGTM! hopefully in the future we can remove some or all of these workarounds

PR #883, opened 2024-03-11
Fix/live seek and media chrome upgrade

cjpillsbury:
This resolves some edge cases with live seeking behavior and corresponding UI, which are particularly relevant for new pdt clipping available from Mux Video.

It also implements the `"seekablechange"`, which can be used by Media Chrome to more reliably propagate `mediaSeekable` state changes.

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -691,13 +691,57 @@ export const loadMedia = (
     }
   };
 
+  let prevSeekableStart: number;
+  let prevSeekableEnd: number;
+
+  const seekableChange = () => {
+    const nextSeekableStart = getSeekable(mediaEl)?.start(0);
+    const nextSeekableEnd = getSeekable(mediaEl)?.end(0);
+    if (prevSeekableEnd !== nextSeekableEnd || prevSeekableStart !== nextSeekableStart) {
+      mediaEl.dispatchEvent(new CustomEvent('seekablechange', { composed: true, bubbles: true }));
+    }
+    prevSeekableStart = nextSeekableStart;
+    prevSeekableEnd = nextSeekableEnd;
+  };
+
   if (mediaEl && shouldUseNative) {
     const type = getType(props);
     if (typeof src === 'string') {
+      // NOTE: This should only be invoked after stream type has been
+      // derived after stream type has been determined.
+      const setupSeekableChangePoll = () => {
+        // Only monitor for seekable updates if StreamType is "live" and duration is not finite.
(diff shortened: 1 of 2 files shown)


packages/mux-player/package.json
```diff
@@ -96,7 +96,7 @@
   "dependencies": {
     "@mux/mux-video": "0.17.3",
     "@mux/playback-core": "0.22.2",
-    "media-chrome": "~2.2.5"
+    "media-chrome": "^3.0.2"
```
luwes, line 99: stick to patch versions?

cjpillsbury: ah yeah good catch/callout. 🦅 👀. Def agree.

packages/playback-core/src/index.ts
```diff
@@ -691,13 +691,57 @@ export const loadMedia = (
...
   };
 
+  let prevSeekableStart: number;
+  let prevSeekableEnd: number;
+
+  const seekableChange = () => {
+    const nextSeekableStart = getSeekable(mediaEl)?.start(0);
+    const nextSeekableEnd = getSeekable(mediaEl)?.end(0);
+    if (prevSeekableEnd !== nextSeekableEnd || prevSeekableStart !== nextSeekableStart) {
+      mediaEl.dispatchEvent(new CustomEvent('seekablechange', { composed: true, bubbles: true }));
```
luwes, line 701: should we keep this the same to native video events?
bubbles is false, composed false also

luwes:
Looks good, 1 blocking ask!

packages/playback-core/src/index.ts
cjpillsbury: We'll need `composed` to be `true` for it to be available to e.g. Media Chrome/`MediaStore` (or do a bunch of extra work to delegate.

`bubbles` can definitely be `false`, and would be consistent with the recent (**_finally_**) updates to our Media Chrome state change events. I used `true` because that's what we use for all of our other custom events in elements, but happy to start undoing that pattern/habit, starting with `seekablechange`.

luwes: ah k I see, then probably best to leave it for now. bubbles might be needed too for it to go from <video> to <mux-video> or does composed handle that...

luwes (approved):
👍

packages/playback-core/src/index.ts
cjpillsbury: that's... a good question. I'll test that now.

cjpillsbury: for archival purposes - `bubbles` not needed for shadow DOM boundary propagation (only `composed`), so removing. 🎉

PR #885, opened 2024-03-15
fix: polish, use easing gradients

luwes:
See https://larsenwork.com/easing-gradients/

After:

SCR-20240315-pipv

Before:

SCR-20240315-pira


AdamJaggard (approved):
Love this!

Issue #881, opened 2024-02-29
Feature Request: change values of seek forward/backward with leftarrow or rightarrow


hoangbits:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player

### Description

By default when press left or right arrow the value of seeking forward/backward is 10 seconds.
I tried to change the value for: `forward-seek-offset` or `backward-seek-offset`.
But it won't apply to hotkey. It only apply to button the get hit by mouse on UI

![image](https://github.com/muxinc/elements/assets/7990827/e0db2d57-407a-4771-8f6d-4d83994ea990)

### Expected Behavior

Could you guys add additional attributes to the config seeking value for hotkeys or help showing me where in the source code I could/patch to change the default seeking forward/backward value from hotkey.

hoangbits:
> Thanks for the feedback, @hoangbits. We've got a few things stacked up in the priority queue right now, but this definitely sounds like a good idea! I'm highly confident we'll want to do this at some point and I'll keep you posted on any progress made or if we for whatever reason end up deciding against it.

We might not need additional config variable to handle our case.

I haven't digging into the source code, But if you could share where could I modify default value of 10 seconds to 0.1 seconds. It would be awesome. We could patch your code.

cjpillsbury:
Thanks for the offer, @hoangbits! Happy to help guide you through this, though it is unfortunately/surprisingly a bit trickier than one might assume, since all of the keyboard interactions are defined by a separate library for building media UIs (that we also own/maintain) called [Media Chrome](https://github.com/muxinc/media-chrome). So step one is updating media chrome to allow this to be configurable, likely adding some attribute(s) to use here https://github.com/muxinc/media-chrome/blob/main/src/js/media-controller.js#L449-L467 when present (otherwise stick with the default). I have a meeting with the team today, and I'll be sure to discuss names here. If you're still interested in contributing, I'd start by creating a [new issue](https://github.com/muxinc/media-chrome/issues/new) and labeling it as 'enhancement' on the Media Chrome, with a link to this issue.

cjpillsbury:
Just an update here. While discussing this feature in the media chrome meeting I mentioned, I started writing pseudo-code that ended up being actual code that I figured I might as well just implement, so https://github.com/muxinc/media-chrome/pull/852/files should hopefully be merged and released soon, after which we can add the feature to Mux Player.

cjpillsbury:
[...]

PR #888, opened 2024-03-19
feat(mux-player): Keyboard seek offsets now update with component off…

cjpillsbury:
…set props.


AdamJaggard (approved):
Nice!

cjpillsbury:
resolves #881

PR #891, opened 2024-03-27
fix(playback-core): Make sure we do not apply holdback to seekable when live streams have ended.

cjpillsbury:
In other words, seekable should only apply holdback (baked into `hls.liveSyncPosition`) while the live stream is "active"/"still live".


AdamJaggard (approved):
I think this is the one!

PR #895, opened 2024-04-09
fix(mux-player-react/lazy): display placeholder before JS load

decepulis:
Before suspense loads the player bundle, we use `--mux-player-react-lazy-placeholder:url({placeholder})` to show the placeholder. However, `{placeholder}` wasn't showing in some cases (especially with the new `@mux/blurup` library) because it was an invalid URL. `'{placeholder}'` is more robust, so this PR adds those quotation marks.

### DIFF
--- a/packages/mux-player-react/src/lazy.tsx
+++ b/packages/mux-player-react/src/lazy.tsx
@@ -52,7 +52,7 @@ const Fallback = (props: FallbackProps) => {
         placeholder={placeholder}
         style={
           {
-            '--mux-player-react-lazy-placeholder': placeholder ? `url(${placeholder});` : '',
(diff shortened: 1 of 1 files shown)


luwes (approved):
👍

luwes:
@dar Vercel builds need to be authorized all separately, clicking on Details...

decepulis:
Understood the root cause and updated the initial comment. `{placeholder}` works in some cases (like those generated by @mux/blurhash) but not others (like those generated by @mux/blurup)

decepulis:
@luwes I'm satisfied with this. Don't have write access, though, so I leave merging to you.

Issue #893, opened 2024-04-04
Bug: `@mux/mux-video-react` has type-error with latest version of `@types/react`


ryami333:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-video-react

### Which browsers are you using?

Other (add details below)

### Which operating systems are you using?

macOS

### Description

If you use the latest version of `@mux/mux-video-react` with the latest version of `@types/react` then you will always get a type error stating that `placeholder`, `onPointerEnterCapture` and `onPointerLeaveCapture` are mandatory props.

```
error TS2739: Type '{}' is missing the following properties from type 'Pick<Partial<Props>, "key" | "height" | "playsInline" | "poster" | "width" | "disablePictureInPicture" | "disableRemotePlayback" | "controls" | "controlsList" | ... 262 more ... | keyof MuxMediaPropTypes>': placeholder, onPointerEnterCapture, onPointerLeaveCapture

4   return <MuxVideoPlayer />;
```

### Steps to reproduce

1. Bump to the latest version of both packages. Eg. `yarn upgrade @mux/mux-video-react @types/react --latest`
2. Create a component file which uses `@mux/mux-video-react`, eg:
```tsx
import MuxVideoPlayer from "@mux/mux-video-react";

export function MyComponent() {
  return <MuxVideoPlayer />;
}
```
3. Perform typechecking: `yarn tsc`.

### Current Behavior

There are typescript errors when you omit the `placeholder`, `onPointerEnterCapture` and `onPointerLeaveCapture` props.

### Expected Behavior

There should be no typescript errors when you omit the `placeholder`, `onPointerEnterCapture` and `onPointerLeaveCapture` props.

### Errors

error TS2739: Type '{}' is missing the following properties from type 'Pick<Partial<Props>, "key" | "height" | "playsInline" | "poster" | "width" | "disablePictureInPicture" | "disableRemotePlayback" | "controls" | "controlsList" | ... 262 more ... | keyof MuxMediaPropTypes>': placeholder, onPointerEnterCapture, onPointerLeaveCapture

4   return <MuxVideoPlayer />;

### What version of the package are you using?

v0.8.5

hipdev:
Yep, same issue in my case.

luwes:
this was actually fixed in https://github.com/muxinc/elements/pull/897
double checked with canary

PR #809, opened 2023-11-03
Experiment/mux embed typings

cjpillsbury:
Draft testing planned "opt in" mux-embed published type declarations.

PR #897, opened 2024-04-09
fix: upgrade mux-embed v5.2.0 & media-chrome v3.2.0

luwes:
- [fix: upgrade mux-embed v5.2.0](https://github.com/muxinc/elements/pull/897/commits/87d4b8c569d234a012f7d251ffaf10400a29de5d)
- [fix: upgrade media-chrome & custom-media-element](https://github.com/muxinc/elements/pull/897/commits/46280766dee3f73ecd98a2cd2fb22be60260962a)
- [fix: upgrade React & React types](https://github.com/muxinc/elements/pull/897/commits/02a7b56442d2681208a0fbf36edc115dd3f48a71)

related https://github.com/muxinc/elements/issues/893 but I think it might be related to a `typescript` v5 upgrade.

should we upgrade to TS v5 and fix the broken demo apps caused by the upgrade?


packages/playback-core/src/types.ts
```diff
@@ -1,5 +1,3 @@
-/* eslint @typescript-eslint/triple-slash-reference: "off" */
```
cjpillsbury, line 1: Per mux embed discussion/typing usage, you should be able to use this PR draft (https://github.com/muxinc/elements/pull/809/files#diff-0550a08b3e80a37aa663c63c5e486747f32d7ddea4e7b52eaa818bec25d044b6R1-R2) as a reference point for how to use the officially published typings (though several other things have changed since it was implemented)

cjpillsbury (approved):
blocked by updating the type refs for `mux-embed`, but otherwise this is 🔥. Pre-approving.

PR #900, opened 2024-04-15
feat: add `disable-tracking` attribute and prop

luwes:
https://github.com/muxinc/next-video/issues/231
https://github.com/muxinc/elements/discussions/614

fixes some eslint warnings as well


packages/mux-audio/src/index.ts
```diff
@@ -20,32 +20,19 @@ import { getPlayerVersion } from './env';
...
   METADATA_URL: 'metadata-url',
   PREFER_PLAYBACK: 'prefer-playback',
   BEACON_COLLECTION_DOMAIN: 'beacon-collection-domain',
+  DISABLE_TRACKING: 'disable-tracking',
   DISABLE_COOKIES: 'disable-cookies',
   TYPE: 'type',
   STREAM_TYPE: 'stream-type',
   START_TIME: 'start-time',
-};
+} as const;
```
cjpillsbury, line 35: praise: 👌

cjpillsbury (approved):
LGTM! One callout - I *think* the type changes means the boolean properties will now get a type error when setting to undefined (which is a valid value in our code), so this may break those cases for folks. Just food for thought.

PR #906, opened 2024-04-18
fix(mux-uploader): uploader components state cleanup

cjpillsbury:
* Upgrade upchunk for various fixes that are relied on for mux-uploader improvements
* Account for various state update permutations
* Account for offline
* Use Attribute Selectors + CSS instead of programmatic style updates
* Prefer `toggleAttribute()` for boolean attrs usage.

### DIFF
--- a/packages/mux-uploader/src/mux-uploader.ts
+++ b/packages/mux-uploader/src/mux-uploader.ts
@@ -71,6 +71,10 @@ interface MuxUploaderElement extends HTMLElement {
 }
 
 class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderElement {
+  static get observedAttributes() {
+    return ['no-drop', 'no-progress', 'no-status', 'no-retry', 'pausible', 'max-file-size'];
+  }
+
   protected _endpoint: Endpoint;
 
   constructor() {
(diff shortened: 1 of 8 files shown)


packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -71,6 +71,10 @@ interface MuxUploaderElement extends HTMLElement {
 }
 
 class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderElement {
+  static get observedAttributes() {
+    return ['no-drop', 'no-progress', 'no-status', 'no-retry', 'pausible', 'max-file-size'];
```
cjpillsbury, line 75: whoops! `pausible` is a future effort. Removing for now

packages/mux-uploader/src/mux-uploader.ts
```diff
@@ -257,36 +257,46 @@ class MuxUploaderElement extends globalThis.HTMLElement implements MuxUploaderEl
...
         new CustomEvent('uploadstart', { detail: { file: upload.file, chunkSize: upload.chunkSize } })
       );
+      this.setAttribute('upload-in-progress', '');
+
+      if (upload.offline) {
+        this.dispatchEvent(new CustomEvent('offline'));
+      }
 
-      upload.on('attempt', (event) => {
+      upload.on('attempt', (event: any) => {
```
mmcc, line 269: can't these be typed as the UpChunk events?

mmcc (approved):

packages/mux-uploader/src/mux-uploader.ts
cjpillsbury: they don't have strict types just yet (we were a bit loosey goosey in our original impl, so those didn't make the cut for my initial round of TS tightening on upchunk).

PR #909, opened 2024-04-23
feat(playback-core, mux-video, mux-player): addChapters interface

AdamJaggard:
- Adds a new interface for working with chapters:  `.addChapters()`, `.activeChapter`, `.chapters`.
- Amends cuepoints interface to match chapters and use a shared `CueLike` shape, with a `startTime` and optional `endTime`. The old shape of `{time: number, value: T}` is still supported for backwards compatibility. Because of this both CuePoints and Chapters support sparse (with gaps) or being joined up end to end
- Chapters unsupported on Live content for this iteration, docs will mention this
- Audio UI works as expected

image

image


packages/playback-core/src/text-tracks.ts
```diff
@@ -1,7 +1,16 @@
...
+import { CuePoint, Chapter } from './types';
 import { addEventListenerWithTeardown } from './util';
 
+// Extracts the start time from a cuepoint, considering legacy "time" prop
+const cuePointStart = (cuePoint: CuePoint): number => {
+  if ('time' in cuePoint) {
+    return cuePoint.time;
+  } else {
+    return cuePoint.startTime;
+  }
```
luwes, line 11:
```suggestion
  }
  return cuePoint.startTime;
```

cjpillsbury (approved):
📖 😎

luwes (approved):
LGTM! 🎉

PR #912, opened 2024-04-25
fix: passthrough props to native media els

luwes:
https://github.com/muxinc/next-video/issues/245


cjpillsbury (approved):
should probably follow up with an improvement on the `PropTypes` reliance, but this is 💯 better than status quo.

PR #916, opened 2024-05-02
fix(playback-core): preload=none not respected for playlist stream info fetching with native safari playback.

### LINKED ISSUE #875: Bug: All instanses of mux-video loaded if preload=none

mux-video

Safari, Safari on iOS

iOS

### Description

I have a carousel with videos and noticed the page crushes on Safari iOS/OSX with many videos. On Safari, <mux-video> started to load all metadata for each video, even if preload="none" was added to the mux-video tag. 

On Chrome, there is no problem; the video is loaded only then it appears on the screen. 

### Reduced test case

https://storistashop.myshopify.com/pages/mux-demo

### Steps to reproduce

1. Open Safari on OSX
2. Open https://storistashop.myshopify.com/pages/mux-demo
3. Open developers tools
4. Click on video
5. See a lot requests with .m3u8

On Safari iOS, after a few swipes, the page will be crushed and reloaded.

### Current Behavior

mux-video doesn't respect preload="none" on Safari

### Expected Behavior

mux-video respect preload="none" on Safari

### Errors

_No response_

0.17.2

cjpillsbury:
## Background
In order to support for "media ui extensions" such as "stream type" in our media elements across as many playback environments as possible, we independently fetch and parse HLS playlists, even when playback of HLS media is managed natively by the Safari `HTMLMediaElement`. We also aim to respect developer preferences such as `preload="none"`. To accomplish this, we previously relied on waiting for the `'loadstart'` event to be dispatched before initiating our fetching logic for cases where `preload="none"`. Unfortunately, per spec, `'loadstart'` will be dispatched regardless of `preload` state. Both Chrome and Safari comply with the specification (Firefox appears to deviate from the spec to get closer to the "spirit of the law," but this also results in cross-browser inconsistencies).

For the exact part of the HTML Living Standard that discusses this, see Steps 7 & 8 of _The **resource selection algorithm**_ in [_§4.8.11.5 Loading the media resource_ ](https://html.spec.whatwg.org/multipage/media.html#concept-media-load-algorithm)

To demonstrate this behavior, load [this codesandbox](https://codesandbox.io/p/sandbox/media-loadstart-behavior-h9htzm) in the respective browsers.

## The fix
To accomplish these goals but still attempt to retrieve and compute the playlist+stream info as early as possible, we can instead listen for the first `'play'` event or `'loadedmetadata'` event, whichever comes first. We also need to make sure we actually set the underlying `HTMLMediaElement`'s `preload` state before applying this logic (which we were not previously doing.

fixes #875

### DIFF
--- a/packages/playback-core/src/index.ts
+++ b/packages/playback-core/src/index.ts
@@ -426,11 +426,11 @@ export const initialize = (props: Partial<MuxMediaPropsInternal>, mediaEl: HTMLM
 
   muxMediaState.set(mediaEl as HTMLMediaElement, {});
   const nextHlsInstance = setupHls(props, mediaEl);
+  const setPreload = setupPreload(props as Pick<MuxMediaProps, 'preload' | 'src'>, mediaEl, nextHlsInstance);
   setupMux(props, mediaEl, nextHlsInstance);
   loadMedia(props, mediaEl, nextHlsInstance);
   setupCuePoints(mediaEl);
   const setAutoplay = setupAutoplay(props as Pick<MuxMediaProps, 'autoplay'>, mediaEl, nextHlsInstance);
-  const setPreload = setupPreload(props as Pick<MuxMediaProps, 'preload' | 'src'>, mediaEl, nextHlsInstance);
 
   return {
     engine: nextHlsInstance,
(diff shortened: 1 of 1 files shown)


luwes (approved):
looks good!

PR #930, opened 2024-06-05
fix(mux-player): Hide cast button by default when using DRM.

cjpillsbury:
Since DRM-protected content requires a custom receiver app for casting (https://support.google.com/widevine/answer/6072130?hl=en), hide the cast button by default whenever we encounter a drm token. Went ahead and introduced a simple generic concept (props-driven css) and a pattern (internal css vars as defaults) that we can conform to for these cases.

### DIFF
--- a/packages/mux-player/src/template.ts
+++ b/packages/mux-player/src/template.ts
@@ -8,8 +8,16 @@ import { i18n, stylePropsToString } from './utils';
 import type { MuxTemplateProps } from './types';
 import { StreamTypes, toMuxVideoURL } from '@mux/playback-core';
 
+const getPropsCSS = (props: MuxTemplateProps) => {
+  const { tokens } = props;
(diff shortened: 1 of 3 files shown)


packages/mux-player/src/template.ts
```diff
@@ -8,8 +8,16 @@ import { i18n, stylePropsToString } from './utils';
 import type { MuxTemplateProps } from './types';
 import { StreamTypes, toMuxVideoURL } from '@mux/playback-core';
 
+const getPropsCSS = (props: MuxTemplateProps) => {
+  const { tokens } = props;
+  if (!tokens.drm) return '';
```
cjpillsbury, line 13: obviously if/when we add additional property-specific styles here, we wouldn't want a simple if + early bail. Didn't want to over-architect the guts until we needed it tho.

PR #937, opened 2024-06-12
feat: clean up type attr. Handle dynamic updates better.

cjpillsbury:
Primary purpose of this PR is to account for changes to props/attrs after initial load. Because of the underlying architecture, there are still potential edge cases around state vs. UI mismatches after upload has begun (e.g. pausing state + progress state, though progress will update on the next progress event to handle some of those cases).


luwes (approved):
👍

PR #946, opened 2024-07-12
chore: remove hard coded stream types

AdamJaggard:
Changes only made to examples and readme's. We've already removed them from our main docs examples so just mimicking here. There are some tests with `streamType` still explicitly defined.


luwes (approved):
LGTM if we set the `default-stream-type` but looks like yes

PR #955, opened 2024-07-18
docs(mux-player): correct mdn reference link for playbackRate prop

kenchandev:
When I was browsing the mux-player-react documentation, I noticed that the MDN link referenced for the `playbackRate` prop was incorrect, so I went ahead and corrected it (this issue was also present in the mux-player documentation).


luwes (approved):
thank you, LGTM

PR #957, opened 2024-07-23
fix(playback-core): playready is ending up with a pssh in a pssh.

cjpillsbury:
May be an hls.js bug due to multi-drm.


cjpillsbury:
Closing. We changed our playlist `#EXT-X-KEY` `URI` values to conform with (out of spec) expectations in hls.js, which sidesteps the need for these changes.

Issue #863, opened 2024-01-19
Bug: Captions disappear in React Player on mobile when full screened


AdamJaggard:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player-react

### Which browsers are you using?

Safari on iOS

### Which operating systems are you using?

iOS

### Description

In certain instances of the React Player, full screening makes the captions disappear on mobile.

Example in our docs, the top video:
https://docs.mux.com/guides/mux-player-web
In the wild:
https://www.popflick.com/play/trailer?slug=to-kill-a-mockingbird&category=crime

A different React example from our docs, where the issue doesn't occur, inside one of our sandboxes:
https://docs.mux.com/guides/player-integrate-in-your-webapp#examples

I tried to produce a reduced test case on codesandbox, but this didn't reproduce the issue. Something to do with how it's bundled?

### Steps to reproduce

1. Open one of the example links on mobile (iOS safari confirmed)
2. Play the video
3. When captions are showing, hit fullscreen and they will disappear

### Current Behavior

Captions disappear

### Expected Behavior

Captions don't disappear

### What version of the package are you using?

v2.3.1

gkobilansky:
I can confirm we found this bug in our app after switching to mux-player. I was able to work around it by using the minimal theme - https://docs.mux.com/guides/player-themes

PR #958, opened 2024-07-25
fix(mux-player): use CSS to disable subtitle shifting for iOS in fullscreen.

cjpillsbury:
Blocked by https://github.com/muxinc/media-chrome/pull/946
Fixes https://github.com/muxinc/elements/issues/863

Smoke tested 👍  on both iOS Safari and iOS Chrome, plus desktop browsers for regression testing.

### DIFF
--- a/packages/mux-player/src/themes/gerwig/gerwig.html
+++ b/packages/mux-player/src/themes/gerwig/gerwig.html
@@ -66,13 +66,25 @@
      * 0.433s is the transition duration for VTT Regions.
      * Borrowed here, so the captions don't move too fast.
      */
-    media-controller ::slotted([slot='media']) {
+    media-controller {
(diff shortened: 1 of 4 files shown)


packages/mux-player/src/themes/gerwig/gerwig.html
```diff
@@ -66,15 +66,27 @@
      * 0.433s is the transition duration for VTT Regions.
      * Borrowed here, so the captions don't move too fast.
      */
-    media-controller ::slotted([slot='media']) {
```
cjpillsbury, line 69: NOTE: Although this change isn't technically required, custom CSS properties (aka "CSS variables") inherit, so removing the narrower selection since it's both unnecessary and adds specificity that is harder to override.

luwes (approved):
nice! 👍

PR #961, opened 2024-07-26
fix: try out TS Media Chrome

luwes:
here to test the player build and examples with TS Media Chrome pre v4


cjpillsbury:
Closing this as it validated TS upgrade. #978 will be the official Media Chrome (et al.) upgrade PR.

PR #965, opened 2024-07-29
feat(playback-core): error migration et al.

cjpillsbury:
A bunch of changes, motivated by but not exhausted by getting granular mux data drm errors, including:
- pushing all of the core error translation down to playback-core
    -  which meant updating build setup/tooling to account for i18n json languages
- introducing new concepts for mux error code + category that can be used as well-defined values for different text/presentation cases (e.g. error messages and context but also error dialog title/text)
   - NOTE: currently not using the more granular error codes due to considerations with mux data + customer filters, etc. though we may want to revisit this later.
- added `tokens` to other elements components. This wasn't obviously valuable before DRM, but now it allows for more consistent data structures across different elements "layers"
    - This included unsurfaced edge case bugs with URL <--> playback id translations that didn't properly account for our extended playback-id syntax (i.e. `"PLAYBACK_ID?query=param1&query2=param2"`
- add + update tests for errors
- add a bunch of code to capture and translate error cases, primarily DRM, but also a few unaccounted for cases for video playback


packages/mux-video-react/src/index.tsx
```diff
@@ -105,7 +107,6 @@ MuxVideo.propTypes = {
   startTime: PropTypes.number,
   streamType: PropTypes.oneOf(Object.values(StreamTypes)),
   targetLiveWindow: PropTypes.number,
-  tokens: PropTypes.any,
```
cjpillsbury, line 108: NOTE: `<mux-player>` and `<MuxPlayer>` have "tokens". `<mux-video>` now has `playbackToken` (as will `<mux-audio>`), in part for things like moving our more precise error logic into playback-core (potentially part of this PR effort). `toMuxVideoSrc()` util fn expects `tokens`, which is used for translating props to `src` URL. `<mux-video>` and `<mux-audio>` also relies on `playbackId` with extra query params (including `?token={PLAYBACK_TOKEN}`) to construct `src` URL... in other words, we should try to unwind some of this...

packages/playback-core/src/index.ts
```diff
@@ -582,13 +613,14 @@ export const getStreamTypeConfig = (streamType?: ValueOf<StreamTypes>) => {
...
+  props: Partial<Pick<MuxMediaPropsInternal, 'src' | 'playbackId' | 'drmToken' | 'customDomain' | 'drmTypeCb'>>
 ): Partial<HlsConfig> => {
   const {
     drmToken,
-    src,
-    playbackId = toPlaybackIdFromSrc(src), // Since Mux Player typically sets `src` instead of `playbackId`, fall back to it here (CJP)
+    playbackId: playbackIdWithOptionalParams, // Since Mux Player typically sets `src` instead of `playbackId`, fall back to it here (CJP)
+    drmTypeCb,
   } = props;
+  const [playbackId] = playbackIdWithOptionalParams ? toPlaybackIdParts(playbackIdWithOptionalParams) : [];
```
cjpillsbury, line 623: NOTE: Need to account for the potential additional params in playback-core because of our `<mux-video>` and `<mux-audio>` implementation of `playbackId`...

packages/playback-core/src/errors.ts
```diff
@@ -17,6 +29,8 @@ export class MediaError extends Error {
 
   name: string;
   code: number;
+  public muxCode?: number;
```
cjpillsbury, line 32: NOTE: plausibly want to drop `errorCategory` and `muxCode`, but keeping for now just in case, since error context, message, and code are information lossy.

packages/playback-core/src/index.ts
```diff
@@ -626,29 +658,41 @@ export const getDRMConfig = (
...
+        return value;
+      });
     },
   };
 };
 
 export const getAppCertificate = async (appCertificateUrl: string) => {
   const resp = await fetch(appCertificateUrl);
+  if (resp.status !== 200) {
+    return Promise.reject(resp);
```
cjpillsbury, line 673: NOTE: we need some way to error with enough info to introspect why the request failed. Currently simply rejecting with the non-200 `Response`, but we can definitely consider alternatives. Since this function scope doesn't have sufficient information to fully validate information (e.g. a mismatched `playbackId` + JWT `sub`), we can't do a full translation here without passing in additional context.

packages/playback-core/src/request-errors.ts
```diff
@@ -0,0 +1,207 @@
...
+        mediaError.muxCode = MuxErrorCode.TOKEN_MALFORMED;
+        return mediaError;
+      }
+      if (isJWTSubMismatch(jwtObj, playbackId)) {
+        const message = i18n(
+          `The video’s playback ID does not match the one encoded in the ${tokenNamePrefix}-token.`,
+          translate
+        );
+        const context = i18n(
+          `Specified playback ID: {playbackId} and the playback ID encoded in the ${tokenNamePrefix}-token: {tokenPlaybackId}`,
```
cjpillsbury, line 131: NOTE: Need to rework this and other strings to parameterize based on the request type/"category" (e.g. `drm-token` vs. `playback-token`).

packages/playback-core/src/request-errors.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  if (status === 403 || status === 400) {
+    if (jwtObj) {
+      if (isJWTExpired(jwtObj, requestTime)) {
+        const dateOptions: any = {
+          timeStyle: 'medium',
+          dateStyle: 'medium',
+        };
+        const message = i18n(`The video’s secured ${tokenNamePrefix}-token has expired.`, translate);
+        /** @TODO move lang build crud from mux-player to playback-core (See: esbuilder.js) (CJP) */
```
cjpillsbury, line 113: NOTE: Callout that this will need to happen as well.

packages/playback-core/src/request-errors.ts
```diff
@@ -0,0 +1,207 @@
...
+       * @TODO invalid playback URLs and playback IDs are 404 for video URL responses but 400 for DRM. Should change this for consistency (CJP)
+       */
+
+      /** @TODO omitted tokens are 403 for video URL responses but 400 for DRM. Should change this for consistency (CJP) */
+      // NOTE: This *should* not happen for DRM, since the drm token
+      // is currently used to detect whether or not DRM should
+      // be setup at all. Including for exhaustiveness. (CJP)
+    } else {
+      const message = i18n(
+        `${status ?? 403} error trying to access this ${category} URL. If this is a signed URL, you might need to provide a ${category} token.`,
```
cjpillsbury, line 179: NOTE: These HTTP statuses should be normalized in prod soon, but we should still potentially consider updating our default message to not include `403` and instead replace it with something like `"Authorization"`

packages/mux-video-react/src/index.tsx
cjpillsbury: Re-added as a feature for better consistency.

packages/playback-core/package.json
```diff
@@ -63,6 +63,7 @@
     "mux-embed": "~5.2.0"
   },
   "devDependencies": {
+    "@mux/esbuilder": "0.1.0",
```
cjpillsbury, line 66: NOTE: Using (an updated) esbuilder for playback-core as well to account for language json files.

packages/playback-core/src/errors.ts
```diff
@@ -1,9 +1,53 @@
+// Identifies what kind of request was made that resulted in an error
+export const MuxErrorCategory = {
+  VIDEO: 'video',
+  THUMBNAIL: 'thumbnail',
+  STORYBOARD: 'storyboard',
+  DRM: 'drm',
+} as const;
+
+export const MuxErrorCode = {
+  NOT_AN_ERROR: 0,
```
cjpillsbury, line 10: NOTE: I implemented unique codes for unique errors for elements, which we can use at various "levels" to translate into additional info (namely, the error dialog for mux player). With the new architecture, we rely on the newly added error category (i.e. was it e.g. DRM or Video, defined above) + error code (i.e. was it a network error or an encrypted error) to provide human readable info (i.e. "Your drm-token was expired")

packages/playback-core/src/request-errors.ts
```diff
@@ -0,0 +1,255 @@
+import type { LoaderResponse } from 'hls.js';
```
cjpillsbury, line 1: NOTE: We may want to clean up our organization of error modules. I'm definitely amenable to relocating this, consolidating, etc.

scripts/esbuilder/esbuilder.js
```diff
@@ -9,9 +9,30 @@ const camelCase = (name) => {
 };
 
 const args = process.argv.slice(3).reduce((processArgs, val) => {
-  let [key, value] = val.split('=');
+  let [key, value = true] = val.split('=');
```
cjpillsbury, line 12: NOTE: Needed to update esbuilder to account for multiple, arg-provided `--external` arguments.

scripts/esbuilder/esbuilder.js
```diff
@@ -56,8 +77,10 @@ const options = {
   bundle: true,
   target: 'es2019',
   minify: args.minify,
+  sourcemap: args.sourcemap ?? false,
```
cjpillsbury, line 80: Added to account for sourcemap usage in playback-core.

packages/mux-player/src/errors.ts
```diff
@@ -1,166 +1,182 @@
...
-  let dialog: DialogOptions = {};
-  let devlog: DevlogOptions = {};
+const capitalizeFirstLetter = (string: string) => string.charAt(0).toUpperCase() + string.slice(1);
+// NOTE: Since playback-core and mux-video have no devlog or dialog, these translations should still live in Mux Player (at least for now).
 
-  switch (error.code) {
-    case MediaError.MEDIA_ERR_NETWORK: {
-      dialog.title = i18n(`Network Error`, translate);
-      dialog.message = error.message;
+const muxMediaErrorToDialogTitle = (mediaError: MediaError, translate = false) => {
```
cjpillsbury, line 8: NOTE: For the error dialog text, I broke each text section (i.e. title + message) into separate functions that translate from code + category into the corresponding text. If this feels a bit too unwieldy, we can refactor to just have a single lookup function for both, though this does mean we'll have redundant messages that we'll need to keep in sync (particularly for titles).

packages/playback-core/src/index.ts
```diff
@@ -667,22 +674,28 @@ export const getDRMConfig = (
...
+  }
+  const keyBuffer = await resp.arrayBuffer();
   return new Uint8Array(keyBuffer);
 };
 
 export const setupNativeFairplayDRM = (
-  props: Partial<Pick<MuxMediaPropsInternal, 'playbackId' | 'drmToken' | 'customDomain' | 'drmTypeCb'>>,
+  props: Partial<Pick<MuxMediaPropsInternal, 'playbackId' | 'tokens' | 'playbackToken' | 'customDomain' | 'drmTypeCb'>>,
```
cjpillsbury, line 698: NOTE: There's a lot of very similar code for error tracking and translation, both here and with the corresponding hls.js function (below). I couldn't think of any straightforward way to abstract this without even larger refactors, but if anything jumps out, even if minor improvements, I'm definitely amenable.

packages/playback-core/src/index.ts
```diff
@@ -1143,10 +1315,44 @@ async function handleNativeError(event: Event) {
...
+      // ever get metadata loaded in the first place. (CJP)
+    }, 500);
+    return;
+  }
+
   if (mediaEl.src && (code !== MediaError.MEDIA_ERR_DECODE || code !== undefined)) {
     // Attempt to get the response code from the video src url.
     try {
-      const { status } = await fetch(mediaEl.src as RequestInfo);
+      const { status } = await fetch(mediaEl.src);
```
cjpillsbury, line 1355: NOTE: I **_think_** this should no longer be necessary, since we now fetch playlists (to e.g. infer stream type) in parallel, even for native playback. Those independent requests should also now translate to errors on non-200 response cases.

packages/mux-audio/src/index.ts
```diff
@@ -220,6 +229,63 @@ class MuxAudioElement extends CustomAudioElement implements Partial<MuxMediaProp
     }
   }
 
+  get customDomain() {
```
cjpillsbury, line 232: NOTE: This was missing from mux-audio.

packages/mux-audio/src/index.ts
```diff
@@ -220,6 +229,63 @@ class MuxAudioElement extends CustomAudioElement implements Partial<MuxMediaProp
...
+      this.setAttribute(Attributes.CUSTOM_DOMAIN, val);
+    } else {
+      this.removeAttribute(Attributes.CUSTOM_DOMAIN);
+    }
+  }
+
+  /**
+   * Get the playback token for signing the src URL.
+   */
+  get playbackToken() {
```
cjpillsbury, line 250: NOTE: This was missing from mux-audio. Don't need DRM token because we (at least currently) don't support DRM-protected audio content.

packages/mux-player/src/errors.ts
```diff
@@ -1,166 +1,182 @@
...
+      return i18n(`{category} URL is formatted incorrectly`, translate).format({ category });
+    }
+    if (
+      [
+        MuxErrorCode.NETWORK_TOKEN_AUD_MISMATCH,
+        MuxErrorCode.NETWORK_TOKEN_AUD_MISSING,
+        // @ts-ignore
+      ].includes(mediaError.muxCode)
+    ) {
+      return i18n(`The {tokenNamePrefix}-token is formatted with incorrect information.`, translate).format({
```
andrewjl-mux, line 82: Nit: wonder if more specificity would help make *how* to fix whatever is wrong with the token easier. In this case calling out that it's the `aud` claim at issue.

Likewise for the other cases where something more specific can be inferred from the `MuxErrorCode`.

andrewjl-mux: Also assuming this is user-facing (please correct me if I am wrong) given `muxMediaErrorToDialogMessage`

packages/playback-core/src/request-errors.ts
```diff
@@ -0,0 +1,255 @@
...
+          tokenPlaybackId: jwtObj.sub,
+        });
+        const mediaError = new MediaError(message, mediaErrorCode, true, context);
+        mediaError.errorCategory = category;
+        mediaError.muxCode = MuxErrorCode.NETWORK_TOKEN_SUB_MISMATCH;
+        mediaError.data = resp;
+        return mediaError;
+      }
+      if (isJWTAudMissing(jwtObj, expectedAud)) {
+        const message = i18n(`The {tokenNamePrefix}-token is formatted with incorrect information.`, translate).format({
```
andrewjl-mux, line 142: Nit: does it make sense to add more specificity here? (Relying on `isJWTAudMissing` to infer the `aud` claim is at issue here.)

Strawman message (rough wording): `The {tokenNamePrefix}-token has a missing or incorrect audience claim.`

andrewjl-mux: Ah okay. I see that detail is in the error context.

packages/mux-player/src/errors.ts
cjpillsbury: Yeah this is user-facing. Our assumption is the user will not be the developer (at least in the vast majority of cases). We've aimed for treating users as different "audiences" than devs with language appropriate to each. In the case of devs, they will have the errors themselves, the dev logs, and Mux Data. In the case of users, they'll have the error dialog UI. The goal (perhaps sometimes falling short) is to make each presentation of an error as actionable as possible for each audience. For the devs, that typically means describing as best as possible "root cause." For the end user, that will very much depend on what kind of actions they can('t) take. In this particular case, a more descriptive error message would be something like:

> It looks like there might be some bad code for loading playback: The {tokenNamePrefix}-token is formatted with incorrect information. You may need to raise an issue with the relevant folks to resolve.

andrewjl-mux: Right, that makes sense.

packages/mux-player/src/index.ts
```diff
@@ -437,6 +437,7 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
       let { detail: error }: { detail: any } = event as CustomEvent;
 
       if (!(error instanceof MediaError)) {
+        console.error('!!!!!!non MediaError error!', error);
```
luwes, line 440: should this stay in?

packages/mux-player/src/index.ts
```diff
@@ -632,42 +628,64 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
...
           }
         }
         break;
       }
       case PlayerAttributes.STORYBOARD_TOKEN: {
         if (newValue) {
-          const { aud } = parseJwt(newValue);
-          if (aud !== 's') {
-            logger.warn(
-              i18n(`The provided storyboard-token should have audience value 'd' instead of '{aud}'.`).format({ aud })
```
luwes, line 658: 😅 'd' ouch

luwes (approved):
code lgtm!

packages/mux-player/src/index.ts
cjpillsbury: Nope! 😅  Thanks

PR #975, opened 2024-08-07
fix: track src being removed bug

luwes:
I ran into this issue where the thumbnails track src was being removed because of a race condition.

the `forceHiddenThumbnails` was called twice before the setTimeout callback so it ended up setting an empty src.
let me know if that doesn't make sense.


luwes:
pretty confused because I'm not reproducing it anymore on player.style. gonna close it until it pops up again

PR #977, opened 2024-08-21
fix: chapters and cuepoints interface doesnt reflect internal types

### LINKED ISSUE #947: Bug: Mismatch between docs, TS types, and behavior for cuepoints and chapters

mux-player-react

Safari

macOS

### Description

The types described in the [docs](https://docs.mux.com/guides/player-advanced-usage#add-chapters-and-time-based-metadata) don't match up with the types in TypeScript for `addCuePoints` and `addChapters` in `@mux/mux-player-react`

### Reduced test case

https://codesandbox.io/p/sandbox/dreamy-pascal-637nl9?file=%2Fsrc%2FApp.tsx%3A9%2C45

#### Cue points
- Try passing an object with type `{ startTime: number; endTime?: number; value: any; }` to `addCuePoints`. TypeScript yells (even though this type is in the Mux Docs).
- Try passing an object with type `{ time: number; value: any }` to `addCuePoints`. TypeScript does not yell (even though this type is not in the Mux Docs)

#### Chapters
- Try passing an object with type `{ startTime: number; endTime?: number; value: any; }` to `addChapters`. TypeScript yells (even though this type is in the Mux Docs).
- Try passing an object with type `{ startTime: number; endTime: number; value: any; }` to `addChapters`. TypeScript does not yell.

#### Cue points
- Docs say `{ startTime: number; endTime?: number; value: any; }`
- TS types say `addCuePoints<T = any>(cuePoints: { time: number; value: T; }[]): Promise<TextTrack | undefined>;`
- I'm using `{ time: number; value: any }` and it's working as expected

#### Chapters
- Docs say `{ startTime: number; endTime?: number; value: any; }`
- TS types say `addChapters(chapters: { startTime: number; endTime: number; value: string; }[]): Promise<TextTrack> | undefined;`. (Notice that `endTime` is required in TS)
- I'm using `{ startTime: number; value: string }` and it's working as expected

### Expected Behavior

I'm assuming the docs are correct here. So...
#### Cue Points
- TS should say `addCuePoints<T = any>(cuePoints: { startTime: number; endTime?: number; value: T; }[]): Promise<TextTrack | undefined>;`
- And maybe `{ time: number; value: T }`  should be supported as a legacy option, since that's how Cue Points used to work, right?

#### Chapters
- TS types should say `addChapters(chapters: { startTime: number; endTime?: number; value: string; }[]): Promise<TextTrack> | undefined;`

### Errors

image

image

2.8.0

AdamJaggard:
When we released chapters we unified the chapter and cuepoint types under the hood, making both support an optional `endTime` with cuepoints still supporting the legacy shape using `time` instead of `startTime`. Our `addChapters` and `addCuePoints` methods on the player don't use these new types though. `endTime` is currently required by the `addChapters` method, but shouldn't be, and only the legacy type is required by `addCuePoints`.

These changes re-use our existing internal types for `Chapter` and `CuePoint` for the external interface.

Fixes #947

### DIFF
--- a/examples/nextjs-with-typescript/pages/MuxPlayerChapters.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayerChapters.tsx
@@ -16,7 +16,7 @@ function MuxPlayerPage() {
   const [activeChapter, setActiveChapter] = useState<Chapter>(undefined);
 
   function addChaptersToPlayer(playerEl: MuxPlayerElement) {
-	playerEl.addChapters(chapters);
+	  playerEl.addChapters(chapters);
   }
 
   return (
--- a/examples/nextjs-with-typescript/pages/MuxPlayerChapters.tsx
+++ b/examples/nextjs-with-typescript/pages/MuxPlayerChapters.tsx
@@ -16,7 +16,7 @@ function MuxPlayerPage() {
   const [activeChapter, setActiveChapter] = useState<Chapter>(undefined);
 
   function addChaptersToPlayer(playerEl: MuxPlayerElement) {
-	  playerEl.addChapters(chapters);
+    playerEl.addChapters(chapters);
   }
 
   return (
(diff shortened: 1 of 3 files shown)


cjpillsbury (approved):
Clear. Simple. Better. 😎.

PR #979, opened 2024-08-26
fix(playback-core): Use userAgentData API to improve android-like platform detection for preferPlayback MSE cases.

cjpillsbury:


luwes (approved):
👍

cjpillsbury:
Was waiting to hear back from customer, but haven't yet. Merging anyway, as this should improve our default prefer MSE playback conditions.

PR #982, opened 2024-09-13
fix(mux-player, mux-player-react): Remove seek buttons from mobile pre-playback UI

AdamJaggard:
They have no use during pre-playback, they're an artefact of the mobile designs that we overlooked.


luwes (approved):
👍

PR #987, opened 2024-09-17
tests: fix tests by using Chrome --headless=old

luwes:
locally this seemed to do the trick. issues caused by the new headless mode I suppose


cjpillsbury (approved):
😎 good sleuthin'.

PR #992, opened 2024-09-20
chore(mux-player,mux-player-react)!: major version update

AdamJaggard:
Wanted to force a major version bump and noticed these ts-ignore seemed to no longer be needed anyway.


cjpillsbury (approved):
👍

PR #997, opened 2024-10-03
fix(playback-core): Handle cases where subs or cc media playlist is default (based on hls.js assumptions).

cjpillsbury:
fixes https://github.com/muxinc/player-and-sdks-team/issues/168

### DIFF
--- a/packages/playback-core/src/text-tracks.ts
+++ b/packages/playback-core/src/text-tracks.ts
@@ -25,13 +25,12 @@ export function setupTextTracks(
         return lang == baseTrackObj?.lang && name === trackObj.label && type.toLowerCase() === trackObj.kind;
       });
 
-      addTextTrack(
-        mediaEl,
-        trackObj.kind as TextTrackKind,
-        trackObj.label,
-        baseTrackObj?.lang,
-        `${trackObj.kind}${idx}`
-      );
+      // NOTE: Undocumented method for determining identifier by hls.js. Relied on for
+      // ensuring CUES_PARSED events can identify and apply cues to the appropriate track (CJP).
+      // See: https://github.com/video-dev/hls.js/blob/master/src/controller/timeline-controller.ts#L640
+      const id = trackObj._id ?? trackObj.default ? 'default' : `${trackObj.kind}${idx}`;
+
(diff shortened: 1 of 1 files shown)


packages/playback-core/src/text-tracks.ts
```diff
@@ -42,17 +41,29 @@ export function setupTextTracks(
...
+          (showingTrack.id === 'default' && defaultTrack) ||
+          (lang == showingTrack.language && name === showingTrack.label && type.toLowerCase() === showingTrack.kind)
+        );
       });
       // After the subtitleTrack is set here, hls.js will load the playlist and CUES_PARSED events will be fired below.
       hls.subtitleTrack = idx;
     }
 
-    if (showingTrack && showingTrack?.id === hlsTrackId) {
+    if (showingTrack?.id === hlsTrackId) {
```
luwes, line 66: what if undefined === undefined ?
that showingTrack null check was needed I think

cjpillsbury: we early bail now

packages/playback-core/src/text-tracks.ts
```diff
@@ -42,17 +41,29 @@ export function setupTextTracks(
       return textTrack.id && textTrack.mode === 'showing' && ['subtitles', 'captions'].includes(textTrack.kind);
     });
 
+    if (!showingTrack) return;
```
cjpillsbury, line 44: cc @luwes (per comment, below) - added this to simplify

packages/playback-core/src/text-tracks.ts
```diff
@@ -137,6 +149,9 @@ export function addTextTrack(
   if (id) {
     trackEl.id = id;
   }
+  if (!!defaultTrack) {
+    trackEl.default = true;
+  }
```
luwes, line 154:
```suggestion
  trackEl.default = !!defaultTrack
```

luwes (approved):
LGTM

packages/playback-core/src/text-tracks.ts
cjpillsbury: I'm intentionally only setting `default` if `true` here (due to **_default_** value of `default`), but I get the sentiment.

PR #998, opened 2024-10-03
fix: chapters disappearing after preload none

luwes:
fixes a bug when starting playback after preload none the chapters disappear

### DIFF
--- a/packages/playback-core/src/text-tracks.ts
+++ b/packages/playback-core/src/text-tracks.ts
@@ -108,7 +108,7 @@ export function setupTextTracks(
     // Keeping this a forEach in case we want to expand the scope of this.
     Array.from(mediaEl.textTracks).forEach((track) => {
       if (['subtitles', 'caption'].includes(track.kind)) return;
-      if (track.label !== 'thumbnails' && track.kind !== 'chapters') return;
+      if (!(track.label === 'thumbnails' || track.kind === 'chapters')) return;
       if (!track.cues?.length) {
         let selector = 'track';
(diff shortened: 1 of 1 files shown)


packages/playback-core/src/text-tracks.ts
```diff
@@ -93,13 +93,16 @@ export function setupTextTracks(
     });
   });
 
-  const forceHiddenThumbnails = () => {
+  const forceHiddenTracks = () => {
     // Keeping this a forEach in case we want to expand the scope of this.
     Array.from(mediaEl.textTracks).forEach((track) => {
       if (['subtitles', 'caption'].includes(track.kind)) return;
-      if (track.label !== 'thumbnails') return;
+      if (track.label !== 'thumbnails' && track.kind !== 'chapters') return;
```
cjpillsbury, line 100: suggestion(non-blocking): these are logically equivalent (see [De Morgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)) but this feels clearer/more expressive of the semantics here imo (take it or leave it)
```suggestion
      if (!(track.label === 'thumbnails' || track.kind === 'chapters')) return;
```

cjpillsbury (approved):
one very minor nb suggestion. Eventually we should probably make this more generic to simply include all non-subtitles/captions tracks to avoid further iterations.

packages/playback-core/src/text-tracks.ts
luwes: I considered this :) sure, let's do the more readable

PR #1004, opened 2024-10-21
fix(playback-core): Make sure seekable TimeRanges is populated when updating state.

cjpillsbury:
Suspect this is the root cause of (non-fatal) error logs showing up in test runs (including CI).


luwes (approved):
👍

PR #1005, opened 2024-10-23
Update React peer dependencies to include RCs

heff:
Might be overkill and I have not tested this. It was throwing this error on a fresh next.js project with react at an RC version.
image


luwes (approved):
LGTM!

https://semver.npmjs.com/
I think it works but it only lists the prereleases in the test page if you use the `^19 || ^19.0.0-0` separately. not sure why

Issue #1012, opened 2024-11-13
add warning event for non-fatal errors


luwes:
remove non-fatal warnings from the error event and dispatch them on a new warning event

PR #1014, opened 2024-11-15
feat: use MC media-error-dialog

luwes:
related #1012
requires https://github.com/muxinc/media-chrome/pull/1024

- fixes an issue where the `.error` object could be null when the error event is dispatched.
this was introduced when we started calling initialize() function deferred in mux-video I believe.
this is fixed by saving the error state each time before we dispatch the error event.
- the warning API is not implemented because after closer inspection we don't really do much with non-fatal errors.
  the only thing is doing a console.warn here:
  https://github.com/muxinc/elements/blob/1e1f24a18d7fe61c5b4bdfd9cf4cfb6d9d57b0ea/packages/mux-player/src/index.ts#L461-L465
we can easily do that in playback-core if we want?


packages/mux-player/src/index.ts
```diff
@@ -179,6 +179,34 @@ function getProps(el: MuxPlayerElement, state?: any): MuxTemplateProps {
   return props;
 }
 
+MediaErrorDialog.formatErrorMessage = (error: { code?: number; message?: string }) => {
```
cjpillsbury, line 182: question: any tradeoffs you see on subclassing `MediaErrorDialog` with a `formatErrorMessage` vs. modifying `MediaErrorDialog`'s `formatErrorMessage`? One pro of this approach: we automatically get granular error messaging for all themes. One con of this approach: on the off chance that folks have `<mux-player>` and some other player on the same page, folks might get unexpected results. I suspect any perf etc. considerations are negligible (only a small amount of addititonal code required for subclassing). Any others on either side you can think of?

packages/mux-player/src/index.ts
```diff
@@ -458,7 +493,7 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
         return;
       }
 
-      const { dialog, devlog } = getErrorLogs(error, false);
+      const { devlog } = getErrorLogs(error, false);
```
cjpillsbury, line 496: nitpick: the `dialog` and `devlog` code are already broken up into separate functions in the `errors.ts` module. Since their usage is no longer colocated, we should just export and use those directly (feel free to rename given their current verbosity).

cjpillsbury (changes requested):
Some questions/callouts on the draft

packages/mux-player/src/index.ts
```diff
@@ -288,10 +315,18 @@ class MuxPlayerElement extends VideoApiElement implements MuxPlayerElement {
   #hotkeys = new AttributeTokenList(this, 'hotkeys');
   #state: Partial<MuxTemplateProps> = {
     ...initialState,
-    onCloseErrorDialog: () => this.#setState({ dialog: undefined, isDialogOpen: false }),
-    onInitFocusDialog: (e) => {
+    onCloseErrorDialog: (event) => {
```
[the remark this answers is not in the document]
luwes: it is possible but we use the `isDialogOpen` state to change 2 attributes in the theme; `disabled` and `nohotkeys`.

enabling `nohotkeys` seems an easy one to handle in Media Chrome only if the dialog is open.

setting `disabled` on all controls directly when the dialog opens feels forceful.
thoughts?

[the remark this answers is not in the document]
luwes: resolve: we disable all state change requests when a fatal error occurs.

adding `disabled` or `inert` to UI controls to disable keyboard focus is for now left up to the consumer.

packages/mux-player/src/index.ts
luwes: we put a pin in this. we will go forward with this static override for now but maybe this could be avoided by having the end user message come from the custom media element directly?

cjpillsbury: ah we (or at least I) actually said we would be mostly fine with this approach with one small addendum - in the `else` case, we would use the base impl of `formatErrorMessage()`. Something like:

```ts
const baseFormatErrorMessage = MediaErrorDialog.formatErrorMessage;
MediaErrorDialog.formatErrorMessage = (error: { code?: number; message?: string }) => {
  if (error instanceof MediaError) {
    // ...
  }
  return baseFormatErrorMessage(error);
};
```

packages/mux-player/src/index.ts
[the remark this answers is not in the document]
cjpillsbury: Fine for now but I think we should be able to remove most/all of this as a followup either here and/or in media chrome

packages/mux-player/src/themes/classic/classic.html
```diff
@@ -521,6 +521,7 @@
     <slot name="media" slot="media"></slot>
     <slot name="poster" slot="poster"></slot>
     <media-loading-indicator slot="centered-chrome" noautohide></media-loading-indicator>
+    <media-error-dialog noautohide></media-error-dialog>
```
cjpillsbury, line 524: nitpick(blocking): `slot="dialog"`

packages/mux-player/tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "esModuleInterop": true,
     "skipLibCheck": false,
     "forceConsistentCasingInFileNames": true,
-    "typeRoots": ["node_modules/@types", "../../types/media-chrome.d.ts"]
```
cjpillsbury, line 21: question(blocking): Why were these types things changed for this PR? This might be right (I'd want to make sure for all the different use cases we have), but unless this is needed for the PR, these changes should be backed out and added to a separate PR.

packages/playback-core/package.json
```diff
@@ -51,7 +51,7 @@
     "build:esm-module": "esbuilder src/index.ts --sourcemap --format=esm --outfile=./dist/playback-core.mjs",
     "build:cjs": "esbuilder src/index.ts --sourcemap --format=cjs --out-extension:.js=.cjs.js --external:mux-embed --external:hls.js",
     "build:iife": "esbuilder src/index.ts --sourcemap --format=iife --outfile=./dist/playback-core.js",
-    "prebuild:types": "shx mkdir -p ./dist/types && shx cp ../../types/** ./dist/types/",
```
cjpillsbury, line 54: Same as question, above.

cjpillsbury (changes requested):
some small callouts and looks like there are some linting errors, but this is def close.

packages/mux-player/src/index.ts
luwes: yes, good catch. I somehow thought it already did that.

packages/mux-player/tsconfig.json
luwes: this was needed to make it build, because the MediaErrorDialog class is imported.

the media-chrome.d.ts file was excluding the types that are now available in the Media Chrome pkg itself

```ts
declare module 'media-chrome' {
  export class MediaTheme extends HTMLElement {}
  export class MediaController extends HTMLElement {
    media: HTMLVideoElement;
  }
}
```

luwes:
test page loading a 404 source 
https://elements-demo-vanilla-git-fork-luwes-media-error-dialog-mux.vercel.app/mux-player-error.html

PR #1017, opened 2024-11-19
fix(mux-video, mux-video-react, mux-audio, mux-audio-react, mux-player, mux-player-react): Expose element name and version via exports and statics for web components.

cjpillsbury:


luwes (approved):
lgtm

Issue #1015, opened 2024-11-18
Bug: mux-react-player/lazy "loop" attribute only works if string, not boolean (Next 15)


arcataroger:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player-react

### Which browsers are you using?

Chrome

### Which operating systems are you using?

macOS

### Description

This is a strange issue...

Using  `<MuxPlayer playbackId="abcde" loop={true}/>` works fine in most environments. However, in Next 15, when using `mux-player-react/lazy`, that does NOT work. It DOES work if you pass `"true"` as a string instead of a boolean.

## Not working:

- Next 15 with lazy: https://codesandbox.io/p/github/arcataroger/next-mux-video-looping/main?file=%2Fapp%2Fpage.tsx

## Working as intended:
- Next 15 without lazy: (same page as above, bottom-most example)
- Next 14: https://codesandbox.io/p/devbox/mux-looping-demo-next-14-t8tkf8?file=%2Fapp%2Fpage.tsx%3A6%2C18
- React without Next: https://codesandbox.io/p/sandbox/mux-looping-demo-react-v7rqv8?file=%2Fsrc%2FApp.tsx%3A7%2C29

### Reduced test case

https://codesandbox.io/p/github/arcataroger/next-mux-video-looping/main?file=%2Fapp%2Fpage.tsx

### Steps to reproduce

1. Spin up a Next 15 demo with `create-next-app`
2. Add `@mux/mux-player-react` as a dependency
3. Import `@mux/mux-player-react/lazy` as a component
4. Use that component with `loop={true}` as a boolean. It does not loop. It DOES loop if `"true"` is passed as a string.

### Current Behavior

Passing `loop={true}` as a boolean does nothing.

Passing `loop="true"` as a string DOES loop, but causes a TypeScript error.

### Expected Behavior

Passing `loop={true}` as a boolean should loop the video.

### What version of the package are you using?

3.1.0

Walkeezy:
Experiencing the same issue

luwes:
thanks for reporting this issue!

I had a look and I believe this is caused by React 19 support for custom elements with the combination of Mux player's props to attributes conversion. The React wrapper converts props to attributes for React 18 and under because it did not support properties on custom elements.

for some reason for the lazy component the loop property is set to the value of an empty string, which is true for a boolean attribute, but not for a property and thus it is made false.

[...]

PR #1026, opened 2024-12-06
fix: React 19 bool prop to attr value bug

luwes:
fix https://github.com/muxinc/elements/issues/1015


packages/mux-player-react/src/common/utils.ts
```diff
@@ -11,6 +22,7 @@ const ReactPropToAttrNameMap = {
   playsInline: 'playsinline',
   autoPlay: 'autoplay',
   playbackRate: 'playbackrate',
+  defaultMuted: 'muted',
```
cjpillsbury, line 25: issue(blocking): This is inaccurate. `muted` needs to apply dynamically over the life of the react player instance (e.g. if someone wants to programmatically toggle `muted` via props in their react app. There is also no attribute called `defaultMuted`. The `muted` HTML attribute will map to the `defaultMuted` HTML element property. Curious why the change was made here.

cjpillsbury (changes requested):
One callout but this general approach should be fine for simple values like booleans in React 19 (with backwards compat). 👌

cjpillsbury (approved):
🚀

packages/mux-player-react/src/common/utils.ts
cjpillsbury: > confusing stuff.
💯

arcataroger:
❤️

Issue #1038, opened 2024-12-11
Bug: mux-player childList observer behavior


IlyaRazuvaev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player

### Which browsers are you using?

Chrome

### Which operating systems are you using?

macOS

### Description

Hello!
I'm working on integrating a custom element into mux-player, and I'm having trouble with web components. I need to insert an element into mux-player with styles that are loaded directly into the page.
I'm using a similar approach to how Mux handles the poster slot, passing the element through the slot tree like this:

`<slot name="my-slot" slot="my-slot"></slot>`

However, when I do this, the internal structure of my custom element is unexpectedly copied into the mux-video element, which breaks the UI of player element.

### Reduced test case

https://codesandbox.io/p/sandbox/t86gpf?file=%2Findex.html%3A7%2C35

### Steps to reproduce

1. Open sandbox
2. Observe elements of mux-video
![Image](https://github.com/user-attachments/assets/44e46edc-cfbe-4852-9804-7a5c2e3cf9d2)

### Current Behavior

Elements are copied

### Expected Behavior

Top level elements are copied only. (My assumption based on the review of the code related to @mux-player/src/video-api)

### What version of the package are you using?

Latest

PR #1039, opened 2024-12-11
fix: Mux-player childList observer behavior

IlyaRazuvaev:
https://github.com/muxinc/elements/issues/1038

My fix is based on the assumption that this part of code is correct. And only top level children should be copied.
image


IlyaRazuvaev:
Thanks for the suggestion! This is much better.
Tested on my environment, the subtree nodes are not copied.

luwes:
Thanks for making that change! 

I just thought of another way which we improved recently,
https://github.com/muxinc/media-elements/pull/49

I'm gonna add a commit here if it's possible with those changes.

PR #1060, opened 2025-01-13
fix(playback-core): Update hls.js version to fix multi-DRM playready bug.

cjpillsbury:
For details on the bug and fix, see:
- https://github.com/video-dev/hls.js/issues/6947
- https://github.com/video-dev/hls.js/pull/6946

Full smoke testing of the issue requires a Windows machine with Edgeium and Widevine DRM disabled. If available, please reach out for details on playback ids + tokens to test.


luwes (approved):
👍

PR #151, opened 2022-03-28
refactor(mux-embed-types): Clean up our mux-embed typings and related…

cjpillsbury:
…, basing them on anticipating declaration file for mux-embed.

PR #1070, opened 2025-02-19
Update upchunk to v3.5.0

decepulis:
to [v3.5.0](https://github.com/muxinc/upchunk/releases/tag/v3.5.0)
- fix: handle edge case for resuming chunk upload case where multiple chunks have previously been uploaded. by @cjpillsbury in [#141](https://github.com/muxinc/upchunk/pull/141)
- feat: dispatch error when endpoint promise fails to provide a string by @decepulis in [#151](https://github.com/muxinc/upchunk/pull/151)


cjpillsbury (approved):
🚀

PR #1075, opened 2025-02-27
fix(mux-player): Make error dialog not accessible when hidden

AdamJaggard:
A customer noticed in an audit that not hiding the error dialog via `visibility` caused some accessibility issues as it was still reachable in some cases. I couldn't see an issue with proactively setting it to `hidden` while we intend to hide it anyway, so have updated it.


luwes:
should be fine but let me quickly add this in media-chrome, that opacity line was also intended to get into MC...

luwes:
fixed in media-chrome v4.6.1
https://github.com/muxinc/media-chrome/pull/1082

PR #1092, opened 2025-03-28
build: Make CD github action look more like CI github action for tests (uses playwright and retries).

cjpillsbury:
This is to avoid painful consequences of failures wrt release-me side effects.

**NOTE:** Ideally, we'd improve this to not rely on brute forcing/probabilistic assurances, but taking this approach per suggestion from @luwes at least as a stopgap improvement.


luwes (approved):
LGTM 👍

PR #1099, opened 2025-04-09
fix: respect hls.js fatal error flag if available

luwes:
fixes an issue where playback is not broken yet but the player throws a fatal error making the error dialog show up while the video is playing underneath.

### DIFF
--- a/packages/playback-core/src/request-errors.ts
+++ b/packages/playback-core/src/request-errors.ts
@@ -40,6 +40,7 @@ export const getErrorFromResponse = (
   resp: Pick<Response, 'status' | 'url'> | Pick<LoaderResponse, 'code' | 'url'>,
   category: MuxErrorCategoryValue,
   muxMediaEl: Partial<Pick<MuxMediaPropsInternal, 'playbackId' | 'drmToken' | 'playbackToken' | 'tokens'>>,
+  fatal?: boolean,
   translate = false,
   offline = !globalThis.navigator?.onLine // NOTE: Passing this in for testing purposes
 ) => {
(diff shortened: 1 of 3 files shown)


packages/playback-core/src/request-errors.ts
```diff
@@ -40,14 +40,16 @@ export const getErrorFromResponse = (
...
   translate = false,
   offline = !globalThis.navigator?.onLine // NOTE: Passing this in for testing purposes
 ) => {
   if (offline) {
     const message = i18n(`Your device appears to be offline`, translate);
     const context = undefined;
     const mediaErrorCode = MediaError.MEDIA_ERR_NETWORK;
-    const mediaError = new MediaError(message, mediaErrorCode, true, context);
+    // Being offline is not immediately a fatal error for playback.
+    const mediaError = new MediaError(message, mediaErrorCode, false, context);
```
cjpillsbury, line 52: 💪

packages/playback-core/src/request-errors.ts
```diff
@@ -240,14 +242,14 @@ export const getErrorFromResponse = (
...
-    const mediaError = new MediaError(message, mediaErrorCode, true, context);
+    const mediaError = new MediaError(message, mediaErrorCode, fatal ?? true, context);
     mediaError.errorCategory = category;
     mediaError.muxCode = MuxErrorCode.NETWORK_INVALID_URL;
     mediaError.data = resp;
     return mediaError;
   }
 
-  const mediaError = new MediaError('', mediaErrorCode, true);
+  const mediaError = new MediaError('', mediaErrorCode, fatal ?? true);
```
cjpillsbury, line 252: nice! treating fatal as an override of default 👌

cjpillsbury (approved):
one concern about the offline assumptions, but otherwise this lgtm. Also, even if the changes result in some things not being marked as fatal when they should be, you were right that the current code is arguably overly presumptuous in the other direction (at least given the possibilities of retries).

packages/playback-core/src/request-errors.ts
luwes: cool!

PR #1105, opened 2025-04-28
fix: default-hidden-captions order bug for React

### LINKED ISSUE #1062: Bug: `default-hidden-captions` attribute not working when using custom subtitles

mux-player

Chrome, Safari, Firefox

macOS

### Description

When using custom subtitles that are not embedded inside a video, `default-hidden-captions` attribute does not work as expected.

In some cases, the subtitles will still be selected and shown.
And in other cases, subtitle will not be shown, but it will be selected inside the subtitle sub-menu. Requiring the user to first turn subtitles `off` and then select the subtitle again in order to have the subtitles shown inside the player.

### Reduced test case

https://8cnj54-5173.csb.app/

### Steps to reproduce

Code example
```
<mux-player
      playback-id="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"
      metadata-video-title="Big Buck Bunny"
      metadata-viewer-user-id="user-id-1234"
      default-hidden-captions>
      <track
        src="https://raw.githubusercontent.com/brenopolanski/html5-video-webvtt-example/refs/heads/master/MIB2-subtitles-pt-BR.vtt"
        kind="subtitles"
        srclang="en"
        label="English" />
</mux-player>
```

### Current Behavior

When using custom external subtitles, `default-hidden-captions` attribute doesn't turn off subtitles by default.

### Expected Behavior

When using custom external subtitles, `default-hidden-captions` attribute should turn subtitles off by default.

### Errors

_No response_

_No response_

luwes:
fix #1062

I could not reproduce the bug anymore in React but I believe this should improve it even more.

This makes sure the first attribute is set to `default-hidden-captions` in React so that is immediately available in the first render of the theme.

### DIFF
--- a/examples/astro/src/pages/mux-player-react-lazy.astro
+++ b/examples/astro/src/pages/mux-player-react-lazy.astro
@@ -13,7 +13,7 @@ import MuxPlayer from '@mux/mux-player-react/lazy';
     streamType="on-demand"
     autoplay={true}
     muted={true}
-    playbackId="Sc89iWAyNkhJ3P1rQ02nrEdCFTnfT01CZ2KmaEcxXfB008"
+    playbackId="ihZa7qP1zY8oyLSQW9TS602VgwQvNdyIvlk9LInEGU2s"
     metadata={{
       video_title: 'Elephants Dream',
     }}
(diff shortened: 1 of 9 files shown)


Jerricho93 (approved):
LGTM!

PR #1106, opened 2025-04-28
feat: retry logic for 412 not playable errors

### LINKED ISSUE #531: Feature request: mux-video should retry on 412 Not Playable
> `<mux-video>` will automatically handle recoverable errors that happen during video playback. This is particularly handy for live streams that may experience disconnects.

However, the player does not retry if the video/livestream if Mux sends `412 Not Playable`.

Currently I do the following:
```js
videoRef.addEventListener('error', () => {
  setTimeout(() => {
    videoRef.playbackId = 'dummy'
    videoRef.playbackId = actualPlaybackId // artificially trigger reload by changing playbackId
  }, 1000)
})
```

But it would be helpful to have builtin support for this

Pintorado:
Closes #531 

This PR refactors how the player handles 412 (NETWORK_NOT_READY) errors.

## Changes
Behavior:
* Error dialog says “Live stream/video is not currently available. Retrying in [xx] seconds…” (different message depending if it's a live stream or VOD).
* Retries after 5s the first time and then the retry time increases to 60s. It retries 5 more times with a 60s gap.
* Stops retrying after 5 minutes and 5s and if there was no success on recovering the live stream/video, the UI shows "Live stream/video is not currently available. Try again later or **click here to retry**” with a retry link.


ronalduQualabs (approved):

packages/playback-core/src/errors.ts
```diff
@@ -80,6 +80,7 @@ export class MediaError extends Error {
   context?: string;
   fatal: boolean;
   data?: any;
+  streamType?: 'live' | 'on-demand' | 'unknown';
```
luwes, line 83: this shouldn't be needed because there is a function called `getStreamType()` in playback core.
you can use that to get the stream type in mux player

luwes: actually scratch that, I see there is no easy way to get the media element in there.

Pintorado: Yep, I had trouble with that exactly

packages/mux-player/src/errors.ts
```diff
@@ -101,9 +102,6 @@ const muxMediaErrorToDialogMessage = (mediaError: MediaError, translate = false)
     if (mediaError.muxCode === MuxErrorCode.NETWORK_NOT_FOUND) {
       return '';
     }
-    if (mediaError.muxCode === MuxErrorCode.NETWORK_NOT_READY) {
-      return i18n(`The live stream or video file are not yet ready.`, translate);
-    }
```
luwes, line 106: why was this block removed?

packages/mux-player/src/errors.ts
```diff
@@ -35,7 +35,8 @@ const muxMediaErrorToDialogTitle = (mediaError: MediaError, translate = false) =
       return i18n(`{category} does not exist`, translate).format({ category });
     }
     if (mediaError.muxCode === MuxErrorCode.NETWORK_NOT_READY) {
-      return i18n(`{category} is not currently available`, translate).format({ category });
+      const mediaType = mediaError.streamType === 'live' ? 'Live stream' : 'Video';
+      return i18n(`{mediaType} is not currently available`, translate).format({ mediaType });
```
luwes, line 39: does this replace the deleted block below?

Pintorado: Yes, exactly

packages/playback-core/src/index.ts
```diff
@@ -1251,8 +1251,71 @@ export const loadMedia = (
...
+          saveAndDispatchError(mediaEl, retryDelayError);
+
+          setTimeout(() => {
+            state.retryCount = retryCount + 1;
+            if (data.details === 'manifestLoadError' && data.url) {
+              hls.loadSource(data.url);
+              hls.attachMedia(mediaEl);
+            } else {
+              hls.recoverMediaError();
+            }
```
luwes: I think this shouldn't be used here. recoverMediaError() is an attempt to recover a media decode error is my understanding or at least it's definitely not for a network error.

packages/playback-core/src/index.ts
```diff
@@ -1251,8 +1251,71 @@ export const loadMedia = (
...
+            error.fatal
+          );
+          Object.assign(retryDelayError, error);
+          saveAndDispatchError(mediaEl, retryDelayError);
+
+          setTimeout(() => {
+            state.retryCount = retryCount + 1;
+            if (data.details === 'manifestLoadError' && data.url) {
+              hls.loadSource(data.url);
+              hls.attachMedia(mediaEl);
```
luwes: is hls.attachMedia needed? was it removed somewhere else?

packages/playback-core/src/index.ts
```diff
@@ -1251,8 +1251,71 @@ export const loadMedia = (
...
+      saveAndDispatchError(mediaEl, error);
     });
+
+    hls.on(Hls.Events.MANIFEST_LOADED, () => {
+      // Clear error state and UI
+      const state = muxMediaState.get(mediaEl);
+      if (state) {
+        state.error = null;
+        state.retryCount = 0;
+      }
```
luwes: isn't this called each time after a hls.loadSource call?
how will retryCount then ever reach max retries?

packages/playback-core/src/request-errors.ts
```diff
@@ -205,6 +207,12 @@ export const getErrorFromResponse = (
     const mediaError = new MediaError(message, mediaErrorCode, fatal ?? true, context);
     mediaError.errorCategory = category;
     mediaError.muxCode = MuxErrorCode.NETWORK_NOT_READY;
+    mediaError.streamType =
+      muxMediaEl.streamType === StreamTypes.LIVE
+        ? 'live'
+        : muxMediaEl.streamType === StreamTypes.ON_DEMAND
+          ? 'on-demand'
+          : 'unknown';
```
luwes, line 215: 👍

packages/playback-core/src/index.ts
Pintorado: You're right, I'll remove it

packages/playback-core/src/index.ts
Pintorado: The `MANIFEST_LOADED` event only fires when the media successfully loads, which means the retry was successful. That's why we reset the retry count to 0 - it indicates we've recovered from the error. If the error persists, this event won't fire and the retry count will continue to increment.

packages/playback-core/src/index.ts
Pintorado: It's redundant since we already call it in line 1323. We only need to call `loadSource` to retry loading the manifest. I'll remove this. Thanks for pointing this out!

packages/mux-player/src/errors.ts
Pintorado: Sorry, this isn't quite correct. This text is the **title** of the error dialog and the other block has the **message** text which was replaced with the following code in [playback-core/src/index.ts](https://github.com/muxinc/elements/pull/1106/files#diff-d4f0a33fe7d9652424deb76bddd87dc31a3486e0a509c7c15c669bd518e994f3R1266-R1270), since I needed to dynamically show the `retryDelay` value.
```
const retryDelayError = new MediaError(
            `Retrying in ${retryDelay / 1000} seconds...`,
            error.code,
            error.fatal
          );
```
Here's a screenshot of how the dialog looks for an `idle` live stream. I hope this makes things clearer!
image

packages/playback-core/src/index.ts
```diff
@@ -1251,8 +1251,68 @@ export const loadMedia = (
...
+
+      const clearError = new MediaError('', MediaError.MEDIA_ERR_CUSTOM, false);
+      clearError.muxCode = MuxErrorCode.NOT_AN_ERROR;
+      mediaEl.dispatchEvent(
+        new CustomEvent('error', {
+          detail: clearError,
+          composed: true,
+          bubbles: true,
+        })
+      );
```
luwes: we should find something else to clear the error dialog. 
this is too hacky :)

I don't love this either but try dispatching an emptied followed by a loadstart event. that feels more inline with what's happening and media-chrome listens to emptied to change media error state.

Try without composed and bubbles, that is only needed for non native video events. if custom-media-element is used, it forwards native video events from the native video element. https://github.com/muxinc/media-elements/blob/45b7383d445d2969409ec369aa34e967c4d32cab/packages/custom-media-element/custom-media-element.ts#L345-L359

Like this:

```js
mediaEl.dispatchEvent(new Event('emptied'));
mediaEl.dispatchEvent(new Event('loadstart'));
```

Pintorado: Thanks, I wasn't sure how to do this part, really helpful explanation! 😊

packages/playback-core/src/index.ts
```diff
@@ -1251,8 +1251,63 @@ export const loadMedia = (
...
+    hls.on(Hls.Events.MANIFEST_LOADED, () => {
+      // Clear error state and UI
+      const state = muxMediaState.get(mediaEl);
+      if (state) {
+        state.error = null;
+        state.retryCount = 0;
+      }
+
+      const clearError = new MediaError('', MediaError.MEDIA_ERR_CUSTOM, false);
+      clearError.muxCode = MuxErrorCode.NOT_AN_ERROR;
```
```suggestion
```

luwes (changes requested):
tests seem to be failing, could you debug?

packages/playback-core/src/index.ts
```diff
@@ -1251,8 +1251,61 @@ export const loadMedia = (
...
+    hls.on(Hls.Events.MANIFEST_LOADED, () => {
+      // Clear error state and UI
+      const state = muxMediaState.get(mediaEl);
+      if (state) {
+        state.error = null;
+        state.retryCount = 0;
+      }
+
+      mediaEl.dispatchEvent(new Event('emptied'));
+      mediaEl.dispatchEvent(new Event('loadstart'));
```
luwes: just noticed here, these shouln't be dispatched if there wasn't an error present on the state first. so something like

```js
      const state = muxMediaState.get(mediaEl);
      if (state && state.error) {
        state.error = null;
        state.retryCount = 0;

        mediaEl.dispatchEvent(new Event('emptied'));
        mediaEl.dispatchEvent(new Event('loadstart'));
      }
```

PR #1127, opened 2025-05-22
feat: add adblocker bypass option, expose player props, and conditional captions menu

ronalduQualabs:
This PR introduces the following improvements:
	•	New `allowPlaybackWithAdBlocker` prop
Allows the player to continue playback even when an adblocker is detected.
	  •	If false (default): shows a blocking popup when an adblocker is present.
	  •	If true: bypasses the restriction and allows normal playback.
	•	Exposes underlying player props on MuxNewsPlayer
Enables more granular configuration and control over the embedded player through MuxNewsPlayer.
	•	Conditional captions menu
Adds a captions menu that appears only when text tracks are available.
	•	Allows users to disable captions if they prefer not to see them.

### DIFF
--- a/packages/mux-video-ads/src/index.ts
+++ b/packages/mux-video-ads/src/index.ts
@@ -19,11 +19,13 @@ const serializeAttributes = (attrs = {}) => {
 const Attributes = {
   AD_TAG_URL: 'adtagurl',
   AD_BREAK: 'adbreak',
+  ALLOW_PLAYBACK_WITH_AD_BLOCKER: 'allow-playback-with-ad-blocker',
 } as const;
 
 class MuxVideoAds extends MuxVideoElement {
   #muxAdManager: MuxAdManager | undefined;
   #lastCurrentime: number | undefined;
+  #hasAllowPlaybackWithAdBlocker: boolean | undefined = false;
 
   static getTemplateHTML = (attrs: Record<string, string>) => {
     return `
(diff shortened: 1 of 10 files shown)


packages/mux-video-ads/src/index.ts
```diff
@@ -94,6 +102,18 @@ video::-webkit-media-text-track-container {
...
 
+  static get observedAttributes() {
+    return [...(super.observedAttributes ?? []), Attributes.ALLOW_PLAYBACK_WITH_AD_BLOCKER];
+  }
+
+  attributeChangedCallback(name: string, _oldValue: string | null, _newValue: string | null) {
+    super.attributeChangedCallback?.(name, _oldValue, _newValue);
+    if (name === Attributes.ALLOW_PLAYBACK_WITH_AD_BLOCKER) {
+      console.log('ALLOW_PLAYBACK_WITH_AD_BLOCKER', this.allowPlaybackWithAdBlocker);
+      this.#hasAllowPlaybackWithAdBlocker = this.allowPlaybackWithAdBlocker;
```
cjpillsbury, line 113: question(blocking): Do we need this private variable? I think we get all of the functionality we need from `get hasAllowPlaybackWithAdBlocker()`. If that's accurate, refactor and use it instead.

packages/mux-video-ads/src/index.ts
```diff
@@ -94,6 +102,18 @@ video::-webkit-media-text-track-container {
   `;
   };
 
+  static get observedAttributes() {
+    return [...(super.observedAttributes ?? []), Attributes.ALLOW_PLAYBACK_WITH_AD_BLOCKER];
```
cjpillsbury, line 106: question(blocking): Assuming we don't need the private var (per other question, below), I think we shouldn't need this to be an actual `observedAttribute` since we (would) have no need for "reactive" code in response to changes to the attribute. If that's correct, remove from observed attributes.

packages/mux-video-ads/src/index.ts
ronalduQualabs: Thank you for the feedback, I made this suggestion.

packages/mux-video-ads/src/index.ts
ronalduQualabs: Done!

packages/mux-video-ads/src/index.ts
cjpillsbury: I think you misunderstood. Why do we need `this.#hasAllowedPlaybackWithAdBlocker` at all?

cjpillsbury (approved):
👍

PR #1128, opened 2025-05-23
feat: Google IMA support for mux-player and mux-video variants

### LINKED ISSUE #1120: Expose ads API modules
Proposal steps to expose the ads related modules consumable by the official public mux-video and mux-player packages.
mux-player-react will export a playlist component but will be not official until further notice.

1. A new module/export in mux-video that “knows about ads”
    1. Not part of the default module
    2. For now, does not need to be a “generic”
    3. Needs to pull in Mux Data SDK extension for ads
2. A new module/export in mux-player that “knows about ads” and uses mux-video export in (1)
3. A new module/export in mux-player-react that is basically moving the playlist player out into its own module
4. Make sure we aren’t duplicating dependencies in bundles/bundle size (e.g. `@mux/mux-data-google-ima` vs. `mux-embed`, `mux-video` vs. `mux-video` (with ads))
5. Split out web component registration vs. web component definition so that e.g. `<mux-video>` can be registered as the ads-flavored version when importing that one

luwes:
fix #1120

This change adds support for the Google IMA client-side to the production variants of mux-video and mux-player.
Support is currently limited to linear pre, mid, and post roll ads. All ads logic is isolated from the standard mux-video and mux-player variants and requires a separate import path to include ad support.

## Examples

- MuxPlayer: https://elements-demo-nextjs-git-ads-video-mixin-mux.vercel.app/MuxPlayerAds
- MuxVideo: https://elements-demo-nextjs-git-ads-video-mixin-mux.vercel.app/mux-video-ads-react
- news player (POC): https://elements-demo-nextjs-git-ads-video-mixin-mux.vercel.app/MuxNewsPlayer

## Readme

- mux-video (ads): https://github.com/muxinc/elements/blob/ads-video-mixin/packages/mux-video/src/ads/README.md

#### web component

```js
import '@mux/mux-player/ads';
```

```html
<mux-player playback-id="..." ad-tag-url="..." allow-ad-blocker></mux-player>
```

#### React

```js
import MuxPlayer from '@mux/mux-player-react/ads';

const Component = () => <MuxPlayer playbackId="..." adTagUrl="..." allowAdBlocker={true} />;
```

#### web component

```js
import '@mux/mux-video/ads';
```

```html
<mux-video playback-id="..." ad-tag-url="..." allow-ad-blocker></mux-video>
```

#### React

```js
import MuxVideo from '@mux/mux-video/ads/react';

const Component = () => <MuxVideo playbackId="..." adTagUrl="..." allowAdBlocker={true} />;
```


packages/mux-video/src/ads/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      },
+    } as const;
+  }
+
+  get castCustomData() {
+    return this.#castCustomData ?? this.muxCastCustomData;
+  }
+
+  set castCustomData(val: Record<string, any> | undefined) {
+    this.#castCustomData = val;
```
luwes, line 77: Yes, this code is repeated in `src/index.ts`. 
I'd love to create a outer mixin to DRY that code but I couldn't get the types to work. 

The downlevel-dts kept complaining about "Error: Bad AST -- accessor parent should be a class declaration."

This is starter code if you like to give it a go:

```ts
import { CastableMediaMixin } from 'castable-video/castable-mixin.js';
import { MediaTracksMixin } from 'media-tracks';
import { Autoplay } from '@mux/playback-core';
import type { MuxVideoBaseElement } from './mux-video-base-element';
import type { CustomMediaElementConstructor } from 'custom-media-element';

export function MuxVideoMixin<T extends CustomMediaElementConstructor<MuxVideoBaseElement>>(BaseElement: T) {
  // castable-video should be mixed in last so that it can override load().
  return class MuxVideoMixinClass extends CastableMediaMixin(MediaTracksMixin(BaseElement)) {
    // Define autoplay in the most outer layer because mux-video accepts string | boolean
    // which is not compatible the CustomVideoElement.autoplay boolean only type.
    /** @ts-ignore */
    get autoplay(): Autoplay {
      const attr = this.getAttribute('autoplay');

      if (attr === null) {
        return false;
      } else if (attr === '') {
        return true;
      } else {
        return attr as Autoplay;
      }
    }

    /** @ts-ignore */
    set autoplay(val: Autoplay) {
      const currentVal = this.autoplay;
      if (val === currentVal) {
        return;
      }

      if (val) {
        this.setAttribute('autoplay', typeof val === 'string' ? val : '');
      } else {
        this.removeAttribute('autoplay');
      }
    }

    // NOTE: CastableMediaMixin needs to be a subclass of whatever implements the load() method
    // (i.e. MuxVideoBaseElement), but we're overriding castCustomData to provide mux-specific
    // values by default, so it needs to be defined here (i.e. in the composed subclass of
    // CastableMediaMixin). (CJP)
    #castCustomData: Record<string, any> | undefined;

    get muxCastCustomData() {
      return {
        mux: {
          // Mux Video values
          playbackId: this.playbackId,
          minResolution: this.minResolution,
          maxResolution: this.maxResolution,
          renditionOrder: this.renditionOrder,
          customDomain: this.customDomain,
          /** @TODO Add this.tokens to MuxVideoElement (CJP) */
          tokens: {
            drm: this.drmToken,
          },
          // Mux Data values
          envKey: this.envKey,
          metadata: this.metadata,
          disableCookies: this.disableCookies,
          disableTracking: this.disableTracking,
          beaconCollectionDomain: this.beaconCollectionDomain,
          // Playback values
          startTime: this.startTime,
          // Other values
          preferCmcd: this.preferCmcd,
        },
      } as const;
    }

    get castCustomData() {
      return this.#castCustomData ?? this.muxCastCustomData;
    }

    set castCustomData(val: Record<string, any> | undefined) {
      this.#castCustomData = val;
    }
  };
}
```

packages/mux-video/src/ads/react.ts
```diff
@@ -0,0 +1,29 @@
+'use client';
```
cjpillsbury, line 1: we didn't discuss this one, but we should decide if we want this to be "owned" by the `mux-video-react` package or the `mux-video` package. This doesn't really fit our status quo pattern, so my S1 is to move it for now, but that's pretty weakly held?

packages/mux-video/src/ads/mixin.ts
```diff
@@ -0,0 +1,426 @@
...
+        console.error('Missing google.ima SDK. Make sure you include it via a script tag.');
+        if (!this.allowAdBlocker) {
+          this.#showAdBlockedMessage();
+        } else {
+          this.#adBreak = false;
+        }
+        return;
+      }
+
+      const config: GoogleImaAdsProviderConfig = {
```
cjpillsbury: thought(non-blocking): we should probably iterate on several of these implementation/naming details but this should be treated as lower priority and definitely as a followup since it's currently just for "under the hood" logic in (extended) mux-video.

packages/mux-video/src/ads/mixin.ts
```diff
@@ -0,0 +1,426 @@
...
+        <h4>Ad experience unavailable.</h4>
+        <span>This may be due to a missing SDK, network issue, or ad blocker.</span>
+      `;
+      this.shadowRoot?.getElementById('mainContainer')?.appendChild(fallback);
+    }
+
+    #setupEventListeners(): void {
+      this.addEventListener(
+        'loadedmetadata',
+        () => {
```
cjpillsbury: issue(blocking): inline event handler... let's make sure we reliably clean up this and the onews below.

packages/mux-video/src/ads/google-ima-ads-provider.ts
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  setupAdsManager() {
+    if (!this.#adDisplayContainer) {
+      this.#adDisplayContainer = new google.ima.AdDisplayContainer(this.#adContainer, this.#videoElement);
+      this.#adsLoader = new google.ima.AdsLoader(this.#adDisplayContainer);
+
+      this.#adsLoader.addEventListener(
+        google.ima.AdsManagerLoadedEvent.Type.ADS_MANAGER_LOADED,
+        (adsManagerLoadedEvent: google.ima.AdsManagerLoadedEvent) => {
```
cjpillsbury: more inline event handlers we should be careful about, especially since this stuff is created/torn down in `connectedCallback`

packages/mux-player-react/src/themes/news-theme.tsx
```diff
@@ -0,0 +1,593 @@
+import 'media-chrome/react';
+import 'media-chrome/react/menu';
+import React from 'react';
+
+export default function NewsTheme() {
```
cjpillsbury: issue(blocking): This bit of refactor slipped through the cracks. From a quick scan at least, it looks like everything new in here is ads specific and not e.g. playlist related (which looks like it's all managed in the jsx/react side for playlist player). We should probably just pull this out and make it like our other html templates. Also move it to `mux-player` package and use it as default for `mux-player` ads player theme.

cjpillsbury: we can circle back around and discuss if you want to try to break this effort up per "shortest path to removing one-off branch"/releases

examples/vanilla-ts-esm/public/mux-player-error.html
```diff
@@ -41,7 +41,7 @@ <h1><a href="/">Elements</a></h1>
     <!-- Correct playbackid: ihZa7qP1zY8oyLSQW9TS602VgwQvNdyIvlk9LInEGU2s -->
     <mux-player
       stream-type="on-demand"
-      playback-id="Sc89iWAyNkhJ3P1rQ02nrEdCFTnfT01CZ2KmaEcxXfB00"
+      playback-id="ihZa7qP1zY8oyLSQW9TS602VgwQvNdyIvlk9LInEGU2"
```
cjpillsbury, line 44: question(blocking): Should this have been committed? (name is "mux-player-error.html" so just confirming here)

packages/mux-player-react/package.json
```diff
@@ -68,19 +81,17 @@
   "license": "MIT",
   "scripts": {
     "clean": "shx rm -rf dist/",
-    "dev:cjs": "npm run build:cjs -- --watch=forever",
```
cjpillsbury, line 71: question(blocking): why were all of the dev scripts removed here? this feels out of scope.

packages/mux-player-react/src/index.tsx
```diff
@@ -88,6 +88,9 @@ export type MuxPlayerProps = {
   programStartTime?: number;
   programEndTime?: number;
   proudlyDisplayMuxBadge?: boolean;
+  /** Allow playback with ad blocker */
+  allowAdBlocker?: boolean;
```
cjpillsbury: issue(blocking): We shouldn't be adding these to our standard `MuxPlayer` implementation. Like the `mux-player` package, we should keep the ads-specific code in a separate module and react class that simply "decorates" this version.

packages/mux-player-react/src/news/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  function selectVideo(index: number) {
+    setIsEndScreenVisible(false);
+    setCurrentIndex(index);
+    setTimeout(() => {
+      mediaElRef.current.play();
+    }, 200);
+  }
+
+  return (
+    <div>
```
cjpillsbury: question(non-blocking): Do we need this parent `<div>`?

packages/mux-player/package.json
```diff
@@ -35,29 +35,11 @@
...
-      "import": "./dist/themes/classic/index.mjs",
-      "require": "./dist/themes/classic/index.cjs.js",
-      "default": "./dist/themes/classic/index.cjs.js"
-    },
-    "./themes/gerwig": {
-      "browser": "./dist/themes/gerwig/index.mjs",
-      "import": "./dist/themes/gerwig/index.mjs",
-      "require": "./dist/themes/gerwig/index.cjs.js",
-      "default": "./dist/themes/gerwig/index.cjs.js"
+    "./themes/*": {
```
cjpillsbury, line 70: question(blocking): Did you confirm this refactor doesn't have any issues with various build tools? if not, this feels out of scope.

packages/mux-player/src/index.ts
```diff
@@ -87,7 +88,9 @@ const PlayerAttributes = {
   CAST_RECEIVER: 'cast-receiver',
   NO_TOOLTIPS: 'no-tooltips',
   PROUDLY_DISPLAY_MUX_BADGE: 'proudly-display-mux-badge',
-};
+  AD_TAG_URL: 'ad-tag-url',
```
cjpillsbury: issue(blocking): Sadly, this isn't quite what we discussed. Mux Player should not be adding ads-specific properties, events, features, or functionality to its official main package export. Like mux-video, this should be a separate export/module.

packages/mux-player/src/template.ts
```diff
@@ -134,9 +135,11 @@ export const content = (props: MuxTemplateProps) => html`
       cast-src="${!!props.src ? props.src : props.playbackId ? toMuxVideoURL(props) : false}"
       cast-receiver="${props.castReceiver ?? false}"
       drm-token="${props.tokens?.drm ?? false}"
+      ad-tag-url="${props.adTagUrl ?? false}"
+      allow-ad-blocker="${props.allowAdBlocker ?? false}"
       exportparts="video"
     >
-      ${props.storyboard
+      ${!props.adBreak && props.storyboard
```
cjpillsbury: thought(non-blocking): I get why this is here, but it feels like the wrong place for this to live longer term. We should instead solve this in the theme itself if possible.

packages/mux-player/src/video-api.ts
```diff
@@ -55,84 +55,41 @@ const emptyTimeRanges: TimeRanges = Object.freeze({
...
-  | 'mux' // NOTE: Because of our global types extension of HTMLMediaElement, `mux` is a property that also needs to be omitted (CJP)
->;
+export const AttributeNames = [...AllowedVideoAttributeNames, ...CustomVideoAttributesNames];
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
-interface VideoApiElement extends PartialHTMLVideoElement, HTMLElement {
-  addEventListener<K extends keyof HTMLVideoElementEventMap>(
-    type: K,
-    listener: (this: HTMLVideoElement, ev: HTMLVideoElementEventMap[K]) => any,
+interface VideoApiElement extends HTMLElement {
```
cjpillsbury: question(blocking): This doesn't look like an accurate description of the `VideoApiElement` interface (aka dropping `PartialHTMLVideoElement` without adding an appropriate equivalent). Why was this changed?

packages/mux-video/package.json
```diff
@@ -60,33 +92,35 @@
     "lint": "ESLINT_USE_FLAT_CONFIG=false eslint src/ --ext .js,.jsx,.ts,.tsx",
     "test": "web-test-runner **/*test.js --port 8002 --coverage --config test/web-test-runner.config.mjs --root-dir ../..",
     "posttest": "replace 'SF:src/' 'SF:packages/mux-video/src/' coverage/lcov.info --silent",
-    "dev:iife": "npm run build:iife -- --watch=forever",
```
cjpillsbury, line 63: question(blocking): why were these dev scripts removed? feels out of scope.

packages/mux-video/src/index.ts
```diff
@@ -1,261 +1,15 @@
...
-    if (val) {
-      this.setAttribute(Attributes.TYPE, val);
-    } else {
-      this.removeAttribute(Attributes.TYPE);
-    }
-  }
+export * from '@mux/mux-video/base';
 
+// castable-video should be mixed in last so that it can override load().
+class MuxVideoElement extends CastableMediaMixin(MediaTracksMixin(MuxVideoBaseElement)) {
```
cjpillsbury, line 10: praise(non-blocking): 😎

cjpillsbury (changes requested):
overall this is heading in the right direction, though I have a few callout concerns

examples/vanilla-ts-esm/public/mux-player-error.html
luwes: yes, this is a playback id with the last character left off so it errors out. I just didn't catch this last time when I did a project wide search/replace of the playback id starting with `Sc89iWAyNkhJ3P1rQ02nrEdCFTnfT01CZ2KmaEcxXfB00`

you can see above the complete playback id was indeed replaced.

packages/mux-player-react/package.json
luwes: I did this as a result of having to add more npm scripts in for the added "packages", it resulted in lot of extra lines in npm scripts and I searched for ways to reduce that. 

if you think having these isolated dev scripts is useful happy to add them back in? 
I didn't think anyone used those separately so this felt more minimal.

packages/mux-player/package.json
luwes: I do feel this is in scope because the news theme is added in here.
very sure this works because it's in the pkg.exports spec and we use this in media-chrome as well.

packages/mux-player/src/template.ts
luwes: yes, maybe another location but in the theme is not possible I think. the media is slotted via the forwarding slot so we don't have much control there.

packages/mux-player/src/video-api.ts
luwes: this change was related to the event listener changes. 

I assumed here that this PartialHTMLVideoElement was with the thought that VideoApiElement extends HTMLVideoElement or CustomVideoElement but it extends HTMLElement and re-implements the video methods.

I'll put this interface back in, it makes more sense now.

packages/mux-video/package.json
luwes: cool, I'll add the dev scripts back in.

packages/mux-player-react/package.json
cjpillsbury: I sometimes use them. It just makes reviews harder when there are many changes that are not related to the scope of the primary effort, especially when there are lots of changes already. It also makes it harder to assess whether there are implications of these changes that I need to reason through.

packages/mux-player/package.json
cjpillsbury: Right we just haven't been doing as much due diligence on build tests these days across different tools. I simply am not sure if any/all of them reliably work with `*`, particularly the older ones. If you're confident that this is fine, then I defer to you.

packages/mux-player/src/template.ts
cjpillsbury: Cool yeah I will have to chew on where/how to abstract this bit... maybe we rethink the logic here and don't avoid slotting/not slotting based on `adBreak`. Feels like the implications of this should be "owned" by media chrome and/or the chromeless media element (e.g. mux-video with ads), but that should also likely be a future effort. I wasn't sure if we could use the template if or similar acrobatics re: theme (e.g. to conditionally disable thumbnails when in an ad break) as a different angle here.

packages/mux-player/src/index.ts
luwes: yes, I think you weren't there that time when I brought this up in a later meeting.

I'll make an attempt to isolate this but it feels lots of splitting for a very small amount of code additions.

packages/mux-player-react/package.json
luwes: yes, understandable. I did have to touch the npm scripts here to add the news bundles so these lines would have changed in any case to include the news bundle.

packages/mux-player/src/ads/mixin/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      super.connectedCallback();
+
+      if (this.media) {
+        this.media.allowAdBlocker = this.allowAdBlocker;
+        this.media.adTagUrl = this.adTagUrl;
+      }
+
+      this.addEventListener('adbreakstart', () => {
+        this.mediaTheme?.toggleAttribute('mediaadbreak', true);
+        // TODO: remove any track elements from mux-video!!!!
```
luwes: remember to fix this for iOS MSE issue!!!!

packages/mux-video/src/ads/react.ts
luwes: we could add an alias but it was here because it uses a React wrapper around the web component similar to `@mux/mux-video/react` which is different than `@mux/mux-video-react`.

but yes `@mux/mux-player-react` is also wrapper so in that sense I see your point.

I would like to really push for simplifying all this like we talked about because `@mux/mux-video-react` is missing a bunch of things now. all of them could be React wrappers around the web component with DSD. possibly even mux-player-react could use ce-la-react.

packages/mux-player/src/ads/mixin/index.ts
luwes: resolved ✅

packages/mux-video/src/ads/react.ts
cjpillsbury: Yup for sure worth keeping in mind, though imo that still doesn't answer the question of where that wrapper should live (aka in its own package or in a separate package).

cjpillsbury (approved):
Phase 1 complete! 🚀

cjpillsbury (changes requested):
Found an issue smoke testing https://elements-demo-nextjs-git-ads-video-mixin-mux.vercel.app/MuxNewsPlayer (play first ad + video, let 2nd play, note that ad does not play)

cjpillsbury (approved):
Confirmed fix works and code changes make sense! One callout re: `src` vs. `playback-id` scenarios, but can treat as a fast follow if it's complicated. Pre-approving either way. 🚀

cjpillsbury (approved):
🏃‍♂️

luwes:
@claude

luwes:
@claude could you give a summary of the changes in this PR?

PR #1130, opened 2025-05-26
Feature/wire up free plan logo mux video

ronalduQualabs:
This PR adds support for displaying the Mux logo [Issue](https://github.com/muxinc/devextravaganza/issues/33) in the video component when the current session is identified as part of the “mux-free-plan”. Also the [Issue](https://github.com/muxinc/devextravaganza/issues/10) which loads the metadata in mp4 urls from mux
- Parses #EXT-X-SESSION-DATA from the HLS manifest to extract the external metadata URL.
- Set metadata Url in the mux-video
- on Fetch metadata url call renderLogo if its "mux-free-plan"


packages/mux-video/src/index.ts
```diff
@@ -831,7 +798,13 @@ class MuxVideoBaseElement extends CustomVideoElement implements Partial<MuxMedia
         if (newValue) {
           fetch(newValue)
             .then((resp) => resp.json())
-            .then((json) => (this.metadata = json))
+            .then((json) => {
+              this.metadata = json;
+              const metadata = json?.[0]?.metadata ?? [];
+              const planMeta = metadata.find((m: { value: string }) => m.value === 'mux-free-plan');
+
+              if (planMeta) this.logo = 'default';
```
luwes, line 806: don't set `this.logo` here but an internal state property like `this.#logo`. 
make the getter fallback to `this.#logo` if the logo attribute is not set.

```js
this.#logo = ''
this.updateLogo()
```

packages/playback-core/src/media-tracks.ts
```diff
@@ -131,3 +136,20 @@ export function setupMediaTracks(
...
   hls.once(Hls.Events.DESTROYING, removeAllMediaTracks);
 }
+
+function fetchMetadata(el: HTMLMediaElement, metadataUrl: string) {
+  fetch(metadataUrl)
+    .then((resp) => resp.json())
+    .then((json) => {
+      // Attempt to set metadata on the custom element
+      if ('metadata' in el) {
+        (el as any).metadata = json;
```
luwes: don't use the public element API internally here. `fetchMetadata` will have to be a private method because we want to set a private property.

packages/playback-core/src/media-tracks.ts
```diff
@@ -131,3 +136,20 @@ export function setupMediaTracks(
...
+      }
+
+      const metadata = json?.[0]?.metadata ?? [];
+      const planMeta = metadata.find((m: { value: string }) => m.value === 'mux-free-plan');
+
+      if (planMeta) el.setAttribute('logo', 'default');
```
luwes: again here don't sprout the logo attribute. the outer HTML of this element shouldn't be changed by something internally. use internal state, private props etc.

packages/mux-video/src/index.ts
```diff
@@ -130,24 +128,17 @@ class MuxVideoBaseElement extends CustomVideoElement implements Partial<MuxMedia
...
-      <slot name="logo">
-        ${hasLogoSrc ? `<img class="logo" part="logo" src="${logoSrc}" />` : muxLogo}
-      </slot>
-    `
-      : '';
-
-    return `
-      ${template}
-      ${logoTemplate}
+      <slot name="logo"></slot>
```
luwes: this is better but we're still missing the img tag in the string representation

packages/playback-core/src/media-tracks.ts
luwes: hmm, just noticed this is playback-core / media-tracks...

- this logic shouldn't be in this file.
- this will have to work differently because what happens if the user has set metadata themselves. this will override it. I think we should create a new getter function in playback core to get session data from hls.js.

packages/playback-core/src/media-tracks.ts
luwes: this logic shouldn't go here but in mux-video

packages/mux-video/src/index.ts
```diff
@@ -848,11 +856,32 @@ class MuxVideoBaseElement extends CustomVideoElement implements Partial<MuxMedia
...
         break;
     }
   }
 
+  getLogoHTML(logoValue: string | null = null): string {
+    const logo = logoValue ?? this.#logo ?? this.logo;
+    if (!logo || logo === 'false') return '';
+
+    return logo === 'default' ? muxLogo : `<img class="logo" part="logo" src="${logo}" />`;
+  }
```
luwes: can you make this a static method?

`this.getLogoHTML(attrs[Attributes.LOGO])` in static
`this.constructor.getLogoHTML(this.#logo ?? this.logo)` in the instance

packages/playback-core/src/index.ts
```diff
@@ -166,10 +166,24 @@ export const getStreamInfoFromSrcAndType = async (src: string, type?: MediaTypes
 export const updateStreamInfoFromSrc = async (
   src: string,
   mediaEl: HTMLMediaElement,
-  type: MediaTypes | '' = getType({ src })
+  type: MediaTypes | '' = getType({ src }),
+  customElement: HTMLMediaElement
 ) => {
   const { streamType, targetLiveWindow, liveEdgeStartOffset } = await getStreamInfoFromSrcAndType(src, type);
 
+  if (streamType === StreamTypes.ON_DEMAND && src.endsWith('.mp4') && !src.includes(MUX_VIDEO_DOMAIN)) {
```
luwes, line 173: it has to be a a Mux stream url right?
use the existing `toPlaybackIdFromSrc()` function

packages/playback-core/src/index.ts
```diff
@@ -1497,3 +1511,34 @@ const getErrorFromHlsErrorData = (
...
+      }
+      const metadata = json?.[0]?.metadata ?? [];
+      const planMeta = metadata.find((m: { value: string }) => m.value === 'mux-free-plan');
+
+      const event = new Event('setdefaultlogo', {
+        bubbles: true,
+        composed: true,
+      });
+
+      if (planMeta) el.dispatchEvent(event);
```
luwes: yes, this needs some rethinking. 

that "metadata" check in the media element shouldn't be there.
playback-core shouldn't have to know about a custom media element.

I think it should be something like playback-core fetches the metadata url 
sets the metadata to internal state 
sends out an event `muxmetadata`.

also noting the struture of this metadata url response metadata object

```json
        [
            {
               "key": "com.mux.video.branding",
               "value": "mux-free-plan"
            }
            // Additional metadata fields can be provided here, flexibly
        ]
```

  needs to be converted to the existing "metadata" property structure which is `{ key: value }`.

another thing is if the user has set some metadata already, it shouldn't be overridden but merged I believe. unless it's the prop indicating the free plan, that has to take priority from the url response.

packages/playback-core/src/index.ts
ronald-urbina: In that function I don't have the element to send to the function `fetchmetadata`

packages/playback-core/src/index.ts
```diff
@@ -166,10 +166,24 @@ export const getStreamInfoFromSrcAndType = async (src: string, type?: MediaTypes
 export const updateStreamInfoFromSrc = async (
   src: string,
   mediaEl: HTMLMediaElement,
-  type: MediaTypes | '' = getType({ src })
+  type: MediaTypes | '' = getType({ src }),
+  customElement: HTMLMediaElement
```
luwes: remove `customElement` here. in playback-core we try not to know about that. (media-tracks is an exception)

packages/playback-core/src/index.ts
```diff
@@ -1135,7 +1153,7 @@ export const loadMedia = (
         });
       };
       const setupNativeStreamInfo = async () => {
-        return updateStreamInfoFromSrc(src, mediaEl, type)
+        return updateStreamInfoFromSrc(src, mediaEl, type, props as HTMLMediaElement)
```
luwes: revert this

packages/playback-core/src/index.ts
luwes: use mediaEl to dispatch the event on for now. not ideal either but that will be known in mux-video-react as well...

packages/playback-core/src/index.ts
luwes: how do you mean?

it's to get the `playbackId` from the `src`

packages/playback-core/src/index.ts
```diff
@@ -1497,3 +1515,52 @@ const getErrorFromHlsErrorData = (
...
+    .then((json) => {
+      const metadata = extractMetadata(json);
+
+      const eventUpdateMetadata = new CustomEvent('muxmetadata', { bubbles: true, composed: true, detail: metadata });
+      el.dispatchEvent(eventUpdateMetadata);
+
+      if (metadata['com.mux.video.branding'] === 'mux-free-plan') {
+        const eventSetDefaultLogo = new Event('setdefaultlogo', { bubbles: true, composed: true });
+        el.dispatchEvent(eventSetDefaultLogo);
+      }
```
luwes: remove this in playback-core. it's more related to UI and fits better in the higher level abstractions like mux-video or mux-video-react

packages/playback-core/src/index.ts
```diff
@@ -1497,3 +1515,52 @@ const getErrorFromHlsErrorData = (
   mediaError.data = data;
   return mediaError;
 };
+
+export const fetchMetadata = (el: HTMLMediaElement, metadataUrl: string) => {
```
luwes: remove the argument `el` in this function. just return the metadata and handle the dispatch logic after an await fetchMetadata

packages/mux-video/src/index.ts
```diff
@@ -105,53 +107,63 @@ class MuxVideoBaseElement extends CustomVideoElement implements Partial<MuxMedia
...
+  `;
   }
   constructor() {
     super();
     this.#defaultPlayerInitTime = generatePlayerInitTime();
+
+    this.addEventListener('setdefaultlogo', (e) => {
+      this.#logo = 'default';
+      this.updateLogo();
+    });
```
luwes: remove this, the logic inside should go in the muxmetadata listener

packages/playback-core/src/index.ts
```diff
@@ -1497,3 +1515,52 @@ const getErrorFromHlsErrorData = (
...
+      for (const [key, value] of Object.entries(first)) {
+        if (typeof value === 'string' || typeof value === 'number') {
+          metadata[key] = String(value);
+        }
+      }
+    }
+  }
+
+  return metadata;
+};
```
luwes, line 1577: please remove this function and do that for each file separately.
in mux-video there is no need for a transformation.

the fetch in playback-core just this part 

```js
      for (const item of json[0].metadata) {
        if (item.key && item.value) {
          metadata[item.key] = item.value;
        }
      }
```

PR #1131, opened 2025-05-28
fix: rendition-menu visual improvements

### LINKED ISSUE #1122: rendition-menu jankiness
we should set a min-width on this menu because it's looks pretty janky at the moment.
also don't make the checkmark (and font-weight) cause a jump in text. 
it should be like the default in Media Chrome.

see at https://www.mux.com/player

Image
Image
Image

Pintorado:
Closes #1122


luwes (approved):
thanks @Pintorado Feels a lot better!

PR #1132, opened 2025-05-29
fix: Ensure ref callback cleanup executes on MuxPlayer unmount

### LINKED ISSUE #1112: Bug: Cleanup function in ref callback not executed on MuxPlayer unmount

mux-player-react

Chrome

None

### Description

cleanup function not triggred when MuxPlayer unmount

```tsx
<MuxPlayer
  ref={() => {
    console.log("mounted");
    return () => console.log("unmounted");
  }
/>
```

```
// console
mounted
```

### Reduced test case

_No response_

### Steps to reproduce

1.
2.
3.

### Current Behavior

not logged in console "unmounted" when unmounted 
```
// console
mounted
```

### Expected Behavior

should be logged in console "unmounted" when unmounted 

```
// console
mounted
unmounted
```

### Errors

_No response_

_No response_

Pintorado:
Fixes #1112 

## Changes
- Modified useCombinedRefs hook to collect cleanup functions returned by the function refs and ensure they are called.
- Updated the toNativeProps utility to explicitly exclude the ref prop, ensuring that React handles the ref mechanism directly.


packages/mux-player-react/src/useCombinedRefs.ts
```diff
@@ -13,15 +13,24 @@ export const useCombinedRefs: useCombinedRefs = (...refs) => {
...
         ref.current = targetRef.current;
       }
     });
+
+    return () => {
+      cleanupFunctions.forEach((cleanup) => cleanup());
+    };
   }, [refs]);
 
   return targetRef;
```
luwes: this looks very good! 

doing some quick search on the internet I did notice some cleanup is missing for the React 18 and under case. 
let's replace this complete function with this implementation 
https://gist.github.com/KurtGokhan/9aafd8e83c9bc6a2946fe2dc7f2c1d19#file-use-combined-refs-new-ts

it takes both the old and new React in account completely.

Pintorado:
It still seems to work only for React 19. I'm testing it with React 18.3.1 and the following code but I don't see the 'unmounted' log.
``` tsx
import MuxPlayer from '@mux/mux-player-react';

export default function App() {
  return (
    <MuxPlayer
      playbackId="a4nOgmxGWg6gULfcBbAa00gXyfcwPnAFldF8RdsNyk8M"
      ref={() => {
        console.log('mounted');
        return () => console.log('unmounted');
      }}
    />
  );
}
```

Also I had to change the `else` in the `applyRef` function since I was getting a typescript error
``` tsx
else {
    (refA as { current: TElement }).current = current
    return () => {
      (refA as { current: TElement | null }).current = null
    }
```

Pintorado:
Testing it like this works for both versions:
``` jsx
import MuxPlayer from '@mux/mux-player-react';
import { useEffect, useRef } from 'react';

export default function App() {
  const playerRef = useRef(null);

  useEffect(() => {
    console.log('mounted');
    return () => console.log('unmounted');
  }, []);

  return (
    <MuxPlayer
      playbackId="a4nOgmxGWg6gULfcBbAa00gXyfcwPnAFldF8RdsNyk8M"
      ref={playerRef}
    />
  );
}
```

Pintorado:
Supports new and old versions now. For React <19 it logs the error correctly:
image

luwes (approved):
Great! thanks @Pintorado

PR #1139, opened 2025-06-13
Fix/newsplayer event cbs

cjpillsbury:
- Fixes some event cases for the ads video element around play/playing
- Removes code that invokes play on play event in ads video element
- Adds event callbacks to mux news player


luwes (approved):
LGTM

luwes (approved):

PR #1140, opened 2025-06-16
feat: add free plan logo

luwes:
fix #33
closes #1130

- [x] add support for native HLS (iOS or Safari w/ preferPlayback="native")

https://elements-demo-nextjs-git-free-plan-logo-mux.vercel.app/MuxPlayerFreeLogo


ronald-urbina:
LGTM

packages/playback-core/src/index.ts
```diff
@@ -1104,7 +1156,16 @@ export const loadMedia = (
 
   if (mediaEl && shouldUseNative) {
     const type = getType(props);
+
     if (typeof src === 'string') {
+      // Fetch the Mux metadata JSON even on preload=none because it's needed for the Mux logo.
+      if (src.endsWith('.mp4') && src.includes(MUX_VIDEO_DOMAIN)) {
+        const playbackId = toPlaybackIdFromSrc(src);
+        const domain = props.customDomain || MUX_VIDEO_DOMAIN;
```
cjpillsbury: nitpick(blocking): I think the condition should be `src.includes(props.customDomain)` and the fallback should be unnecessary if we are defaulting `customDomain` appropriately. See, e.g.: https://github.com/muxinc/elements/blob/main/packages/playback-core/src/index.ts#L325

packages/playback-core/src/util.ts
```diff
@@ -204,3 +204,19 @@ class IntlMessageFormat {
...
 }
+
+export const fetchMetadata = async (url: string) => {
+  const resp = await fetch(url);
+  if (!resp.ok) throw new Error(`Failed to fetch metadata: ${resp.status} ${resp.statusText}`);
+  const json = await resp.json();
+
+  const metadata: Record<string, string> = {};
+
+  for (const item of json[0].metadata) {
```
cjpillsbury: thought(non-blocking): Fine for now, but we should probably make this just return the JSON chapter data (with types) and move the plucking of the 0th entry elsewhere. This is not urgent so no need to do anything for the time being.

packages/playback-core/src/index.ts
```diff
@@ -644,6 +689,13 @@ export const setupHls = (
       ..._hlsConfig,
     }) as HlsInterface;
 
+    hls.on(Hls.Events.MANIFEST_PARSED, async function (_event, data) {
```
cjpillsbury, line 736: question(blocking): should this be a `once` since we don't recycle hls.js instances? Also, should this have teardown/destroy logic (e.g. if preload is none and the source/playback id is updated)? If yes, update. Otherwise, feel free to resolve the comment/ask.

packages/playback-core/src/index.ts
```diff
@@ -77,31 +78,42 @@ export const toDRMTypeFromKeySystem = (keySystem: string): DRMTypeValue | undefi
...
+  // See https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-12#section-4.4.6.4
+  const sessionDataLines = playlist.split('\n').filter((line) => line.startsWith('#EXT-X-SESSION-DATA'));
+  if (!sessionDataLines.length) return {};
+
+  const sessionData: Record<string, string> = {};
+
+  for (const line of sessionDataLines) {
+    const payload = line.split('#EXT-X-SESSION-DATA:')[1]?.trim();
+    if (!payload) continue;
+    const [dataIdAttribute, valueAttribute] = payload.split(',');
```
cjpillsbury: issue(blocking): These are named tag attributes, not ordered, so we can't/shouldn't assume Mux Video will return them in a particular order or that we won't add other `#EXT-X-SESSION-DATA` attributes later.

cjpillsbury (changes requested):
a few callouts but overall this is lookin' 😎

packages/playback-core/src/index.ts
luwes: I think this is correct. Teardown is not done for hls.js event listeners in playback core because the instance is not recycled. I feel better if this is a `on` instead of `once` because we're making an assumption that the event is only called once. Maybe in the future we call loadSource more than once in playback-core.

packages/playback-core/src/index.ts
luwes: thanks for catching this, my bad

packages/playback-core/src/index.ts
```diff
@@ -77,31 +77,53 @@ export const toDRMTypeFromKeySystem = (keySystem: string): DRMTypeValue | undefi
...
+  const sessionDataLines = playlist.split('\n').filter((line) => line.startsWith('#EXT-X-SESSION-DATA'));
+  if (!sessionDataLines.length) return {};
+
+  const sessionData: Record<string, { VALUE: string }> = {};
+
+  for (const line of sessionDataLines) {
+    const payload = line.split('#EXT-X-SESSION-DATA:')[1]?.trim();
+    if (!payload) continue;
+
+    const attrs = payload.split(',');
```
cjpillsbury: thought(non-blocking): There's still a possibility of parsing errors here, though unlikely so not treating as a blocker. quoted string attribute values may also contain commas. To account for this, you can use e.g.

```ts
const SessionDataAttributeNames = ['DATA-ID', 'VALUE', 'URI', 'FORMAT', 'LANGUAGE'];
const toNamedAttributeKVStrings = (attributesString: string, attributeNames: string[]) =>
  attributesString.split(',').reduce((attrKVStrs, commaSplitStr) => {
    const trimmedCommaSplitStr = commaSplitStr.trim();
    if (attributeNames.some((attrName) => trimmedCommaSplitStr.startsWith(`${attrName}=`))) {
      attrKVStrs.push(trimmedCommaSplitStr);
    } else {
      const prevAttrKVStr = attrKVStrs[attrKVStrs.length - 1];
      // NOTE: Concat then trim to avoid mutating quoted string values. This will still
      // have an edge case if the value is ATTR-NAME="my    , quoted string value", but we
      // can resolve for that if/when it is an issue.
      attrKVStrs[attrKVStrs.length - 1] = (prevAttrKVStr + commaSplitStr).trim();
    }
    return attrKVStrs;
  }, [] as string[]);

const attrs = toNamedAttributeKVStrings(payload, SessionDataAttributeNames);
```

cjpillsbury (approved):
Approved with one callout for an improvement on hls attribute parsing.

PR #1145, opened 2025-06-23
fix(playback-core): prefer uri value for chapters session metadata when available.

cjpillsbury:


packages/playback-core/src/index.ts
```diff
@@ -110,12 +110,15 @@ export const getMultivariantPlaylistSessionData = (playlist: string) => {
...
-  // This handles quoted values that may contain commas
-  const regex = /(\w+[-\w]*)="([^"]*)"/g;
-  const matches = [...str.matchAll(regex)];
-  return Object.fromEntries(matches.map(([, key, value]) => [key, value]));
+const QUOTE_CHAR = '"';
+const isQuotedString = (value: string) =>
+  value.indexOf(QUOTE_CHAR) === 0 && value.lastIndexOf(QUOTE_CHAR) === value.length - 1;
+// matches all HLS attribute name=value pairs, with or without quotes, using per spec rules
+// for matching AttributeName (See: https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-17#section-4.2)
+const ATTR_LIST_REGEX = /([A-Z0-9-]+)=(".*?"|.*?)(?:,|$)/g;
```
luwes: why is the manual `QUOTE_CHAR` and `isQuotedString()` / slice logic needed? 
that could be done by the regex no?

cjpillsbury: because `EXT-X-SESSION-DATA` has the `FORMAT` attribute, which is an unquoted string that may be present in these tags.

cjpillsbury: updated per @luwes out of band recommendation.

PR #1146, opened 2025-06-23
fix: support URI and VALUE for session data

luwes:
this is for the hls.js case, the native hls case is already handled here https://github.com/muxinc/elements/pull/1145


cjpillsbury (approved):
🚀

PR #1167, opened 2025-07-14
feat(astro): add native Astro support with `MuxPlayer` and `MuxUploader` components

ascorbic:
### Summary

This adds official Astro support to Mux Elements\! I've created two new packages that wrap the existing web components:
- `@mux/mux-player-astro` - Video player for Astro
- `@mux/mux-uploader-astro` - Uploader for Astro

These are basically thin Astro component wrappers around the existing `@mux/mux-player` and `@mux/mux-uploader` web components, with proper TypeScript support and native Astro integration.

For the APIs I've kept as close to the web components as possible, with changes mostly to be closer to Astro idioms and some inspiration from the React components. The biggest difference is ptobably the theme handling, where I added some tricks to ensure everything is loaded in the right place.

The READMEs are more comprehensive than the others, and some of that could probably be moved to the docs.

### What's New

**@mux/mux-player-astro**
- Native Astro video player component
- Full TypeScript support with proper type definitions for props
- All the existing mux-player features work
- Built-in theme system (classic, gerwig, microvideo, minimal, news)
- Separate `/ads` export
- Proper prop validation and utilities

**@mux/mux-uploader-astro**
- Native Astro uploader component  
- TypeScript support
- Full integration with Mux upload API
- Progress tracking and error handling

### Technical Details

- Works with Astro 4.0+ and 5.0+ (peer dependency)
- ESM-only exports with proper module resolution
- Modular theme exports for easy customization
- Built on top of the existing web components, so they inherit all the same functionality

### Usage

```astro
---
import { MuxPlayer } from '@mux/mux-player-astro';
import { MuxUploader } from '@mux/mux-uploader-astro';
---

<\!-- Video Player -->
<MuxPlayer
  playbackId="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"
  metadata={{
    video_title: 'Big Buck Bunny',
    viewer_user_id: 'user-id-1234',
  }}
/>

<\!-- Uploader -->
<MuxUploader
  endpoint="https://my-upload-endpoint.com/upload"
  id="my-uploader"
/>
```

### Updated Example Site

I've updated the existing Astro example site to use these new native components instead of the React ones it was using before. All the demo pages now showcase the `@mux/mux-player-astro` and `@mux/mux-uploader-astro` components with examples of themes, advanced features (chapters, cuepoints, renditions, audio tracks), ads, and different upload scenarios. These are mostly lifted from the web component or React examples.

### Questions/Future Considerations

Not sure if separate Astro components for `mux-audio` and `mux-video` are needed too, or if `mux-player` covers those use cases sufficiently. Would love feedback on this.

### Breaking Changes
None - this is purely additive and doesn't affect any existing packages.

### DIFF
--- a/packages/mux-player-astro/README.md
+++ b/packages/mux-player-astro/README.md
@@ -62,15 +62,16 @@ import { MuxPlayer } from '@mux/mux-player-astro';
 
 ## With Themes
 
+You can use one of the built-in themes by passing the theme name as a string to the `theme` prop:
+
 ```astro
 ---
 import { MuxPlayer } from '@mux/mux-player-astro';
-import classic from '@mux/mux-player-astro/themes/classic';
 ---
 
 <MuxPlayer
   playbackId="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"
-  theme={classic}
+  theme="classic"
   style={{
     display: 'block',
     aspectRatio: '16/9',
@@ -80,12 +81,12 @@ import classic from '@mux/mux-player-astro/themes/classic';
 
 Available themes:
 - `classic` - Classic player theme
-- `minimal` - Minimal player theme  
+- `minimal` - Minimal player theme
 - `microvideo` - Microvideo theme for short-form content
 - `gerwig` - P, pretty, I, intelligent, N, never sad, K, cool
 - `news` - News theme
 
-Alternatively you can use a [Media Chrome theme](https://www.mux.com/docs/guides/player-themes#media-chrome-themes) by passing the theme name as a string:
+Alternatively you can use a [Media Chrome theme](https://www.mux.com/docs/guides/player-themes#media-chrome-themes) by passing the theme name as a string and including the theme `<template>` element with that ID in your page:
 
 ```astro
 ---
@@ -101,7 +102,7 @@ import { MuxPlayer } from '@mux/mux-player-astro';
 </template>
 <MuxPlayer
   playbackId="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"
-  mediaChromeTheme="tiny-theme"
+  theme="tiny-theme"
 />
 ```
 
@@ -121,21 +122,21 @@ import { MuxPlayer } from '@mux/mux-player-astro';
 
 <script>
   import type { MuxPlayerElement } from '@mux/mux-player-astro';
-  
+
   const player = document.getElementById('my-player') as MuxPlayerElement;
-  
+
   player.addEventListener('play', (event) => {
     console.log('Player started playing!');
   });
-  
+
   player.addEventListener('pause', (event) => {
     console.log('Player paused!');
   });
-  
+
   player.addEventListener('timeupdate', (event) => {
     console.log('Current time: ', player.currentTime);
   });
-  
+
   player.addEventListener('ended', (event) => {
     console.log('Video ended!');
   });
--- a/packages/mux-player-astro/package.json
+++ b/packages/mux-player-astro/package.json
@@ -17,8 +17,7 @@
   ],
   "exports": {
     ".": "./src/index.ts",
-    "./ads": "./src/ads/index.ts",
-    "./themes/*": "./src/themes/*/index.ts"
+    "./ads": "./src/ads/index.ts"
   },
   "peerDependencies": {
     "astro": "^4.0.0 || ^5.0.0"
@@ -28,6 +27,7 @@
     "@mux/playback-core": "0.30.1"
   },
   "devDependencies": {
-    "astro": "^5.11.0"
+    "astro": "^5.11.0",
+    "vite": "^6.0.0"
   }
 }
--- a/packages/mux-player-astro/src/utils.ts
+++ b/packages/mux-player-astro/src/utils.ts
@@ -1,7 +1,12 @@
+/// <reference types="vite/client" />
+import type { MuxVideoTheme } from './types';
+
 const toKebabCase = (string: string) =>
   string.replace(/[A-Z]/g, (match) => `-${match.toLowerCase()}`).replaceAll('_', '-');
 
 const FLATTEN_PROPS = new Set(['metadata', 'tokens', 'castCustomData']);
+
+export const themes = import.meta.glob<{ default: MuxVideoTheme }>('./themes/*/index.ts');
 
 function stringifyValue(value: any): string {
   if (Array.isArray(value)) {
@@ -14,6 +19,22 @@ function stringifyValue(value: any): string {
     return '';
   }
   return String(value);
+}
+
+export async function getThemeObject(theme: string | MuxVideoTheme | undefined): Promise<MuxVideoTheme> {
+  if (!theme) {
+    return {};
+  }
+  if (typeof theme !== 'string') {
+    return theme;
+  }
+  const themeModule = themes[`./themes/${theme}/index.ts`];
+  if (!themeModule) {
+    return {
+      name: theme,
+    };
+  }
+  return (await themeModule()).default;
 }
 
 export const toNativeAttributes = (props: { [key: string]: any } = {}): Record<string, string> => {
--- a/examples/astro/src/pages/mux-player-classic-theme.astro
+++ b/examples/astro/src/pages/mux-player-classic-theme.astro
@@ -1,7 +1,6 @@
 ---
 import Layout from '../layouts/Layout.astro';
 import { MuxPlayer } from '@mux/mux-player-astro';
-import classic from '@mux/mux-player-astro/themes/classic';
 ---
 
 <Layout>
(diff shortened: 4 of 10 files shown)


sarah11918:
So excited for this! And also noting that if we have Astro-specific components, the [/video-for/astro/](https://www.mux.com/video-for/astro) page should probably be updated, too, as it makes sense for the example to show these components instead of the `<mux-player>`/`<mux-uploader>` components.

packages/mux-player-astro/README.md
```diff
@@ -0,0 +1,237 @@
...
+  <media-controller>
+    <slot name="media" slot="media"></slot>
+    <media-control-bar>
+      <media-play-button></media-play-button>
+    </media-control-bar>
+  </media-controller>
+</template>
+<MuxPlayer
+  playbackId="DS00Spx1CV902MCtPj5WknGlR102V5HFkDe"
+  mediaChromeTheme="tiny-theme"
```
luwes:
```suggestion
  theme="tiny-theme"
```

packages/mux-player-astro/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "astro-component"
+  ],
+  "type": "module",
+  "files": [
+    "src"
+  ],
+  "exports": {
+    ".": "./src/index.ts",
+    "./ads": "./src/ads/index.ts",
+    "./themes/*": "./src/themes/*/index.ts"
```
luwes: neat, if this is sufficient?

packages/mux-player-astro/README.md
luwes: Ah I see that's a new prop in the Astro component. Could it still be named `theme` for a string theme name as well?

luwes:
Seems the lockfile is not in sync? failing CI

packages/mux-player-astro/README.md
ascorbic: The reason I did them as separate props is to make it clearer that they behave differently: if you pass a string you need to manually include the theme template yourself, and if you want the automatic handling you need to pass an import. I could combine them, but I'm worried about people trying to set e.g. `theme="minimal"` rather than `theme={minimal}` and wondering why nothing happens.

packages/mux-player-astro/package.json
ascorbic: Yeah, subpath patterns do a direct string replacement of the matched pattern in the source with the `*` in the target

packages/mux-player-astro/README.md
ascorbic: @sarah11918 I wonder if you have thoughts on the DX here. The way it's set up now is that if you import a theme like e.g. `import classic from "@mux/mux-player-astro/themes/classic"` and pass that to `theme={classic}` it handles all the theme setup, styles etc. If you want a custom theme you construct it like I show in this example, and pass the id as a string. Would it be better to have a single prop `theme` that could handle both, or would it be confusing that if you pass a string you need to also define the template?

sarah11918: Ooh, interesting situation!

It *would* be confusing to have a single `theme` prop that works differently depending on the kind of theme (available from Mux vs a Chrome Media Theme). But admittedly, it could also feel confusing to have two props that both do the same thing (set the theme to use). I'm not sure there's one obviously correct choice here from the standpoint of telling a good story about how this works. But I lean towards the two props.

The first thing that comes to mind when I think of one theme prop that *sometimes* requires an import is, after "now I have to keep straight when I do or don't need an import", is "Why does the thing Mux has actually built require an import, but a custom theme I made wouldn't? Shouldn't using a Mux included theme be *easier* than using something else?" (I'm assuming each of these players are separate imports so that you're not importing more than you need, but that's what I'd be thinking when learning about this dual-purpose prop.)

That's one advantage to keeping the props separate: not only do they work differently, and you have to write different code in each case, but also, now you're making people learn about *both* ways to set a theme even if (and before they know) they only need one. I imagine choosing from the available themes might be a much more common way to get started, in which case someone only needs to consider a Chrome Media Theme if they want to do something more advanced/custom. So having a second prop for this particular user story also feels fine to me, instead of having one `theme` prop need to cover both cases. (I would also be thinking about what specific docs links would I want to have available to me, and it seems like "using a custom theme" is something I'd want to be able to send people to *directly*, so documenting it separately would more easily let me do that than if they need to know *all* about the one `theme` prop.)

So, if this were something *I* were responsible for documenting, I'd probably be leaning towards two entirely different props for simplicity of story.

But, I also like to think of wild ideas, so I might also ask my devs if there's a way to pass options to a player,  to e.g. have an additional "included" player theme choice that is `custom` and THAT option allows you to specify a Chrome Media Theme. So you're *always* importing a player from Mux and setting a `theme` in one place (maybe that's wasteful? but the user story is consistent) and only in the `custom` player case, passing a Chrome Media Theme option as a string to it. 😅 . But of course, that's even more code than you'd need to do the same thing right now, so solutions like that would be only if you value having one theme prop and a consistent way to set a theme (custom or not). But, I do like thinking about these things! (That doesn't mean these thoughts should be entertained, though.)

ascorbic: The built-in ones are easier – you just need to import them. For the non-built-in ones you need to define the whole theme template by hand, which is certainly harder than just importing one.

dylanjha:
This is very cool to see!

packages/mux-player-astro/README.md
luwes: Thanks for thinking with us about these tough decisions, naming and these kind of intrcacies are always hard.
I took another look at how this has been handled for the web component and React component and the main method is setting a string to the `theme` prop / attribute.

You see this export of the React theme returns a string 
https://github.com/muxinc/elements/blob/main/packages/mux-player-react/src/themes/gerwig.ts

Would that be possible with Astro?

ascorbic:
@luwes You mean remove all the exports and automatic handling of the scripts?

packages/mux-player-astro/README.md
ascorbic: We need a way to tell Astro to load it in the client, which is why I set up the exports with the component too. React runs in the browser, so can just do the side effect import, but Astro is server-rendered so we need to inject a script tag to load the theme.

luwes: not sure I follow, can't this be changed? from 

```js
import type { MuxVideoTheme } from '../../types';
import component from './gerwig.astro';
export default {
  component,
  name: 'gerwig',
} satisfies MuxVideoTheme;
```

to 

```js
import type { MuxVideoTheme } from '../../types';
import component from './gerwig.astro';
export default 'gerwig';
```

or something in those lines

ascorbic: No, because just importing the component isn't enough: it needs to be added to the page as a script tag or it won't be available in the browser. That's why I'm exporting an object with both the component and the name, so the player component can add it to the page

ascorbic: Look at `packages/mux-player-astro/src/MuxPlayer.astro` to see how I'm handling it

ascorbic: OK, I've had a bit of a think and worked out a way that I think is the best of both worlds. You now pass in a string for built-in themes too, and if it's supported it will load the component automatically. I've updated the readme, and @luwes if you're happy with this approach we'll need to update the docs too. cc @sarah11918

packages/mux-player-astro/src/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+    return theme;
+  }
+  const themeModule = themes[`./themes/${theme}/index.ts`];
+  if (!themeModule) {
+    return {
+      name: theme,
+    };
+  }
+  return (await themeModule()).default;
+}
```
ascorbic, line 38: @luwes Here's the change. I used `import.meta.glob` to dynamically import the theme modules.

packages/mux-player-astro/src/utils.ts
```diff
@@ -0,0 +1,68 @@
+/// <reference types="vite/client" />
+import type { MuxVideoTheme } from './types';
+
+const toKebabCase = (string: string) =>
+  string.replace(/[A-Z]/g, (match) => `-${match.toLowerCase()}`).replaceAll('_', '-');
+
+const FLATTEN_PROPS = new Set(['metadata', 'tokens', 'castCustomData']);
+
+export const themes = import.meta.glob<{ default: MuxVideoTheme }>('./themes/*/index.ts');
```
ascorbic, line 9: Load available themes

packages/mux-player-astro/README.md
luwes: 👍 Nice, this looks good to me! This way we keep the external API more uniform across frameworks.

Yes, the main worry was that we were adding a new prop only for Astro. 

I'm totally fine with requiring an explicit import of the theme, that's how the web and React component work as well but if it's automatically handled for the internal themes even better.

sarah11918: Nice job, Matt! That looks great, and feels pretty darn ergonomic.

PR #1181, opened 2025-08-06
feat: Add switch to control the behavior of the ended event

### LINKED ISSUE #952: Bug: Set currentTime close to the end of video - video will end

mux-player-react

Chrome

macOS

### Description

This is simple video player. No styling, nothing - default progress bar

Try to click to 11 or 12 second - video will fire onEnd event.
If I want to change it using `currentTime = value `- result the same
But if you don’t change time - video plays well

### Reduced test case

https://codesandbox.io/p/sandbox/mux-player-change-time-dp6k5p?file=%2Fsrc%2FPlayer.js

### Steps to reproduce

1. click to 11 or 12 second

### Current Behavior

video will fire onEnd event

### Expected Behavior

currentTime should be changed to new value

### Errors

During video loading I see this (not when I change time)

image

2.8.0

Pintorado:
Closes #952 

This PR adds a prop to disable dispatching the ended event when the `currentTime` is close to the end of the video.

## Changes
- Add `disablePseudoEnded` prop in mux-player
- This prop is passed to mux-video by adding it to the template
- Dispatch a custom event that checks for this attribute before dispatching the `ended` event. 
- Updates type definitions in mux-player and mux-player-react to include the new prop
- Add prop to the MuxPlayer example

Can be tested with `playbackId="AzhcksAuZkwtmjsSe8ViifJh2rtwhtDhSu1Oh1Wm3mk"` using mux-player-react


harshmandan:
Not sure if it's related but please make sure that the `disablePseudoEnded` behaviour switch is also available in the `mux-video` element. Thank you!

Pintorado:
Yes! I was planning to add that, thanks for pointing it out!

packages/mux-video/src/base.ts
```diff
@@ -167,6 +167,11 @@ export class MuxVideoBaseElement extends CustomVideoElement implements IMuxVideo
         this.updateLogo();
       }
     });
+
+    this.addEventListener('mux-check-disable-pseudo-ended', (event: Event): void => {
+      (event as CustomEvent<{ shouldDisable: boolean }>).detail.shouldDisable =
+        this.hasAttribute('disable-pseudo-ended');
+    });
```
luwes: this should be removed and we should add a get / set disablePseudoEnded prop in this class...
then in playback-core we can access that prop

packages/playback-core/src/index.ts
```diff
@@ -1175,6 +1175,18 @@ export const loadMedia = (
...
+    // Check if disablePseudoEnded is enabled by dispatching a custom event
+    const checkEvent = new CustomEvent('mux-check-disable-pseudo-ended', {
+      bubbles: true,
+      composed: true,
+      cancelable: true,
+      detail: { shouldDisable: false },
+    });
+
+    mediaEl.dispatchEvent(checkEvent);
+    if (checkEvent.detail.shouldDisable) return;
```
```suggestion
    if (props.disablePseudoEnded) return;
```

luwes: something like that ^^, fix types etc.

packages/mux-video/src/base.ts
Pintorado: Done!

packages/playback-core/src/index.ts
Pintorado: Done!

luwes (approved):
LGTM 👍

Issue #1174, opened 2025-07-24
How to remove box-shadow from center-controls?


kumardeepakme:
How to remove drop-shadow from the parent SHADOW element "center-controls"?
```JSX
<div slot="centered-chrome" class="center-controls pre-playback">
  <media-play-button part="center play button pre-play" tabindex="0" role="button" aria-label="play" mediapaused="">
    <!-- SVG Icon -->
  </media-play-button>
</div>
```

I tried CSS Parts - but it only targets `media-play-button part="center play button pre-play"`::

```CSS
mux-player::part(center play button) {
  shadow: none;
  filter: none;
}
```

ronald-urbina:
@kumardeepakme Thanks for bringing this up! We’ve worked on a solution for this, and it will be available in the next release of elements.

PR #1183, opened 2025-08-12
feat(themes): expose centered layout

ronald-urbina:
Resolves [Issue 1174](https://github.com/muxinc/elements/issues/1174?reload=1?reload=1) by:

•	Add to the centered-chrome <div> the part centered-layout in both classic and gerwig themes.
•	Exposed this centered-layout element for styling and customization.

Usage: 

Add style to remove filter to centered-layer children as 

`mux-player::part(centered-layer) {
  filter: none;
}`

### DIFF
--- a/packages/mux-audio/package.json
+++ b/packages/mux-audio/package.json
@@ -77,7 +77,7 @@
     "downlevel-dts": "^0.11.0",
     "esbuild": "^0.25.1",
     "eslint": "^9.22.0",
-    "hls.js": "~1.6.6",
+    "hls.js": "~1.6.13",
     "npm-run-all": "^4.1.5",
     "replace": "^1.2.2",
     "shx": "^0.4.0",
--- a/packages/mux-elements-codemod/tsconfig.json
+++ b/packages/mux-elements-codemod/tsconfig.json
@@ -19,7 +19,7 @@
     "esModuleInterop": true,
     "skipLibCheck": false,
     "forceConsistentCasingInFileNames": true,
-    "typeRoots": ["node_modules/@types"]
+    "types": ["node"]
   },
   "include": ["src/**/*"]
 }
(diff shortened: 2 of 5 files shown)


packages/mux-player/src/themes/classic/classic.html
```diff
@@ -637,7 +637,7 @@
           <media-control-bar part="control-bar top" slot="top-chrome">
             {{>TitleDisplay}}
           </media-control-bar>
-          <div slot="centered-chrome" class="center-controls">
+          <div slot="centered-chrome" part='centered-layer' exportparts="centered-layer" class="center-controls">
```
luwes: this should be reverted.

the `exportparts` goes on the web component which has the shadow dom which contains the CSS part so for this case it's the media-controller element but as you can see it already exports `centered-layer`.

maybe it's not exported on `media-theme` ?

luwes (changes requested):
needs a correct css export

packages/mux-player/src/themes/gerwig/gerwig.html
```diff
@@ -1032,7 +1032,12 @@
     <template if="!audio">
       <!-- Pre-playback UI -->
       <!-- same for both on-demand and live -->
-      <div slot="centered-chrome" class="center-controls pre-playback">
+      <div
+        slot="centered-chrome"
+        part="centered-layer"
+        exportparts="centered-layer"
+        class="center-controls pre-playback"
+      >
```
luwes: revert as well

packages/mux-player/src/themes/classic/classic.html
ronald-urbina: I revert those changes and move the filter property to the button so the reporter can use to

`mux-player::part(center play button) {
  shadow: none;
  filter: none;
}`

to remove the filter (shadow) to the button

PR #1190, opened 2025-08-27
Fix: Remove error dialog for audio only

### LINKED ISSUE #1189: Bug: Error Dialog causes layout issues for audio only

mux-player

Other (add details below)

Other (add details below)

### Description

General issue. Quick fix solution: for `<template if="...">` audio rendering cases of themes, simply remove the error dialog.

Would be good to check player.style for similar issues as well.

### Reduced test case

_No response_

### Steps to reproduce

1.
2.
3.

### Current Behavior

Layout gets janky.

### Expected Behavior

Layout doesn't get so janky.

### Errors

_No response_

_No response_

Pintorado:
Fixes #1189 

## Changes
Error dialog is no longer rendered if the media is an audio.


luwes (approved):
LGTM 👍

PR #1216, opened 2025-10-21
feat: typed CSSProperties for mux-player custom CSS vars

### LINKED ISSUE #1198: Feature Request: Properly typed CSSProperties for mux-player custom CSS vars

mux-player-react

### Description

In https://www.mux.com/docs/guides/player-customize-look-and-feel#controls-sections, we see that you can hide `mux-player` controls with custom CSS like:

```js
<mux-player
  style="--center-controls: none; --top-captions-button: none;"
></mux-player>
```

This is fine when using JS, but in TypeScript, the `--center-controls` (or any of the [available custom CSS vars used by mux-player](https://www.mux.com/docs/guides/player-customize-look-and-feel#available-css-variables)) are not recognized as valid CSS properties.

That's because the type [`VideoApiAttributes`](https://github.com/muxinc/elements/blob/7f9656fdc962b11a79fc05769dcc5982bb9a96b0/packages/mux-player-react/src/types.ts#L33) exports its `style` as the base React `CSSProperties`. It doesn't have any knowledge of the custom CSS vars at all and you have to `@ts-ignore` or `@ts-expect-error` it to get it to work:

Image

### Expected Behavior

Probably there should be a `MuxPlayerProperties extends CSSProperties` intermediate definition, with something like:

```typescript
interface MuxPlayerProperties extends CSSProperties {
        "--controls"?: "none"; // or null, or whatever else this can be
        "--dialog"?: "none";
        // etc
    }
```

ignasio-mux:
This PR closes #1198 

Created MuxPlayerCSSProperties interface that extends CSSProperties with support for any CSS custom property starting with --


N1Knight (approved):

packages/mux-player-react/src/types.ts
```diff
@@ -130,3 +141,14 @@ export type MuxPlayerProps = {
   onChapterChange?: GenericEventListener<MuxPlayerElementEventMap['chapterchange']>;
 } & Partial<MuxMediaPropTypes> &
   Partial<VideoApiAttributes>;
+
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'mux-player': React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement> & {
```
luwes: Hmm, what about all the other supported props. `src`, `muted`, etc. Are those types still intact?

ignasio-mux: I decided to restore JSX IntrinsicElements for the <mux-player> web component. I tried using React.HTMLAttributes<MuxPlayerElement>, but during testing I realized it allows any random attribute (like nacho="true"), which makes TypeScript checks pointless.

There doesn’t seem to be a simple way to only allow valid HTML, data-*, aria-*, and known component attributes at the same time.

I think this PR still makes sense since it fixes the issue for the React component.

arcataroger:
Hello @luwes and @ignasio-mux,

Thank you for this fix. However, given [this](https://github.com/ignasio-mux/elements/blob/ab5bfc98f469f52e916105798a96bcdced8090bb/packages/mux-player-react/src/types.ts):

```js
/**
 * CSS properties interface for Mux components
 * Extends standard CSSProperties to include custom CSS variables
 */
export interface MuxCSSProperties extends CSSProperties {
  [key: `--${string}`]: string | undefined;
}
```

Isn't that essentially the same as `@ts-ignore`? Done that way, there's no actual type hinting, and no guarantee that a property either exists or is properly used.

Is there no way to individually enumerate the [actually-supported custom CSS props and values](https://www.mux.com/docs/guides/player-customize-look-and-feel#available-css-variables), e.g.:

`"--controls"?: 'none'` or `"--controls-backdrop-color"?: CSSProperties["color"]`

If that is too much work, I understand... at least this PR will stop the Typescript errors. Thank you either way!

Issue #853, opened 2023-12-14
Feature Request: official event for pausing and starting preloading 


szymonhernik:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player, mux-player-react

### Description

I wish there was an option that stops preloading/buffering after you start the video. This would be useful in cases like when the video leaves the screen stop fetching the video data. Currently, I found a walkaround this by doing `_hls.stopLoad()` and `_hls.stopLoad()` on the player but I don't think this is official solution. 

This has been discussed here as well:
https://github.com/muxinc/elements/discussions/422

### Expected Behavior

Stop and start prerendering video data after calling a function.

mmcc:
Yeah, agreed, having to dig into the underlying playback tech directly to do this doesn't feel right. Thanks for the feedback, let us think on what the path could look like here.

spuppo-mux:
Hello, thanks for raising this. The current workaround is fine to use for now; we are going to take this into account for future versions. You may also want to look into `Hls.pauseBuffering()` and `Hls.resumeBuffering()` ([reference](https://hlsjs.video-dev.org/api-docs/hls.js.hls)). I do want to note that these functions behave a bit differently on Safari and Chrome with `mse` playback.

PR #1218, opened 2025-10-22
Feat: add support to enable/disable HLS buffering

N1Knight:
Resolves https://github.com/muxinc/elements/issues/853 by adding startBuffering, stopBuffering and isBuffering methods for mux-player.

Also, adding new examples for buffering feature usage.


spuppo-mux (approved):
Tested this, LGTM.

For future reference, we use both (`pauseBuffering`, `resumeBuffering`) and (`stopLoad`, `startLoad`), because those sets of functions behave a bit differently between Safari and Chrome (both with hls.js). Mainly we note that `bufferingEnabled` is set to true after seeking only on Safari, but remains false (as expected) in Chrome. Only using `startLoad` and `stopLoad` would also behave incorrectly on Chrome when seeking.

luwes:
Wait with this one please, we've generally done this with the preload attribute in the past. 
Not sure if we want to expose this fine grained of an API.

spuppo-mux:
Closing this as discussed in this week's office hours.

PR #1219, opened 2025-10-22
fix: override player minmax

cjpillsbury:


packages/playback-core/src/index.ts
```diff
@@ -715,6 +715,9 @@ export const setupHls = (
           contentId: metadata?.video_id,
         }
       : undefined;
+
+    const capLevelControllerObj = _hlsConfig.capLevelToPlayerSize ? {} : { capLevelController: MinCapLevelController };
```
luwes: Didn't this need an explicit check for strict `false` or `true`?
Or we will only disable MinCapLevelController on a truthy value

luwes (approved):
LGTM 👍

PR #1220, opened 2025-10-23
feat: Add internationalization support to MuxUploader

### LINKED ISSUE #1149: Feature Request: Change the texts of MuxUploader (intl)

mux-uploader

### Description

Hi,

I need to use the **<MuxUploader />** for my app but I need to translate the texts (I use next-intl but it souldn't matters).

Would it be possible to add some props to the **<MuxUploader />** component to change the texts that are displayed ? Or maybe there is another way I didn't find.

Image

Thanks you.

### Expected Behavior

Either something like this

```tsx
<MuxUploader locale="fr" />
```

or this

```tsx
<MuxUploader texts={{upload: t('upload'), please_wait: t('please_wait') ...}} />
```

Pintorado:
Closes #1149 

This PR adds the `locale` attribute to mux-uploader to specify the required language.

**Supported locales:**
- `'en'` (default if not set)
- `'es'` (Spanish)
- `'fr'` (French) 
- `'de'` (German)
- `'pt'` (Portuguese)
- `'it'` (Italian)
- `'zh'` (Chinese)

**Features:**
- Automatic fallback to English for unsupported locales
- Browser language detection when no locale is specified

### DIFF
--- a/packages/mux-video/src/base.ts
+++ b/packages/mux-video/src/base.ts
@@ -857,6 +857,35 @@ export class MuxVideoBaseElement extends CustomVideoElement implements IMuxVideo
           this.updateLogo();
         }
         break;
+      case Attributes.DISABLE_TRACKING: {
+        if (newValue == null || newValue !== oldValue) {
+          const currentTime = this.currentTime;
+          const paused = this.paused;
+          this.unload();
+          this.#requestLoad().then(() => {
+            this.currentTime = currentTime;
+            if (!paused) {
+              this.play();
+            }
+          });
+        }
+        break;
+      }
+      case Attributes.DISABLE_COOKIES: {
+        if (newValue == null || newValue !== oldValue) {
+          const disabled = this.disableCookies;
+          if (disabled) {
+            document.cookie.split(';').forEach((c) => {
+              if (c.trim().startsWith('muxData')) {
+                document.cookie = c
+                  .replace(/^ +/, '')
+                  .replace(/=.*/, '=;expires=' + new Date().toUTCString() + ';path=/');
+              }
+            });
+          }
+        }
+        break;
+      }
     }
   }
 


packages/mux-uploader/src/mux-uploader-drop.ts
```diff
@@ -197,6 +200,20 @@ class MuxUploaderDropElement extends globalThis.HTMLElement {
       opts
     );
   }
+
+  updateText() {
+    const locale = (this.#uploaderEl as any)?.locale || 'en';
```
spuppo-mux: Should we add locale to the MuxUploaderElement interface to avoid using `as any`

Pintorado: Yes, I'll do this, thank you!

packages/mux-uploader/package.json
```diff
@@ -49,6 +49,7 @@
     "clean": "shx rm -rf dist/ && shx rm -rf src/polyfills",
     "test": "web-test-runner **/*test.js --port 8005 --coverage --config test/web-test-runner.config.mjs --root-dir ../..",
     "posttest": "replace 'SF:src/' 'SF:packages/mux-uploader/src/' coverage/lcov.info --silent",
+    "i18n": "npm run build:esm -- --keep-names && i18n-utils dist/index.mjs ./lang",
```
spuppo-mux: Running `npm run i18n` modifies all language json files. Mainly removing the end line, but most importantly for en.json, it changes all values to it's keys. For example, it leads to this:
```
{
  "dropText": "dropText",
  "separatorText": "separatorText",
  "uploadCompleteText": "uploadCompleteText",
  "retryText": "retryText",
  "pausingText": "pausingText",
  "resumeText": "resumeText",
  "pauseText": "pauseText",
  "uploadButtonText": "uploadButtonText",
  "noEndpointError": "noEndpointError"
}
```
Looking at the way this is implemented in `playback-core`, the keys are already defined the same as the text (ex: `"Network Error": "Network Error",` in en.json). I guess this would make the code clearer to read when an i18n call appears  (ex: `i18n("Network Error");` - Not the best example, but with longer texts would make more sense).
 
 _If_ we want to follow that approach, we could change the jsons to the format:
 ```
{
  "Drop a video file here to upload": "Drop a video file here to upload",
  "or": "or",
  "Upload complete!": "Upload complete!",
  "Retry": "Retry",
  "Pausing...": "Pausing...",
  "Resume": "Resume",
  "Pause": "Pause",
  "Upload a video": "Upload a video",
  "No url or endpoint specified -- cannot handleUpload": "No url or endpoint specified -- cannot handleUpload"
}
```

File: packages/mux-uploader/src/utils/i18n.ts [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+const translationsLanguages: Record<string, TranslateDictionary> = {
+  en: En,
+  es: Es,
+  fr: Fr,
+  de: De,
+};
+
+export const addTranslation = (langCode: string, languageDictionary: TranslateDictionary) => {
+  translationsLanguages[langCode] = languageDictionary;
+};
```
spuppo-mux, line 25: Here we should also add the langCode into the supportedLocales array so getEffectiveLocale works with new languages

PR #1226, opened 2025-10-28
fix: avoid prerender errors caused by date access

### LINKED ISSUE #1225: Feature Request: Next.js 16 `cacheComponents` support

mux-video-react, mux-player-react, mux-audio-react

### Description

NextJS 16 has an option called `cacheComponents` that changes the cache behavior. With it enabled, any access to uncached data must be wrapped in `Suspense` or marked as cacheable.

Presumably, Mux elements access date internally during their initialization which causes Next to throw an error:
```
## Error Type
Console Error

## Error Message
Route "/example" used `Date.now()` inside a Client Component without a Suspense boundary above it. See more info here: https://nextjs.org/docs/messages/next-prerender-current-time-client


    at MuxVideoExample (app/mux-video.tsx:30:6)
    at Page (app/example/page.tsx:15:27)
    at LogSafely (<anonymous>:null:null)

## Code Frame
  29 | 				{video?.data?.playbackId && (
> 30 | 					<MuxVideo
     | 					^
  31 | 						autoPlay
  32 | 						loop
  33 | 						muted

Next.js version: 16.0.1 (Turbopack)
```

This can of course be solved by wrapping the video in a Suspense tag, but that is not always desirable.

### Expected Behavior

Mux video components should [only access the time in the browser](https://nextjs.org/docs/messages/next-prerender-current-time-client#only-access-the-time-in-the-browser) to avoid errors when pre-rendering

RJWadley:
fixes #1225


packages/playback-core/src/index.ts
```diff
@@ -372,6 +372,10 @@ const shouldDefaultToMSE = (mediaEl: Pick<HTMLMediaElement, 'canPlayType'>) =>
   isAndroidLike || isChromeWithNativeHLS(mediaEl);
 
 export const generatePlayerInitTime = () => {
+  // bail during SSR to avoid triggering prerender errors
+  // actual time will be generated during hydration
+  if (typeof window === 'undefined') return 0;
```
```suggestion
  if (typeof window === 'undefined') return undefined;
```

Thanks for the contribution! 

I'm a little worried that a zero here could result in miss calculation down the line in case it actually gets set to Mux data. 

`undefined` is safer as the value wouldn't be used in Mux data...

PR #1244, opened 2025-11-25
fix: upgrade hls.js to 1.6.15

Pintorado:
This PR upgrades `hls.js` to the latest version across all element packages. 
https://github.com/video-dev/hls.js/releases/tag/v1.6.15

Fixes audio track issues on Safari/iOS: selecting the Default track makes the video freeze without being able to unpause it for some videos.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14487,12 +14487,6 @@
         "tslib": "^2.0.3"
       }
     },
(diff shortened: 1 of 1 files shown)


luwes:
Also upgrade the package lock file

PR #1245, opened 2025-11-25
fix: Properly setup DRM to support Airplay

spuppo-mux:
Closes [#183](https://github.com/muxinc/devextravaganza/issues/183)

I also reorganized the `onFpEncrypted` callback to get a better understanding of the code and compare it with the [FairPlay Streaming Overview](https://developer.apple.com/streaming/fps/FairPlayStreamingOverview.pdf).

The main issue here was that we were ignoring the Apple TV SPC when generating the DRM license. This is reflected on the `session.addEventListener('message', ...)` changes. We were listening just once to this message, but more recent versions of Apple TV send new messages when using AirPlay. Therefore I modified that listener to ask for that license again and update the session for each new message.

Note for reviewers: Please pay extra attention to the way I reformatted this code (not sure if I should change some Promise.reject), the new error messages and/or if I should be removing the event listener at some point.


packages/playback-core/src/index.ts
```diff
@@ -859,132 +859,111 @@ export const setupNativeFairplayDRM = (
...
+      console.error('Failed to generate license request', e);
+      const message = i18n(
+        'Failed to generate a DRM license request. This may be an issue with the player or your protected content.'
+      );
+      const mediaError = new MediaError(message, MediaError.MEDIA_ERR_ENCRYPTED, true);
+      mediaError.errorCategory = MuxErrorCategory.DRM;
+      mediaError.muxCode = MuxErrorCode.ENCRYPTED_GENERATE_REQUEST_FAILED;
+      saveAndDispatchError(mediaEl, mediaError);
+    });
+    session.addEventListener('message', async (event) => {
```
spuppo-mux: This is the main big change

packages/playback-core/src/index.ts
```diff
@@ -996,16 +975,41 @@ export const setupNativeFairplayDRM = (
         }
         return Promise.reject(errOrResp);
       });
-      await session.update(response).catch(() => {
-        // eslint-disable-next-line no-shadow
+
+      // This is the same call whether we are local or AirPlay.
+      // Safari will forward CKC to Apple TV automatically.
+      await session.update(ckc).catch(() => {
```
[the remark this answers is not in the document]
spuppo-mux: I realized I left this halfway commited, are we ok with logging these errors to console like this @luwes? Any message format you suggest? Previously this error was being ignored as we create a custom error below

packages/playback-core/src/index.ts
luwes: could you move this addEventListener above the `session.generateRequest()`? I know it's async but just to be safe

packages/playback-core/src/index.ts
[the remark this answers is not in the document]
luwes: I think what you have is good. 
Don't accept this Copilot suggestion.

packages/playback-core/src/index.ts
```diff
@@ -859,153 +859,179 @@ export const setupNativeFairplayDRM = (
...
           'Failed to update DRM license. This may be an issue with the player or your protected content.'
         );
         const mediaError = new MediaError(message, MediaError.MEDIA_ERR_ENCRYPTED, true);
         mediaError.errorCategory = MuxErrorCategory.DRM;
         mediaError.muxCode = MuxErrorCode.ENCRYPTED_UPDATE_LICENSE_FAILED;
-        return Promise.reject(mediaError);
+
+        saveAndDispatchError(mediaEl, mediaError);
       });
+    };
```
luwes: This is not right yet. Use try / catch blocks here and add a `return` at the end of the catch or you can also do nested try / catch blocks like this: 

```js
const onMessage = async (event: MediaKeyMessageEvent) => {
  const spc = event.message;
  
  try {
    const ckc = await getLicenseKey(spc, toLicenseKeyURL(props, 'fairplay'));
    
    try {
      // This is the same call whether we are local or AirPlay.
      // Safari will forward CKC to Apple TV automatically.
      await session.update(ckc);
    } catch {
      const message = i18n(
        'Failed to update DRM license. This may be an issue with the player or your protected content.'
      );
      const mediaError = new MediaError(message, MediaError.MEDIA_ERR_ENCRYPTED, true);
      mediaError.errorCategory = MuxErrorCategory.DRM;
      mediaError.muxCode = MuxErrorCode.ENCRYPTED_UPDATE_LICENSE_FAILED;
      
      saveAndDispatchError(mediaEl, mediaError);
    }
  } catch (errOrResp) {
    if (errOrResp instanceof Response) {
      const mediaError = getErrorFromResponse(errOrResp, MuxErrorCategory.DRM, props);
      console.error('mediaError', mediaError?.message, mediaError?.context);
      
      if (mediaError) {
        saveAndDispatchError(mediaEl, mediaError);
        return;
      }
      
      const error = new Error('Unexpected error in license key request');
      saveAndDispatchError(mediaEl, error);
      return;
    }
    
    saveAndDispatchError(mediaEl, errOrResp);
  }
};
```

spuppo-mux: Adding this I saw that for non MediaErrors (i.e. Unexpected error and errOrResp) the function saveAndDispatchError just ignores it so I guess I can just ignore those cases? Maybe log them?

packages/playback-core/src/index.ts
```diff
@@ -859,158 +859,184 @@ export const setupNativeFairplayDRM = (
...
+          if (mediaError) {
+            saveAndDispatchError(mediaEl, mediaError);
+            return;
+          }
+
+          console.error('Unexpected error in license key request', errOrResp);
+          return;
+        }
+
+        console.error(errOrResp);
```
luwes, line 1011: 👍

luwes (approved):
LGTM 👍

PR #1248, opened 2025-11-27
fix: Upgrade ce-la-react

spuppo-mux:
Relates to https://github.com/muxinc/next-video/issues/397 and https://github.com/muxinc/ce-la-react/pull/7

Upgrades ce-la-react to prevent missing key warning when using custom components on SSR.


luwes (approved):
LGTM

PR #1252, opened 2025-12-03
fix: button text for file-select slot can now be customized

Pintorado:
## Changes

- Prevent translating the text if the file picker button is a custom one.
- Added a test for this case


packages/mux-uploader/src/mux-uploader-file-select.ts
luwes: this can be simplified if you keep the code in the else block.


```suggestion
const buttonEl = this.shadowRoot?.querySelector('#file-select') as HTMLButtonElement;
if (buttonEl) {
  buttonEl.textContent = text;
}
```

PR #1251, opened 2025-12-03
feat: Expose capLevelToPlayerSize property 

spuppo-mux:
Closes https://github.com/muxinc/elements/issues/1238

The purpose of this PR is to expose [`hls.capLevelToPlayerSize`](https://github.com/video-dev/hls.js/blob/master/docs/API.md#capleveltoplayersize) property while maintaining the current behavior (if nothing is set, `hls.capLevelToPlayerSize = true` and we use our `MinCapLevelController` instead of HLS's default controller).

Added two new attributes `cap-level-to-player-size` and `disable-cap-level-to-player-size`; the latter mainly to be used on React components.

Web Components (`<mux-video>` and `<mux-player>` )
- `cap-level-to-player-size`
  - If present will set `hls.capLevelToPlayerSize` when the HLS element is initialized using its own default `CapLevelController`. Note: If present will be true unless set to the string `"false"`.
  - If not present, `hls.capLevelToPlayerSize` is set to true and `MinCapLevelController` is used instead.
- `disable-cap-level-to-player-size` If present (and not set to `"false"`), will have the same behavior as setting `cap-level-to-player-size="false"`. Takes precedence over `cap-level-to-player-size`.

React Components (`MuxVideo` and `MuxPlayer`)
- Note: On the React Components setting `reactAttribute={false}` is equivalent to setting `react-attribute=undefined` on the Web Components.
- `capLevelToPlayerSize : boolean | undefined`
  - If present (`true`): sets `hls.capLevelToPlayerSize` to **`true`** and sets HLS's default CapLevelController. 
  - If not present (`false` or `undefined`): `hls.capLevelToPlayerSize` is set to true and `MinCapLevelController` is used instead.
- `disableCapLevelToPlayerSize : boolean | undefined`
  - If present (`true`): sets `hls.capLevelToPlayerSize` to **`false`** and sets HLS's default CapLevelController. 
  - If not present (`false` or `undefined`): Does nothing.

For both cases (React and Web Components), getting the `capLevelToPlayerSize` property will return:
- `false` if disableCapLevelToPlayerSize=true;
- whatever is set on the `_hlsConfig.capLevelToPlayerSize` if that value is defined;
- whether that element's `cap-level-to-player-size` attribute !== `"false"` otherwise.

PR #1263, opened 2026-01-15
feat: cap-rendition-to-player-size

### LINKED ISSUE #1238: Feature Request: Expose a `capLevelToPlayerSize` property and attribute on Mux media elements

mux-video, mux-video-react, mux-player, mux-player-react, playback-core

### Description

Mux Player and related currently rely on hls.js for HLS media playback unless using native HLS playback support. In those cases, hls.js has a built in [`capLevelToPlayerSize`](https://github.com/video-dev/hls.js/blob/master/docs/API.md#capleveltoplayersize) config property.

By default, we currently set this to [`true`](https://github.com/muxinc/elements/blob/main/packages/playback-core/src/index.ts#L714) and use a [custom cap level controller](https://github.com/muxinc/elements/blob/main/packages/playback-core/src/min-cap-level-controller.ts) that will never cap the level below 720p.

Recently, we added some additional logic so that if a user of Mux Player (et al.) explicitly sets `capLevelToPlayerSize` via the `_hlsConfig` (see, e.g. [`<mux-player>`](https://github.com/muxinc/elements/blob/main/packages/mux-player/src/base.ts#L1660-L1676) or [`<mux-video>`](https://github.com/muxinc/elements/blob/main/packages/mux-video/src/base.ts#L742-L748)), we will [not use our custom cap controller](https://github.com/muxinc/elements/blob/main/packages/playback-core/src/index.ts#L729-L730).

The feature request is simply to make setting `capLevelToPlayerSize` easier from a developer experience perspective by exposing and documenting a `capLevelToPlayerSize` property / `cap-level-to-player-size` attribute on:

* `<MuxPlayer />` (react)
* `<mux-player>` (web component)
* `<MuxVideo />` (react)
* `<mux-video>` (web component)

Be sure to update the developer documentation markdown files in the repository as well.

### Expected Behavior

* When `capLevelToPlayerSize` is explicitly set concurrently with or before setting `src` or `playbackId` (to either `true` or `false`), apply 
* When `capLevelToPlayerSize` is not explicitly set, use current default player
* If `capLevelToPlayerSize` is set after setting `src` or `playbackId`, expected behavior is not defined, but no errors should occur and playback should still work
* If `capLevelToPlayerSize` is set and native playback is being used (either because that is the default playback preference for an environment like Safari or `preferPlayback="native"` and native playback is available), expected behavior is not defined, but no errors should occur and playback should still work.

As an implementation suggestion, you may want to use a pattern that is similar to how we handle Mux Data metadata, where we automatically add component-level properties to the metadata object for a single "source of truth" (see, e.g. [here](https://github.com/muxinc/elements/blob/main/packages/mux-player/src/base.ts#L1646-L1655).

cjpillsbury:
Exposes hls.js's `capRenditionToPlayerSize` option with a Mux-optimized default.

Closes #1238 (based off of initial effort from @spuppo-mux in #1251)

### Behavior

| Value | Behavior |
|-------|----------|
| `undefined` (default) | Caps to player size, 720p minimum floor |
| `true` | Standard hls.js capping (may go below 720p) |
| `false` | No capping |

### Test plan

- [ ] `npm run dev` → http://localhost:3000/MuxPlayer
- [ ] Toggle "Cap Level To Player Size" between None/true/false
- [ ] Verify no console errors
- [ ] Verify config: `document.querySelector('mux-player').media.nativeEl._hls.config.capLevelToPlayerSize`
- [ ] Trigger a rendition switch/re-selection via ABR - e.g. open Chrome Dev Tools, resize player size, seek backwards during playback to prompt segment reloading, and confirm that rendition does(n't) get capped under different permutations

### DIFF
--- a/packages/mux-player/test/player.test.js
+++ b/packages/mux-player/test/player.test.js
@@ -1,6 +1,4 @@
 import { fixture, assert, aTimeout, waitUntil, oneEvent, nextFrame } from '@open-wc/testing';
-import { CapLevelController } from 'hls.js';
-import { MinCapLevelController } from '@mux/playback-core';
 import '../src/index.ts';
 
 const isSafari = /.*Version\/.*Safari\/.*/.test(navigator.userAgent);
@@ -1097,10 +1095,10 @@ describe('<mux-player> seek to live behaviors', function () {
       await waitUntil(() => playerEl.media?._hls, 'hls.js instance should be created');
 
       assert.equal(playerEl.media._hls.config.capLevelToPlayerSize, true, 'should default to true');
-      assert.equal(
-        playerEl.media._hls.config.capLevelController,
-        MinCapLevelController,
-        'should use MinCapLevelController'
+      // MinCapLevelController has a static minMaxResolution property, standard CapLevelController does not
+      assert.isDefined(
+        playerEl.media._hls.config.capLevelController.minMaxResolution,
+        'should use MinCapLevelController (has minMaxResolution property)'
       );
     });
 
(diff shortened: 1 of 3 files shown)


luwes (approved):

heff (approved):
API updates look good, thanks 👍

Issue #1261, opened 2026-01-13
Bug: Mux Player DRM AirPlay failure on iOS 26.1+


flpymonkey:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player-react

### Which browsers are you using?

Safari

### Which operating systems are you using?

iOS

### Description

The Mux Player fails to AirPlay DRM videos on iOS 26.1+.

When I tried it with the Shaka player, I was able to get the DRM content to work by using legacy Apple Media Keys as described here: [FairPlay Shaka Player guide](https://github.com/shaka-project/shaka-player/blob/main/docs/tutorials/fairplay.md)

Is there a way to force the Mux Player to use these legacy media keys to work around this?

### Steps to reproduce

1. Set up a DRM enabled video in Mux following this guide: [here](https://www.mux.com/docs/guides/protect-videos-with-drm)
2. On an iOS 26.1+ device, play a DRM-enabled video via the Mux Player or Mux React Player 
3. AirPlay the video using the built in Mux Player AirPlay button
4. For DRM enabled videos, AirPlay will fail in Safari for iOS 26.1+

### Current Behavior

The player always fails when I AirPlay a DRM-enabled Mux video on iOS 26.1+

### Expected Behavior

The video should be allowed to AirPlay without the DRM error.

### Errors

Mux Error:
Error: Code 5
failed to generate a drm license request. this may be an issue with the player or your protected content.

### What version of the package are you using?

v3.10.2

spuppo-mux:
Hi, thanks for reporting! Coincidentally this is an issue we were already investigating internally.

Just added a PR https://github.com/muxinc/elements/pull/1266 to try to fix it.

If you have time, would really appreciate if you could help test it with your setup. 

The testing process should be the same as you describe on your "Steps to Reproduce" section, but adding a `'use-webkit-fairplay'` attribute to your `<mux-player/>` element or `useWebkitFairplay` on React (We may go for another attribute name before merging) and of course building from the PR branch on my fork.

jkheianna:
Hi, hope you're all doing well!

On a project I'm working on we faced the same issue as reported. 
We've tested the PR #1266 fix as @spuppo-mux described, building the player from it and casting a DRM video through AirPlay works fine again! - just to let you know our results if it helps. 

Thanks!

flpymonkey:
[...]

PR #1266, opened 2026-01-21
fix: Added option to fallback to WebKit FairPlay Streaming setup.

spuppo-mux:
Fixes https://github.com/muxinc/elements/issues/1261

This PR addresses a bug in Apple devices on newer OS versions when attempting to cast protected media using AirPlay. If the sending device is on OS version 26.1 or 26.2, playback will fail if when casting because of a `NotSupportedError` on `session.generateRequest` call.

Added `useWebkitFairplay` property and 'use-webkit-fairplay' attribute to fallback to webkit prefixed FairPlay Streaming DRM setup. (`com.apple.fps.1_0` key system)

Also added `drmSetupFallback` property, which will be called when the specific call fails with a `NotSupportedError`. This property can be set to `undefined` if the user wants to avoid fallbacks for any reason or override it to choose a different approach. The default fallback consists on setting `useWebkitFairplay` and reloading the player.

### DIFF
--- a/packages/mux-video/src/base.ts
+++ b/packages/mux-video/src/base.ts
@@ -807,10 +807,17 @@ export class MuxVideoBaseElement extends CustomVideoElement implements IMuxVideo
   drmSetupFallback = async () => {
     this.useWebkitFairplay = true;
 
-    // TODO: Can we somehow wait till load is done before unloading?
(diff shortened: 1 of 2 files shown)


spuppo-mux:
Closed, reworked on https://github.com/muxinc/elements/pull/1277

PR #1277, opened 2026-02-18
fix: Add WebKit FairPlay DRM fallback for AirPlay on newer OS versions

spuppo-mux:
Fixes https://github.com/muxinc/elements/issues/1261

This PR addresses a bug in Apple devices on newer OS versions when attempting to cast protected media using AirPlay. If the sending device is on OS version 26.1 or 26.2, playback will fail if when casting because of a `NotSupportedError` on `session.generateRequest` call.

# Summary
- Extracted EME FairPlay DRM logic from `index.ts` into `eme-fairplay.ts` module.
- Added `webkit-fairplay.ts`, a legacy FairPlay implementation using WebKit-prefixed APIs (`WebKitMediaKeys`, `webkitneedkey`, etc) to work around bug.
- Added fallback mechanism on DRM setup:
    - when EME `session.generateRequest` fails with `NotSupportedError` during AirPlay, the player tears down the EME session and re-initializes using the legacy WebKit FairPlay setup.
- To work around having to teardown and initialize the core, we now store the PlaybackCore reference in `muxMediaState` (and get it via new `getCoreReference()`).
    - Refactored `mux-video` and `mux-audio` to derive `#core` from `muxMediaState` instead of owning it as a private field, ensuring they always see the latest core after a DRM fallback re-initialization.
    
## New module: `eme-fairplay.ts`
- Contains `setupEmeNativeFairplayDRM()` the standard EME-based FairPlay flow (previously inlined in `index.ts`).
- Main changes done to this function (did them in different commits to visualize them more clearly):
    - Added comments detailing steps and setups, and reorganized code to follow the order of those steps
    - Returns a cleanup function (and uses that same function on teardown)
    - Catches NotSupportedError on session.generateRequest during wireless playback and triggers `fallbackToWebkitFairplay()`
    
## New module: `webkit-fairplay.ts`
- Contains `setupWebkitNativeFairplayDRM()`, the legacy WebKit-prefixed FairPlay flow
- Mirrors (as close as possible) the EME flow structure but uses `WebKitMediaKeys`, `webkitneedkey`, `webkitkeymessage`, and `webkitkeyerror` events

## Changes in index.ts
- `setupNativeFairplayDRM()` now branches between EME and WebKit paths based on internal `props.useWebkitFairplay`
- Shared config (certificate/license fetching, error handling, DRM type callback) is extracted into `commonConfig`
`fallbackToWebkitFairplay` callback on `initialize()` tears down, sets `useWebkitFairplay = true`, re-initializes, then restores playback position and state
- coreReference added to `muxMediaState` so components always get the current core

## Changes in mux-video / mux-audio
- `#core` changed from a stored field to a getter that reads from `getCoreReference(this.nativeEl)`, keeping it in sync after fallback re-initialization

### DIFF
--- a/packages/playback-core/src/webkit-fairplay.ts
+++ b/packages/playback-core/src/webkit-fairplay.ts
@@ -157,20 +157,21 @@ export const setupWebkitNativeFairplayDRM = ({
     // If this becomes an issue, we should implement better cleanup.
     // For example separating event cleanup from session.close
     const newTeardown = () => {
+      // Remove all event listeners first to prevent events firing during cleanup
       // @ts-ignore
       session.removeEventListener('webkitkeymessage', onWebkitKeyMessage);
       // @ts-ignore
       session.removeEventListener('webkitkeyerror', onWebkitKeyError);
       mediaEl.removeEventListener('teardown', newTeardown);
+      if ('webkitCurrentPlaybackTargetIsWireless' in mediaEl) {
+        mediaEl.removeEventListener('webkitcurrentplaybacktargetiswirelesschanged', newTeardown);
+      }
+
+      teardownSession = null;
 
       try {
         session.close();
       } catch {}
-
-      if ('webkitCurrentPlaybackTargetIsWireless' in mediaEl) {
-        mediaEl.removeEventListener('webkitcurrentplaybacktargetiswirelesschanged', newTeardown);
-      }
-      teardownSession = null;
     };
 
     if ('webkitCurrentPlaybackTargetIsWireless' in mediaEl) {
@@ -187,12 +188,16 @@ export const setupWebkitNativeFairplayDRM = ({
   };
 
   const teardownWebkit = () => {
-    teardownSession?.();
-
-    wkMediaEl.webkitSetMediaKeys(null);
+    // Remove event listeners first to prevent new key requests during cleanup
     // @ts-ignore
     mediaEl.removeEventListener('webkitneedkey', onWebkitNeedKey);
     mediaEl.removeEventListener('teardown', teardownWebkit);
+
+    teardownSession?.();
+
+    try {
+      wkMediaEl.webkitSetMediaKeys(null);
+    } catch {}
   };
 
   // @ts-ignore


packages/playback-core/src/index.ts
```diff
@@ -603,6 +614,33 @@ export const initialize = (props: Partial<MuxMediaPropsInternal>, mediaEl: HTMLM
...
+
+    props.useWebkitFairplay = true;
+
+    // Prevent duplicate mux data
+    let muxDataKeepSession = props.muxDataKeepSession;
+    props.muxDataKeepSession = true;
+
+    const oldCore = muxMediaState.get(mediaEl)?.coreReference;
+    // This will tear oldCore down internally
+    initialize(props, mediaEl, oldCore);
```
cjpillsbury, line 630: praise: 😎 smart.

packages/playback-core/src/index.ts
```diff
@@ -624,11 +662,18 @@ export const initialize = (props: Partial<MuxMediaPropsInternal>, mediaEl: HTMLM
...
   const setAutoplay = setupAutoplay(props as Pick<MuxMediaProps, 'autoplay'>, mediaEl, nextHlsInstance);
 
-  return {
+  const newCore = {
     engine: nextHlsInstance,
     setAutoplay,
     setPreload,
   };
+
+  const state = muxMediaState.get(mediaEl);
```
cjpillsbury, line 672: praise: 👌

packages/mux-video/src/base.ts
```diff
@@ -796,12 +800,11 @@ export class MuxVideoBaseElement extends CustomVideoElement implements IMuxVideo
   }
 
   load() {
-    this.#core = initialize(this as Partial<MuxMediaProps>, this.nativeEl, this.#core);
+    initialize(this as Partial<MuxMediaProps>, this.nativeEl, this.#core);
```
cjpillsbury, line 803: **NOTE:** Like other playback-core state, `#core` has been refactored to use our media map pattern in playback-core. This was based on needs of teardown+reinitialize for DRM edge cases. See `getCoreReference` for details.

packages/playback-core/src/eme-fairplay.ts
```diff
@@ -0,0 +1,277 @@
...
+    mediaEl.removeEventListener('encrypted', onFpEncrypted);
+    mediaEl.removeEventListener('teardown', teardownEme);
+    if (teardownSession) {
+      await teardownSession();
+    }
+    await mediaEl.setMediaKeys(null).catch(() => {});
+  };
+
+  mediaEl.addEventListener('encrypted', onFpEncrypted);
+  mediaEl.addEventListener('teardown', teardownEme, { once: true });
```
cjpillsbury, line 274: praise: 😎

packages/playback-core/src/index.ts
```diff
@@ -877,190 +923,72 @@ export const getLicenseKey = async (message: ArrayBuffer, licenseServerUrl: stri
...
+      return Promise.reject(errOrResp);
+    });
 
-      await setupMediaKeySession(initDataType, initData);
-      // @ts-ignore
-    } catch (error: Error | MediaError) {
-      saveAndDispatchError(mediaEl, error);
-      return;
-    }
+  const commonConfig = {
```
cjpillsbury, line 969: praise: 👌

packages/playback-core/src/index.ts
```diff
@@ -877,190 +923,72 @@ export const getLicenseKey = async (message: ArrayBuffer, licenseServerUrl: stri
...
+    getLicenseKey: getLicenseKeyHandler,
+    saveAndDispatchError,
+    drmTypeCb: () => {
+      props.drmTypeCb?.(DRMType.FAIRPLAY);
+    },
   };
 
-  addEventListenerWithTeardown(mediaEl, 'encrypted', onFpEncrypted);
+  if (props.useWebkitFairplay) {
+    setupWebkitNativeFairplayDRM(commonConfig);
```
cjpillsbury, line 981: question(blocking): Shouldn't we expect an equivalent teardown for `setupWebkitNativeFairplayDRM`?

packages/playback-core/src/webkit-fairplay.d.ts
```diff
@@ -0,0 +1,55 @@
+declare global {
```
cjpillsbury, line 1: praise: 👌

packages/playback-core/src/webkit-fairplay.ts
```diff
@@ -0,0 +1,268 @@
+/// <reference types="./webkit-fairplay.d.ts" />
+
+import { MediaError, MuxErrorCategory, MuxErrorCode } from './errors';
+import { i18n } from './util';
+
+interface WebkitNativeFairplayConfig {
```
cjpillsbury: question(non-blocking): This interface/type should be identical for both webkit-fairplay and eme-fairplay implementations, right? Very low priority/not a blocker, but, if I'm correct, you may want to consider having a single, shared type definition.

packages/playback-core/src/webkit-fairplay.ts
```diff
@@ -0,0 +1,268 @@
...
+
+  /**
+   * Listener for "webkitneedkey" event.
+   * Will be called when we receive EXT-X-KEY tags and will setup media keys and session.
+   */
+  const onWebkitNeedKey = (ev: WebkitNeedKeyEvent) => {
+    // note: Promises only resolve once, so we wrap it in an anonymous funciton to use as an event listener
+    const handle = async () => {
+      try {
+        // Setup 1 - set up the keys based on the data type (currently only "skd" for Native Safari FPS DRM)
```
cjpillsbury, line 56: praise: Love that you added equivalent "steps" code comments for the webkit impl! 😍

packages/playback-core/src/index.ts
spuppo-mux: Well, here I am open to suggestions:

- On one hand it looked a bit cleaner (to me) to set teardown event inside the setup function. Also, it's nice to also remove teardown function on teardown - also felt cleaner to do both adding and removal close to each other (see the end of the setup functions).
- On the other hand, I can see how having only one of the setup functions return a teardown function can lead to confusion. 

At first I was setting teardown outside the setup function and adding {once: true} on the addEventListener, this is generally enough, except for when the teardown is ran outside of an event listener (i.e. in the fallback). Anyway running teardown a second time would not be too much of a problem so I think we can opt for whatever feels cleaner.

packages/playback-core/src/webkit-fairplay.ts
spuppo-mux: Tried to make them as similar as possible. Only difference is that EME interface also has a fallback. Maybe I can export from EME, make the fallback optional, and import from Webkit, since the latter will eventually be removed.

packages/mux-video/src/base.ts
spuppo-mux: Should I add this as a comment somewhere in the code?

cjpillsbury: Not a bad idea, but not a blocker. We should make sure it's captured somewhere(s) just so other maintainers aren't surprised (code comments, release notes, etc). Wherever you think is fit.

packages/playback-core/src/webkit-fairplay.ts
cjpillsbury: Non-blocking, so don't fret if this is going to be onerous. I'll defer to you on what you think is best here.

packages/playback-core/src/index.ts
cjpillsbury: I think as long as we are making sure teardown works for other cases (unloading media, replacing the playback-id with a new one, etc), we should be good, though consistency is also useful for making sense of the code (especially for others less familiar with it). Since you've also been digging into memory leak/memory management issues, that would be one (but not the only) reason we want to just make sure teardown works as expected generally. Defer to you on smoke testing and final decisions here.

packages/playback-core/src/webkit-fairplay.ts
spuppo-mux: Pushed a change with the common interface

packages/playback-core/src/index.ts
spuppo-mux: Given that generally our setup functions don't return a teardown function I opted for just pointing out the difference with a comment. 

For the memory leak work we should choose an approach to follow and can change the way this is set up/torn down depending on what we choose.

PR #1326, opened 2026-04-29
fix: defer load() until connected to prevent metadata race on first Mux Data beacon

### LINKED ISSUE #1199: Mux player WC metadata race condition
Reported internally.

When the web component is defined before the parsing of the custom element the metadata attributes are not reflected immediately in Mux data. This is troublesome for the video-id because only the first view / request captures that value.

This is a race condition bug.

Example: https://codesandbox.io/p/sandbox/lively-dawn-d37l6j

R-Delfino95:
## Problem

Fixes #1199 

When `<mux-player>` (or `<mux-video>`) is created programmatically and attributes are set before the element is appended to the DOM — with any async yield between attribute setting and appendChild (e.g. await Promise.resolve(), React's async reconciliation, Next.js rendering, etc.) — the first Mux Data analytics beacon is sent with an incorrect `video_id`.

Root cause: `mux-video.#requestLoad()` schedules `load()` in a microtask via `await Promise.resolve()`. Under the following sequence, that microtask fires before `mux-player.connectedCallback` has a chance to propagate metadata-* attributes to the inner `<mux-video>`:

```
const el = document.createElement('mux-player');
el.setAttribute('playback-id', '...');
el.setAttribute('metadata-video-id', 'my-custom-video-id'); // not in observedAttributes
await Promise.resolve(); // ← load() microtask fires here — metadata is empty
document.body.appendChild(el);  // connectedCallback propagates metadata too late
```
metadata-* attributes are not in mux-video's observedAttributes. They are read lazily from the element's own DOM attributes by the metadata getter. Since mux-player's template does not forward metadata-* attributes to the inner <mux-video>, propagation only happens in `mux-player.connectedCallback`. Any async gap before appendChild causes the first `mux.monitor()` call — which sets `video_id` permanently for the view — to receive empty metadata. Because video_id is fixed at view start and cannot be corrected retroactively, the beacon is permanently wrong.

## Fix
Added an `isConnected` guard inside mux-video's `#requestLoad()`

When the element is not yet connected at microtask time, the deferred `load()` is skipped. The existing `connectedCallback` already re-triggers `#requestLoad()` once the element connects (guarded by `!this.#core`), so `load()` runs after `mux-player.connectedCallback` has propagated metadata

---

> [!NOTE]
> **Medium Risk**
> Touches custom element lifecycle and overrides `setAttribute`/`removeAttribute`, which can subtly affect attribute semantics and timing-sensitive load/analytics behavior.
> 
> **Overview**
> Fixes a race where `metadata-*` attributes set on `<mux-player>` before DOM connection could be missed by the inner `<mux-video>` at initial load time.
> 
> `mux-player` now (1) seeds `mux-video.metadata` during `#init()` for any pre-existing `metadata-*` attributes, and (2) overrides `setAttribute`/`removeAttribute` to synchronously re-sync `metadata` whenever any `metadata-*` attribute changes (since these attributes are open-ended and not covered by `observedAttributes`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bca459e69d8c003ee379169528024cfb7fc4668. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes (approved):

luwes (changes requested):
actually could we solve this without preventing the load() to go through when not connected?

R-Delfino95:
> actually could we solve this without preventing the load() to go through when not connected?

I updated the fix so `load()` always runs as before. Changes are now only in `mux-player`: when `#init()` runs it immediately propagates any `metadata-*` attrs already set, and also override `setAttribute`/`removeAttribute` to push those changes synchronously as they come in. That way by the time the `load()` microtask fires, the metadata is already in place regardless of when `appendChild` happens

luwes (approved):
seems fine to me!

PR #1332, opened 2026-05-12
fix: remove inline onclick CSP violation in retry link

Jerricho93:
Closes #1331

The "click here to retry" link shown after HLS manifest load retries are exhausted used an inline `onclick` attribute, which violates `script-src-attr 'none'` CSP policies.

### Changes
- **`playback-core`**: Replaces the inline `onclick` with a typed `reload` flag on `MediaError`. The message is now plain text, so direct consumers of `playback-core`/`mux-video` are no longer exposed to raw HTML in the error message.
- **`mux-player`**: Builds the retry link HTML (using `data-mux-reload`) in `muxMediaErrorToDialogMessage`, and handles clicks via event delegation on the player element using `composedPath()` to reach through shadow DOM boundaries.

---

> [!NOTE]
> **Low Risk**
> Low risk, localized to error messaging and retry-click handling; main risk is regressions where the reload link no longer appears or reloads in some shadow DOM/event-path edge cases.
> 
> **Overview**
> Removes the CSP-violating inline `onclick` reload handler from the HLS manifest retry exhaustion error.
> 
> `playback-core` now marks these errors with a `MediaError.reload` flag and emits a plain-text message, while `mux-player` conditionally renders the retry anchor (`data-mux-reload`) and handles retry clicks via delegated `click` handling using `composedPath()` (with proper listener cleanup).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cebafedac7669f1798d341db621310e9a57d4d76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1342, opened 2026-07-14
fix: replace inline event handlers with addEventListener to comply wi…

### LINKED ISSUE #1341: Bug: mux-player error dialog uses inline `on*` event-handler attributes that violate a strict CSP

mux-player, mux-player-react, playback-core

Chrome, Safari, Firefox, Edge ("Edgeium")

macOS

### Description

`@mux/mux-player` renders its **error dialog** using inline HTML event-handler attributes.
Under a strict `Content-Security-Policy` where `script-src` does **not** include `'unsafe-inline'`
(and there is no `script-src-attr` override), the browser refuses these inline handlers and raises
`script-src-attr` / `blocked-uri: inline` violations.

Three inline handlers exist in the error-dialog rendering path:

| Handler | Element | Behavior | When it fires |
|---|---|---|---|
| `onclick="window.location.reload(); return false;"` | retry `<a>` inside the error message | reloads the page | on click of the retry link |
| `onclose="${onCloseErrorDialog}"` | error dialog `<media-theme>` | dialog close handler | on dialog close |
| `onfocusin="${onFocusInErrorDialog}"` | error dialog `<media-theme>` | focus-in handler | on dialog open (not click-gated) |

The `onclose`/`onfocusin` handlers are not gated on user interaction; they trip CSP essentially
whenever an error dialog appears. None of these fire on normal (non-error) playback.

Note: even though the player's `html` template processor eventually rebinds `on*` values via
`addEventListener` and strips the attribute, the literal `on*=` text still lives in the template
HTML string that is parsed/cloned into the live DOM, so the browser attempts to compile the inline
handler (and CSP refuses it) *before* the processor removes the attribute. The retry-link `onclick`
is a genuine raw inline handler assembled in a `MediaError` message string in `playback-core`.

The fix is the same class for all three: stop expressing behavior as inline HTML attributes; attach
handlers to real DOM nodes via `addEventListener` after they exist. No new reliance on
`'unsafe-inline'`; behavior must stay identical.

### Reduced test case

_No response_

### Steps to reproduce

1. Serve a page whose `Content-Security-Policy` has a `script-src` with **no** `'unsafe-inline'` and no `script-src-attr` override (or use `Content-Security-Policy-Report-Only` and listen for `securitypolicyviolation`).
2. Load `<mux-player>` on that page.
3. Force an error state (e.g. a bad/expired playback ID) so the error dialog renders.

### Current Behavior

- The browser refuses the dialog's inline `onclose`/`onfocusin` handlers on dialog open, and the retry link's inline `onclick` on click.
- CSP violations are reported with `violated-directive: script-src-attr` and `blocked-uri: inline`.

### Expected Behavior

- The error dialog appears; focus/close behave normally; the retry link reloads the page.
- **No** `script-src-attr` / `blocked-uri: inline` violation is raised.
- Works without the consuming page needing `'unsafe-inline'` in `script-src`.

### Errors

Refused to execute inline event handler because it violates the following Content Security Policy
directive: "script-src 'self' ...". Either the 'unsafe-inline' keyword, a hash, or a nonce is
required to enable inline execution.
(violated-directive: script-src-attr, blocked-uri: inline)

3.11.6

Jerricho93:
Closes #1341

The error dialog's `<media-theme>` element used inline `onclose` and `onfocusin` attributes, which violate a strict `script-src` CSP (no `'unsafe-inline'`, and no `script-src-attr` override). Unlike the retry link (fixed in #1332), these are not gated on user interaction; they fire on the dialog's own focus/lifecycle events, so they trip CSP essentially whenever an error dialog appears.

Although the player's `html` template processor rebinds `on*` values via `addEventListener` and strips the attribute at runtime, the literal `on*=` text still lives in the template HTML string that is parsed/cloned into the live DOM. The browser therefore attempts to compile the inline handler (and CSP refuses it, reporting `violated-directive: script-src-attr`, `blocked-uri: inline`) *before* the processor removes the attribute.

### Changes
- **`mux-player`**: Removes the `onclose`/`onfocusin` attributes from the `<media-theme>` template. The two handlers are now bound programmatically on the `media-theme` element via `addEventListener` in `#setUpErrors`, with matching cleanup in `disconnectedCallback`. The handlers keep their existing `composedPath()` / `media-error-dialog` guards, so behavior is unchanged.
- Removes the now-unused `onCloseErrorDialog` / `onFocusInErrorDialog` template props from `MuxTemplateProps`.
- **test**: Adds a `template.test.js` regression test asserting the rendered template contains no inline `on*` event-handler attributes. It checks the raw template string (pre part-processing), since that is what the browser parses into the DOM and what CSP evaluates (the post-render DOM would hide the regression because the processor strips the attributes).

### Testing
- `npm test` in `packages/mux-player` — 134 passing, 0 failed.
- Existing error-dialog open/close tests (`player.test.js`) continue to pass, confirming the rebound `close`/`focusin` handlers behave identically.
- Static check: `grep -oE '\bon[a-z]+="[^"]*"' dist/mux-player.mjs` returns no error-dialog handlers after build.

### Notes
- This is a follow-up to #1332, which removed the inline `onclick` from the retry link. Together they clear all inline event-handler attributes from the error-dialog path so consumers (e.g. the `player.mux.com` iframe embed) can enforce a strict `script-src` without `'unsafe-inline'`.
- No reliance on `'unsafe-inline'` is introduced; the fix is purely `addEventListener`-based.

---

> [!NOTE]
> **Low Risk**
> Small, localized wiring change with existing tests and a CSP regression guard; no auth or data-path changes.
> 
> **Overview**
> **Removes inline `onclose` / `onfocusin` from the `<media-theme>` template** so strict `script-src` (no `'unsafe-inline'`) no longer trips `script-src-attr` when the error dialog appears. The browser evaluates inline handlers in the parsed template string before the `html` processor can strip them.
> 
> **Error-dialog close and focus behavior is unchanged:** `#onCloseErrorDialog` and `#onFocusInErrorDialog` are class fields with the same `composedPath()` / `media-error-dialog` guards, wired in `#setUpErrors` with `addEventListener('close'|'focusin')` on `mediaTheme` and removed in `disconnectedCallback`. The corresponding props are dropped from `MuxTemplateProps`.
> 
> **Adds a regression test** that the raw template HTML (pre part-processing) contains no `on*` event-handler attributes, since post-render DOM would hide a reintroduction.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0dbffa3f476de4e35816b92e44c19f20cce18a7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/mux-player/src/base.ts
+++ b/packages/mux-player/src/base.ts
@@ -305,22 +305,24 @@ class MuxPlayerElement extends VideoApiElement implements IMuxPlayerElement {
       window.location.reload();
     }
   };
+  // Bound programmatically (see #setUpErrors) instead of via inline on* template
+  // attributes, which are refused by a strict CSP (script-src without 'unsafe-inline').
+  #onCloseErrorDialog = (event: Event) => {
+    const localName = (event.composedPath()[0] as HTMLElement)?.localName;
+    if (localName !== 'media-error-dialog') return;
+
+    this.#setState({ isDialogOpen: false });
+  };
+  #onFocusInErrorDialog = (event: Event) => {
+    const localName = (event.composedPath()[0] as HTMLElement)?.localName;
+    if (localName !== 'media-error-dialog') return;
+
+    const isFocusedElementInPlayer = containsComposedNode(this, document.activeElement);
+    if (!isFocusedElementInPlayer) event.preventDefault();
+  };
   #captionsMovementCleanup?: () => void;
   #state: Partial<MuxTemplateProps> = {
     ...initialState,
-    onCloseErrorDialog: (event) => {
-      const localName = (event.composedPath()[0] as HTMLElement)?.localName;
-      if (localName !== 'media-error-dialog') return;
-
-      this.#setState({ isDialogOpen: false });
-    },
-    onFocusInErrorDialog: (event) => {
-      const localName = (event.composedPath()[0] as HTMLElement)?.localName;
-      if (localName !== 'media-error-dialog') return;
-
-      const isFocusedElementInPlayer = containsComposedNode(this, document.activeElement);
-      if (!isFocusedElementInPlayer) event.preventDefault();
-    },
   };
 
   static get NAME() {
@@ -497,6 +499,8 @@ class MuxPlayerElement extends VideoApiElement implements IMuxPlayerElement {
     this.media?.removeEventListener('loadstart', this.#onLoadStart);
     this.removeEventListener('error', this.#onError);
     this.removeEventListener('click', this.#onReloadClick);
+    this.mediaTheme?.removeEventListener('close', this.#onCloseErrorDialog);
+    this.mediaTheme?.removeEventListener('focusin', this.#onFocusInErrorDialog);
     if (this.media) {
       this.media.errorTranslator = undefined;
     }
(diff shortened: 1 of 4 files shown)


spuppo-mux (approved):

luwes (approved):
LGTM 👍

PR #1344, opened 2026-07-17
fix: Add Network interruption recovery 

### LINKED ISSUE #1343: Bug: Player fails to recover after network drops mid-playback
When network fully drops during playback and the buffer drains <mux-player>/<mux-video> gets stuck in rebuffering and never recovers when the network returns. 
Loader spins forever, play/pause/seek do nothing, only a full reload fixes it.

## Reproduction:
- Play this [docs demo](https://www.mux.com/docs/guides/play-your-videos)
- DevTools Network -> Offline and let buffer drain 
- Switch network back to online. 

**Expected**: resumes (or controls usable). 
**Actual**: frozen until reload. Also reproducible by reloading while offline then going back online.

**Environment:** mux-player / mux-video / playback-core, hls.js (mse) path.

spuppo-mux:
closes https://github.com/muxinc/elements/issues/1343

## Fix
Adds **opt-in** network-interruption recovery to playback-core's hls.js error handling, off by default so existing integrations are unaffected.

Enable it with `maxReconnectRetries` (a positive value sets the per-stall retry budget; unset / `0` = off). On a fatal connectivity error (offline, no HTTP response, or 5xx), retry `hls.startLoad()` with exponential backoff (~1s → 30s cap), up to that many attempts. Each give-up drives the next attempt, up to `maxReconnectRetries`, then we fall back to the terminal reload error.

- Listen for the browser online event as a fast path for an immediate retry on reconnect.
- Surface a non-fatal `NETWORK_RECONNECTING` ("Reconnecting…") state during recovery -> reflects what's happening without disabling controls or counting as a Mux Data failure.
- Retry loop + reconnecting state are gated on actually rebuffering (same condition as the loading spinner); while there's still playable buffer, hls.js's own retry handles it.
- Confirm recovery only on `FRAG_BUFFERED` (a real buffered fragment). A manifest reload alone does NOT end recovery (segments can still be failing), and `playing` ends the stall UI but keeps recovery armed.
- HTTP 4xx (auth / not-found) is intentionally excluded as retrying won't help.

## Changes
- **playback-core**: recovery mechanics extracted into their own `network-recovery.ts` module (`setupNetworkRecovery` → `{ handleHlsError, onManifestLoaded }`), wired into the hls.js `ERROR` / `MANIFEST_LOADED` handlers in `loadMedia`; new `NETWORK_RECONNECTING` error code and `networkError` state flag; gated on the new `maxReconnectRetries` prop.
- **mux-video / mux-player**: numeric `max-reconnect-retries` attribute threaded through to playback-core; "Reconnecting…" dialog title/message.
- **i18n**: registered the new user-facing strings ("Reconnecting…", "Attempting to reconnect…", the reconnecting message, and the terminal "Network error, try reloading.") in playback-core's `lang/en.json`.

## Testing
- playback-core (`network-recovery.test.js`): recovery flow (interruption → reconnecting → online → resume → clear), 404-no-retry gating, HTTP status-0 arming, the opt-in default-off contract, `loadSource` when the manifest never loaded, and manifest-reload-does-not-end-recovery — plus unit tests of the scheduler (one retry per give-up with no self-perpetuation, bounded give-up after `maxReconnectRetries`, and `FRAG_BUFFERED` cancelling the pending retry). Full suite: 37 passed.
- mux-player: reconnecting dialog title/message — 19 passed.
- Manually verified against the vanilla example with a local controllable HLS server (drop / hang / 503 / reset) and DevTools offline toggling.

## Notes
- Complements (doesn't replace) hls.js's own `fragLoadPolicy` / `playlistLoadPolicy` retries: hls.js retries first, then escalates to fatal and stops loading; this layer takes over after that (resume-after-fatal + `online` fast path + reconnecting UX), which hls.js retry config can't express.
- Because each attempt also waits out hls.js's full internal retry cycle, total time-to-terminal scales with `maxReconnectRetries` (each attempt ≈ hls.js's internal cycle + our backoff). Pick a smaller value if a shorter give-up window is preferred.
- The terminal error after the retry budget is fatal and disables controls; within the retry window controls stay usable.
- Native (Safari) playback path is unchanged; this targets the hls.js/MSE path where the bug reproduces.

---

> [!NOTE]
> **Medium Risk**
> Touches core hls.js error handling and playback lifecycle; default-off limits blast radius, but enabled recovery changes when fatal errors appear and how MANIFEST_LOADED clears state.
> 
> **Overview**
> Adds **opt-in** network stall recovery for the hls.js path, controlled by **`maxReconnectRetries`** (unset/`0` = unchanged behavior).
> 
> **playback-core** introduces `network-recovery.ts`, wired from hls.js `ERROR` / `MANIFEST_LOADED`: after fatal connectivity failures (offline, HTTP 0, 5xx—not 4xx), it retries `startLoad` / `loadSource` with exponential backoff up to the configured budget, listens for `online` for an immediate retry, and surfaces a non-fatal **`NETWORK_RECONNECTING`** state while rebuffering. Recovery clears on **`FRAG_BUFFERED`**; manifest reload alone no longer clears errors during an active recovery episode.
> 
> **mux-video** / **mux-player** expose **`max-reconnect-retries`**; mux-player maps reconnecting errors to dialog copy and i18n strings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd213052eb270f662b53d843d0ca16e8618e4243. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/playback-core/src/network-recovery.ts
```diff
@@ -0,0 +1,248 @@
...
+    if (!hasManifestLoaded && src) {
+      hls.loadSource(src);
+      return;
+    }
+    // Resume loading from the current playhead. Passing the position (instead of startLoad()'s
+    // default "resume from last") keeps recovery aligned with a seek that happened during the
+    // outage: if the user seeked to an unbuffered spot the seek stays pending (seeking never
+    // resolves without data), so we must load *that* position for it to complete once the
+    // origin returns - not fetch a stale one or stall waiting for the seek to settle.
+    hls.startLoad(mediaEl.currentTime);
```
luwes, line 111: I think this is fine but I faintly remember calling hls.startLoad multiple times could trigger an error in hls.js.

luwes (approved):
LGTM 👍 good it's gated behind an opt-in config

packages/mux-player/src/base.ts
```diff
@@ -828,6 +829,12 @@ class MuxPlayerElement extends VideoApiElement implements IMuxPlayerElement {
         }
         break;
       }
+      case MuxVideoAttributes.MAX_RECONNECT_RETRIES: {
+        if (newValue == null || newValue !== oldValue) {
+          this.maxReconnectRetries = Number(newValue);
+        }
```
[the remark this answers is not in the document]
spuppo-mux: This calls media.maxReconnectRetries which does NaN validation

packages/playback-core/src/network-recovery.ts
spuppo-mux: Just in case I changed the behavior a bit so it waits fully for hlsjs to give up (fatal network error) to trigger a new recovery -> new startLoad.
Happens to fix what Cursor commented after updating from main ([this](https://github.com/muxinc/elements/pull/1344#discussion_r3624500173))

luwes (approved):

Issue #1193, opened 2025-09-08
Bug: Is react player not reactive with respect to tracking props?


tommhuth:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which Mux Elements/Packages does this apply to? Select all that apply

mux-player-react

### Which browsers are you using?

Chrome

### Which operating systems are you using?

macOS

### Description

We respect the players cookie choice, and have tracking disabled by default and only enable it when cookie information has been accepted. when i do this i get seemingly no tracking at all, even when `disableTracking` and `disableCookies` are set to false after page load. is this a non reactive prop that must be initialised to true to get any tracking?

when i default to `undefined` before setting `true`, i do get tracking, but is that really the same as always having `true` if tracking is enabled by default?

### Steps to reproduce

1. initialze with `disableTracking` and `disableCookies` false
2. update to true 
3. no tracking in network tab even after enabled

### Current Behavior

no tracking if inited to false 

### Expected Behavior

should start tracking after prop is true

### What version of the package are you using?

0.24.0

ronald-urbina:
@tommhuth Thanks for raising this issue. We found a solution and the fix will be included in the next version of `<mux‑player>` which should be published in the coming days.

PR #1349, opened 2026-08-04
fix: make disableCookies reactive

R-Delfino95:
Related to https://github.com/muxinc/elements/issues/1193 — that one made `disableTracking` reactive, `disableCookies` was left behind.

### Description

`disableCookies` gates the persistent `muxData` cookie that players write through their built-in Mux Data. Today it only takes effect when a player initialises. Changing it afterwards doesn't work in either direction:

- **Turning it on** deletes the cookie, but the running Mux Data monitor writes it straight back on its next beacon, about 10 seconds later.
- **Turning it off** does nothing at all. The visitor's `mux_viewer_id` never comes back for the rest of the page.

That's a problem for consent flows. A consent manager answers after the page has rendered, so the value is almost always set *after* the player exists. The only way around it today is to not render the player until consent is known, which means it can't be server-rendered.

This PR makes the prop reactive, so a player can be rendered up front and told about consent later.

### Why it wasn't reactive

Mux Data reads `disableCookies` once, when its monitor is created, and there's no setter for it afterwards. So the only way to apply a change is to create a new monitor. `videojs/v10` reaches the same conclusion in its `MuxData` component, and this follows its shape — including its guards, so the two implementations stay comparable.

### How it works

When `disableCookies` changes, Mux Data is re-attached: the old monitor is torn down and a new one is created with the new value. The `<video>` element and the hls.js instance are left alone, so **playback isn't interrupted and anything attached to the element imperatively survives** — chapters, caption tracks, listeners. That's the difference from the `disable-tracking` branch right above it, which reloads the media and loses all of that.

`<mux-video>` and `<mux-audio>` both handle the attribute this way, and `<mux-player>` picks it up through the `<mux-video>` it renders. (`disable-tracking` is still init-only on `<mux-audio>`, same as before — that one reloads the media, so it's a separate conversation.)

Turning cookies back on reuses the existing `muxData` cookie if it's still there, so a returning visitor keeps their identity.

Two behaviours worth being explicit about:

**A change while a video is playing splits the Mux Data view in two** — one `viewend`, one new `viewstart`. It's inherent to re-creating the monitor, and it's the same behaviour as `videojs/v10`. Applying the change *before* playback starts costs nothing, because Mux Data only closes a view that had already started, and that's the common case for a consent banner.

**A player that initialises with cookies disabled no longer deletes an existing cookie.** It used to, and that turned out to be harmful: a server-rendered page can't read consent, so it always renders the cookie-less state, and deleting there wiped a returning visitor's `mux_viewer_id` moments before the client granted consent. Mux Data then minted a new identity on every page load. The element can't tell "the visitor declined" from "consent isn't known yet" — same markup, opposite meaning — so it now leaves the cookie alone and never writes to it. Apps that need a stale cookie cleared should do it themselves, where the real consent state is known.

### The React side

Making the element reactive isn't enough for server-rendered pages, because **React never reconciles attributes during hydration** — it warns that the mismatch "won't be patched up" and leaves the server's value in the DOM. So a server-rendered `disable-cookies` would stay there forever and the player would be stuck in the denied state.

`mux-player-react` and the `@mux/mux-video/react` wrapper now also apply `disableCookies` as a property after mount, which is what `mux-player-react` already does for `playbackId`, `metadata` and friends. That's the piece that makes a server-rendered, consent-gated player actually work.

`@mux/mux-video-react` and `@mux/mux-audio-react` are separate implementations with no custom element, so they got their own equivalent: an effect that watches the prop.

### Testing

13 unit tests in `packages/mux-video/test/index.test.js` (46 in the file now), covering both directions, no media reload, the cookie left alone at init, several changes in one tick collapsing into a single monitor, a change reverted in the same tick, and the cookie being cleared even when a monitor flushes its final beacon late or when several players are disabled at once.

`<mux-audio>` and `<mux-player>` get the same shape at a smaller scale, 6 and 5 tests, since everything they share lives in `playback-core`: both directions, no re-attach when the same value is re-applied, no media reload, the cookie cleared at runtime and left alone at init.

One existing test in the `<mux-video>` suite changed: the chapters one now waits for the chapters track instead of the `loadstart` event, which can fire before the listener is attached and made the test hang.

Verified manually with real playback on every flavour — `<mux-video>`, `<mux-audio>`, `<mux-player>`, `@mux/mux-video-react`, `@mux/mux-video/react`, `@mux/mux-audio-react` and `@mux/mux-player-react` — one at a time and then all seven on one page sharing a cookie, plus a server-render-and-hydrate harness on React 18 and 19:

- Revoking mid-playback deletes the cookie and it stays deleted well past the 10s heartbeat.
- Granting mid-playback brings `mux_viewer_id` back within a second, reusing the existing one.
- Zero `emptied` / `loadstart` / `waiting` events on any change, and the playhead never stops.
- Chapters and the caption `TextTrack` object identity survive every change.
- On a server-rendered page, a returning visitor keeps the same `mux_viewer_id` across reloads.
- One `viewstart` and zero `viewend` per player when consent resolves before playback starts.
- With seven players on one page, revoking clears the shared cookie once and none of them writes it back.

---

> [!NOTE]
> **Medium Risk**
> Changes Mux Data monitoring, cookie handling, and viewer identity timing; mid-playback toggles split analytics views, but behavior is scoped and heavily tested for consent scenarios.
> 
> **Overview**
> Makes **`disableCookies`** take effect after the player loads, so consent managers can flip cookie behavior without remounting or delaying SSR.
> 
> **playback-core** adds `applyDisableCookies`, `reinitMuxData`, and `clearMuxDataCookies`. When the prop changes, Mux Data is torn down and re-monitored with the new option (mux-embed latches `disableCookies` at creation). Playback and hls.js stay untouched. Runtime disable clears the `muxData` cookie on a microtask (after teardown flushes); init with cookies off no longer deletes an existing cookie, so SSR “unknown consent” does not wipe `mux_viewer_id`. Bursts of changes in one tick coalesce to a single re-attach.
> 
> **Custom elements** (`mux-video`, `mux-audio`) route `disable-cookies` attribute changes through `applyDisableCookies` instead of only clearing cookies inline.
> 
> **React**: `mux-video-react` / `mux-audio-react` watch `props.disableCookies`; `mux-player-react` forwards it via `useObjectPropEffect`; `@mux/mux-video/react` re-applies `disableCookies` as a property after mount to fix hydration attribute mismatches.
> 
> Adds broad **unit tests** for re-attach, no media reload, cookie clearing, and multi-player edge cases; some existing tests get longer `waitUntil` timeouts for CI.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac5faa12748c27fa1b87e818469587cce7d8ee08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/mux-audio-react/src/index.tsx
+++ b/packages/mux-audio-react/src/index.tsx
@@ -5,6 +5,7 @@ import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import {
   allMediaTypes,
+  applyDisableCookies,
   initialize,
   teardown,
   MuxMediaProps,
@@ -44,22 +45,22 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
     setSrc(toMuxVideoURL(props) ?? outerSrc);
   }, [outerSrc, playbackId]);
 
-  useEffect(() => {
-    const propsWithState = {
-      // NOTE: Applying playerInitTime first as a simple way of overriding it if/when folks update
-      // the value via props after initial load (e.g. when swapping src)
-      playerInitTime,
-      ...props,
-      src,
-      playerSoftwareName,
-      playerSoftwareVersion,
-      autoplay: autoPlay,
-    };
+  const getPropsWithState = () => ({
+    // NOTE: Applying playerInitTime first as a simple way of overriding it if/when folks update
+    // the value via props after initial load (e.g. when swapping src)
+    playerInitTime,
+    ...props,
+    src,
+    playerSoftwareName,
+    playerSoftwareVersion,
+    autoplay: autoPlay,
+  });
 
+  useEffect(() => {
     // mediaEl required caching here so the ref was not null in the unmount callback.
     let mediaEl = mediaElRef.current;
     if (mediaEl) {
-      playbackCoreRef.current = initialize(propsWithState, mediaEl, playbackCoreRef.current);
+      playbackCoreRef.current = initialize(getPropsWithState(), mediaEl, playbackCoreRef.current);
     }
 
     return () => {
@@ -68,6 +69,12 @@ const MuxAudio = React.forwardRef<HTMLAudioElement | undefined, Partial<Props>>(
       playbackCoreRef.current = undefined;
     };
   }, [src]);
+
+  // mux-embed latches `disableCookies` when the monitor is created and offers no setter, so Mux Data
+  // has to be re-attached to pick up a change. Unlike a src change, that leaves the media alone.
+  useEffect(() => {
+    applyDisableCookies(getPropsWithState(), mediaElRef.current, playbackCoreRef.current);
+  }, [props.disableCookies]);
 
   useEffect(() => {
     playbackCoreRef.current?.setAutoplay(autoPlay);
--- a/packages/mux-audio/src/index.ts
+++ b/packages/mux-audio/src/index.ts
@@ -1,5 +1,6 @@
 import { globalThis } from './polyfills';
 import {
+  applyDisableCookies,
   initialize,
   teardown,
   generatePlayerInitTime,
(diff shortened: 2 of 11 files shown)


decepulis (changes requested):
This _seems_ ok to me, but I think it should include mux-audio, too!

decepulis (approved):

PR #1357, opened 2026-09-15
fix: bumped hls.js version

spuppo-mux:
Closes https://github.com/muxinc/devextravaganza/issues/290

Fixes subtitles rendering at the wrong time on assets that contain an `EXT-X-DISCONTINUITY`. Cues are shifted forward by the start time of the first subtitle fragment loaded after the discontinuity, so they either appear at the wrong moment or land past the end of the video and never show at all.

Related hlsjs links:
- [Issue#4888: Cue offset is not being calculated correctly when stream is played from a non-zero continuity segment](https://github.com/video-dev/hls.js/issues/4888)
- [Issue#7850: Possible bug in TimelineController._parseVTTs: VTT fragments parsed with undefined initPTS when maxAvCC > -1, causing subtitle gap after SSAI discontinuities](https://github.com/video-dev/hls.js/issues/7850)
- [PR#7852: Align WebVTT segments missing X-TIMESTAMP-MAP header with media timestamps](https://github.com/video-dev/hls.js/pull/7852)

### Also in this PR
- `min-cap-level-controller.ts`: `HlsConfig.capLevelController` became optional in 1.7, so `class … extends Hls.DefaultConfig.capLevelController` now fails. Kept the value on DefaultConfig and narrowed the type; importing CapLevelController by name would compile but break the CJS bundle, since `dist/hls.js` (UMD) exports only the Hls class; the named exports exist only in `dist/hls.mjs`. 

---

> [!NOTE]
> **Medium Risk**
> This is a minor-version bump of the core HLS engine used for adaptive streaming and subtitles, so regressions in playback, ABR, or caption sync are possible despite the targeted bugfix motivation.
> 
> **Overview**
> Upgrades **hls.js** from `~1.6.15` to `~1.7.0` across the monorepo (`@mux/playback-core` as a runtime dependency; mux-audio, mux-player, mux-video as dev deps), with lockfile hoisting to a single `hls.js@1.7.3`. The intent is to pick up upstream fixes for **WebVTT cue timing** on streams with `EXT-X-DISCONTINUITY` (cues no longer shifted or dropped after SSAI-style breaks).
> 
> **`min-cap-level-controller.ts`** is adjusted because hls.js 1.7 types `Hls.DefaultConfig.capLevelController` as optional: the controller base class is asserted/narrowed and a defensive runtime check logs if it is missing (still sourced from DefaultConfig for CJS bundle compatibility).
> 
> **`playback-core` tests** for `initial-estimate-segments` now stub fragments with `sn: 0` so hls.js 1.7’s stricter `fragmentsAreEqual` handling does not throw before custom listeners run.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a06e5d70c3d511d610fad80d207cc4ba366fd807. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


luwes (approved):
👍

